PR #1, opened 2016-12-09
Initial code

kyessenov:
This is the initial code skeleton to get things started.
The build environment is both bazel and go-run compatible.


kyessenov:
I needed someone to approve since I cannot self-approve.

kyessenov:
What we have right now is:
- basic abstract config object and store definitions in /model
- implementation in /env for Kubernetes using 3rd party resources
  - I'll take your suggestion and rename this appropriately to make it clear that Kubernetes is one supported platform, not the only
- unit tests and integration test for Minikube

What we want to get to:
- concrete config model, as being worked out by Mixer
- config distribution model, not clear at this point, being designed
- config versioning, not designed yet
- validation, dry-runs, etc.

Bazel is not required to build or test this code, just make sure to check out client-go at the right SHA (see WORKSPACE) or the imports will fail.

Speaking of Bazel, I added rules to create a docker image as a target in docker/ with some hacks. Seems like this might save us some work (see [1] and [2]).

[1]: https://github.com/bazelbuild/rules_go/pull/98
[2]: https://github.com/bazelbuild/rules_go/issues/144

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+		}
+	}
+	return out
+}
+
+// CamelCaseToKabobCase converts "MyName" to "my-name"
+func CamelCaseToKabobCase(s string) string {
+	var out bytes.Buffer
+	for i := range s {
+		if 'A' <= s[i] && s[i] <= 'Z' {
```
douglas-reid: do you want [unicode.IsUpper](https://golang.org/pkg/unicode/#IsUpper)?

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+			if i > 0 {
+				out.WriteByte('-')
+			}
+			out.WriteByte(s[i] - 'A' + 'a')
```
douglas-reid: strings.ToLower(string(s[i])) ?

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+		err := kr.client.Extensions().ThirdPartyResources().Delete(apiName, &api.DeleteOptions{})
+		if err != nil {
+			out = multierror.Append(out, err)
+		}
+	}
+	return out
+}
+
+// CamelCaseToKabobCase converts "MyName" to "my-name"
+func CamelCaseToKabobCase(s string) string {
```
douglas-reid: i feel like this could be written using strings.Map, a la [strings.Title](https://golang.org/src/strings/strings.go?s=13125:13152#L475).

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+		Metadata: api.ObjectMeta{Name: encodeName(obj.ConfigKey)},
+		Data:     data,
+	}
+
+	err = kr.dyn.Post().
+		Namespace(kr.Namespace).
+		Resource(obj.Kind + "s").
+		Body(out).
+		Do().Error()
+	if err != nil {
```
douglas-reid: we can just remove this and the following `return nil` correct?

simply: `return err` ?

or even: `return kr.dyn.Post()...` ?

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+
+func (kr *KubernetesRegistry) Delete(key model.ConfigKey) {
+	err := kr.dyn.Delete().
+		Namespace(kr.Namespace).
+		Resource(key.Kind + "s").
+		Name(encodeName(key)).
+		Do().Error()
+	if err != nil {
+		log.Printf(err.Error())
+
```
douglas-reid: nit: remove newline

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+}
+
+func (kr *KubernetesRegistry) List(kind string) []*model.Config {
+	_, ok := kr.mapping[kind]
+	if !ok {
```
douglas-reid: collapse into single line: `if _, ok := kr.mapping[kind]; !ok {`

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+
+	// Marshal to JSON bytes
+	str, err := json.Marshal(config.Data)
+	if err != nil {
+		return nil, err
+	}
+
+	// Unmarshal from bytes to proto
+	err = jsonpb.UnmarshalString(string(str), pb)
+	if err != nil {
```
douglas-reid: here and above, collapse into single line.

env/kube_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"k8s.io/client-go/1.5/kubernetes"
+	"k8s.io/client-go/1.5/pkg/api"
+	"k8s.io/client-go/1.5/pkg/api/v1"
+
+	"istio.io/manager/model"
+	"istio.io/manager/test"
+)
+
+func TestCamelKabob(t *testing.T) {
+	if CamelCaseToKabobCase("ExampleNameX") != "example-name-x" {
```
douglas-reid: here, and elsewhere: prefer table-driven tests with got, want.  Begin log messages with something like: "CamelCaseToKabobCase(%s) => ..."

model/functions.go
```diff
@@ -0,0 +1,78 @@
...
+	Name() string
+	// Generators produce configuration for the component
+	Generators() []Generator
+	// Distribute pushes configuration to the component
+	Distribute([]*ConfigOutput) error
+}
+
+// Manager defines the data-flow of the configuration artifacts
+type Manager struct {
+	Registry        Registry
```
douglas-reid: comments on the export items?

model/store.go
```diff
@@ -0,0 +1,117 @@
...
+	Name string
+	// Config version, e.g. "prod", "canary", "v1".
+	// "default" version is reserved, use ""
+	Version string
+}
+
+// Config object holds the normalized config objects defined by Kind schema
+type Config struct {
+	ConfigKey
+	Content interface{}
```
douglas-reid: comments on exported ?

model/store.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Registry of the configuration objects
+type Registry interface {
+	Get(key ConfigKey) (*Config, bool)
```
douglas-reid: comments on exported?

test/mocks.go
```diff
@@ -0,0 +1,158 @@
...
+const MockKind = "MockConfig"
+const MockName = "my-qualified-name"
+
+var MockKey = model.ConfigKey{
+	Kind: MockKind,
+	Name: MockName,
+}
+
+var MockConfigObject = MockConfig{
+	Pairs: []*ConfigPair{
```
douglas-reid: i would just make this all one line.

test/mocks.go
```diff
@@ -0,0 +1,158 @@
...
+	var out = make([]*model.Config, 0)
+	for _, v := range r.store {
+		if v.Kind == kind {
+			out = append(out, v)
+		}
+	}
+	return out
+}
+
+const MockKind = "MockConfig"
```
douglas-reid: can we move the const and var blocks to the top, and then collapse into const() and var() blocks?

test/mocks.go
```diff
@@ -0,0 +1,158 @@
...
+
+func NewMockRegistry() model.Registry {
+	return &MockRegistry{
+		store:   make(map[model.ConfigKey]*model.Config),
+		mapping: MockMapping,
+	}
+}
+
+func (r *MockRegistry) Get(key model.ConfigKey) (*model.Config, bool) {
+	out, err := r.store[key]
```
douglas-reid: just `return r.store[key]` ?

env/kube_config.go
```diff
@@ -0,0 +1,85 @@
...
+	return &el.TypeMeta
+}
+
+// GetListMeta adapts to ListMetaAccessor
+func (el *ConfigList) GetListMeta() unversioned.List {
+	return &el.Metadata
+}
+
+// The code below is used only to work around a known problem with third-party
+// resources and @ugorji. If/when these issues are resolved, the code below
```
kyessenov: Upstream bug in Kubernetes. It turns out there is a whole bunch of bugs in API server regarding Third Party resources. ugorji is the author of JSON serialization library that requires this workaround.

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+	"k8s.io/client-go/1.5/pkg/api/unversioned"
+	"k8s.io/client-go/1.5/pkg/api/v1"
+	"k8s.io/client-go/1.5/pkg/apis/extensions/v1beta1"
+	"k8s.io/client-go/1.5/pkg/runtime"
+	"k8s.io/client-go/1.5/pkg/runtime/serializer"
+	"k8s.io/client-go/1.5/rest"
+	"k8s.io/client-go/1.5/tools/clientcmd"
+)
+
+const IstioAPIGroup = "istio.io"
```
kyessenov: Yes, I'll edit exported names to group them and add comments.

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+		return nil, err
+	}
+
+	version := unversioned.GroupVersion{
+		Group:   IstioAPIGroup,
+		Version: IstioResourceVersion,
+	}
+
+	config.GroupVersion = &version
+	config.APIPath = "/apis"
```
kyessenov: Maybe, but that constant should have been defined upstream in Kubernetes. The name seems to be required to be "apis".

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+// RegisterResources creates third party resources
+func (kr *KubernetesRegistry) RegisterResources() error {
+	var out error
+	for kind, v := range kr.mapping {
+		apiName := KindToAPIName(kind)
+		// initialize resource if it does not exist
+		res, err := kr.client.Extensions().ThirdPartyResources().Get(apiName)
+		if err != nil {
```
kyessenov: Good point.

env/kube_registry.go
```diff
@@ -0,0 +1,331 @@
...
+					ObjectMeta: v1.ObjectMeta{
+						Name: apiName,
+					},
+					Versions: []v1beta1.APIVersion{
+						{Name: IstioResourceVersion},
+					},
+					Description: v.Description,
+				}
+				_, err = kr.client.Extensions().ThirdPartyResources().Create(tpr)
+				if err != nil {
```
kyessenov: Wouldn't the line be too long?

env/kube_registry.go
kyessenov: This is ASCII (see regex validation). Unicode is not allowed in Kubernetes names.

env/kube_registry.go
kyessenov: Again, runes are a lot slower than ASCII.

env/kube_registry.go
kyessenov: The intent is that err will be consumed by a log service.
It cannot be just returned to the client since that would leak Kubernetes details.

env/kube_registry.go
kyessenov: Thanks.

platform/kube/minikube_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"k8s.io/client-go/1.5/kubernetes"
+	"k8s.io/client-go/1.5/pkg/api"
+	"k8s.io/client-go/1.5/pkg/api/v1"
+
+	"istio.io/manager/model"
+	"istio.io/manager/test"
+)
+
+func TestCamelKabob(t *testing.T) {
+	if CamelCaseToKabobCase("ExampleNameX") != "example-name-x" {
```
douglas-reid: suggest table-drive tests. that is more idiomatic.

env/kube_test.go
kyessenov: OK, I'll start using table-driven tests. Is there a specific benefit to them as opposed to just code?

platform/kube/registry.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+// RegisterResources creates third party resources
+func (kr *KubernetesRegistry) RegisterResources() error {
+	var out error
+	for kind, v := range kr.mapping {
+		apiName := KindToAPIName(kind)
+		// initialize resource if it does not exist
+		res, err := kr.client.Extensions().ThirdPartyResources().Get(apiName)
+		if err != nil {
```
douglas-reid: commented on this a bunch on the previous commit, but we should really strive to reduce the nesting here.

something like:

```
if err == nil {
    log...
    continue
}
if !errors.IsNotFound(err) {
    out = ...
    continue
}
tpr := &....
...
```

platform/kube/registry.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	return out
+}
+
+// DeregisterResources removes third party resources
+func (kr *KubernetesRegistry) DeregisterResources() error {
+	var out error
+	for kind := range kr.mapping {
+		apiName := KindToAPIName(kind)
+		err := kr.client.Extensions().ThirdPartyResources().Delete(apiName, &api.DeleteOptions{})
```
douglas-reid: collapse these two lines.

platform/kube/registry.go
```diff
@@ -0,0 +1,331 @@
...
+		if err != nil {
+			out = multierror.Append(out, err)
+		}
+	}
+	return out
+}
+
+// CamelCaseToKabobCase converts "MyName" to "my-name"
+func CamelCaseToKabobCase(s string) string {
+	var out bytes.Buffer
```
douglas-reid: can we use strings.Map, a la strings.Title impl?

platform/kube/registry.go
```diff
@@ -0,0 +1,331 @@
...
+	for i := range s {
+		if 'A' <= s[i] && s[i] <= 'Z' {
```
douglas-reid: unicode.IsUpper?

PR #5, opened 2016-12-11
Introduce cfgeditor, a GUI editor for Istio config files.

geeknoid:
This is very basic right now, but gets the structure in place
for future work in this area.


geeknoid:
In order to build, this has a dependency on the go-bindata binary. I tried a variety of things but couldn't get the WORKSPACE file setup to pick up that dependency properly. Some help would be appreciated with getting the right voodoo in place.

kyessenov:
Code looks OK. This would be a good placeholder to start working on the viewer/visualizer of the config/state as well. 

I can take a look at the data packaging for the go binary.

PR #16, opened 2016-12-17
Kubernetes Controller implementation

kyessenov:


platform/kube/controller.go
```diff
@@ -0,0 +1,131 @@
...
+	)
+
+	return out
+}
+
+func (c *Controller) notify(obj interface{}) error {
+	if !c.HasSynced() {
+		return errors.New("Waiting till full synchronization")
+	}
+	log.Printf("%#v", obj)
```
geeknoid: In the mixer we're using glog. Would it be valuable to use a consistent logging story throughout the platform?

platform/kube/queue.go
```diff
@@ -0,0 +1,99 @@
...
+// Queue of work tickets processed using a rate-limiting loop
+type Queue interface {
+	// Push a ticket
+	Push(Task)
+	// Run the loop until a signal on the channel
+	Run(chan struct{})
+}
+
+// Task object for the event watchers; processes until handler succeeds
+type Task struct {
```
geeknoid: I think that you could replace the Task struct with a TaskFunc instead:

  type TaskFunc func() error

Callers can then just do:

  myQ.Push(myFunc)

which will work for either kind of function:

  func myFunc()
  func (t *myType) myFunc()

platform/kube/queue.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"log"
+	"sync"
+	"time"
+
+	"k8s.io/client-go/pkg/util/flowcontrol"
+)
+
+// Queue of work tickets processed using a rate-limiting loop
+type Queue interface {
```
geeknoid: Test suite?

kyessenov: Added

platform/kube/queue.go
kyessenov: The `obj` argument is coming from the event watcher and needs to be passed to `func`.
I'll probably need to add another argument for the event type (`add`, `remove`, `update`).

platform/kube/controller.go
kyessenov: I would like to use glog but I cannot include it together with Kubernetes without vendoring (due to a bug in `client-go` or `glog`). Any recommendation for another logging library?

kyessenov:
Thanks for a review!

platform/kube/controller.go
mandarjog: With a make file I had handled that by simply removing k8s client go's vendored Glog. That forces it to use the copy that is directly vendored by the top level app. If you can figure out how to do the same with bazel, it will work.

PR #17, opened 2016-12-19
Remove version and add status field to config objects

kyessenov:


model/registry.go
```diff
@@ -14,110 +14,46 @@
...
 // ConfigKey is the identity of the configuration object
 type ConfigKey struct {
 	// Config object kind, e.g.  "MyKind"
 	Kind string
 	// Config object name, e.g. "my-name"
 	Name string
-	// Config version, e.g. "prod", "canary", "v1".
-	// "default" version is reserved, use ""
-	Version string
+	// Config object namespace, e.g. "default"
```
douglas-reid: I might stick with convention here (start with export in comment).

Maybe something like: `// Namespace encodes the k8s-specific namespace of the ConfigKey, e.g. "default"`

model/validation.go
```diff
@@ -0,0 +1,106 @@
...
+	kindRegexp      = regexp.MustCompile("^[a-zA-Z][a-zA-Z0-9]*$")
+)
+
+// IsDNS1123Label tests for a string that conforms to the definition of a label in
+// DNS (RFC 1123).
+func IsDNS1123Label(value string) bool {
+	return len(value) <= dns1123LabelMaxLength && dns1123LabelRex.MatchString(value)
+}
+
+// Validate names in the config key
```
douglas-reid: nit: `// Validate confirms that the names in the ConfigKey are appropriately formatted.`

model/validation.go
```diff
@@ -0,0 +1,106 @@
...
+	if !IsDNS1123Label(k.Name) {
+		return fmt.Errorf("Invalid name: %q", k.Name)
+	}
+	if !IsDNS1123Label(k.Namespace) {
+		return fmt.Errorf("Invalid namespace: %q", k.Name)
+	}
+	return nil
+}
+
+// Validate mapping
```
douglas-reid: nit: `// Validate checks that each name in the KindMap conforms to spec and that the ProtoMessage type is valid.`

model/validation.go
```diff
@@ -0,0 +1,106 @@
...
+			return fmt.Errorf("Invalid kind: %q", k)
+		}
+		if proto.MessageType(v.MessageName) == nil {
+			return fmt.Errorf("Cannot find proto message type: %q", v.MessageName)
+		}
+	}
+	return nil
+}
+
+// ValidateConfig object
```
douglas-reid: nit: `// ValidateConfig ensures that a Config is valid by ...`

platform/kube/config.go
```diff
@@ -27,9 +27,8 @@ import (
 type Config struct {
 	v1.TypeMeta `json:",inline"`
 	Metadata    api.ObjectMeta         `json:"metadata"`
-	Spec        map[string]interface{} `json:"data"`
-	// TODO: enable structured status field
-	Status string
+	Spec        map[string]interface{} `json:"spec"`
```
douglas-reid: worth doc (these are exported)?

model/registry.go
```diff
@@ -14,110 +14,46 @@
...
 }
 
 // Config object holds the normalized config objects defined by Kind schema
 type Config struct {
 	ConfigKey
-	// Content holds the configuration struct
-	Content interface{}
+	// Spec holds the configuration struct
+	Spec interface{}
+	// Status holds the status information (may be null)
```
douglas-reid: what sort of Status information do expect to be here?  Perhaps we could add some examples?  interface{} makes it hard to get a feel.

model/registry.go
kyessenov: Ack

model/validation.go
kyessenov: Ack

model/validation.go
kyessenov: Ack

model/validation.go
kyessenov: Ack

platform/kube/config.go
kyessenov: They're same as the main declaration but exposed for Kubernetes serialization code.

PR #18, opened 2016-12-20
Cached registry and controller interface

kyessenov:
* Use cache for registry

* Extensible chained handlers

* Optimize Update event

### DIFF
--- a/model/registry.go
+++ b/model/registry.go
@@ -36,7 +36,8 @@ type Config struct {
 }
 
 // Registry of the configuration objects
-// Modification of object pointers violates thread-safety
+// Object pointers supplied and returned from this interface should be
+// treated as read-only. Updating them might violate thread-safety.
 type Registry interface {
 	// Get retrieves a configuration element, bool indicates existence
 	Get(key ConfigKey) (*Config, bool)
--- a/model/registry.go
+++ b/model/registry.go
@@ -36,10 +36,12 @@ type Config struct {
 }
 
 // Registry of the configuration objects
+// Modification of object pointers violates thread-safety
 type Registry interface {
+	// Get retrieves a configuration element, bool indicates existence
 	Get(key ConfigKey) (*Config, bool)
-	Delete(key ConfigKey)
-	Put(obj Config) error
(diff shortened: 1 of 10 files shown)


model/registry.go
```diff
@@ -36,10 +36,12 @@ type Config struct {
 }
 
 // Registry of the configuration objects
+// Modification of object pointers violates thread-safety
```
enricoschiattarella: The comment is not entirely clear to me.
Do you mean that client should not modify the object pointed by the *Config returned by Get() and List() or should not modify the object pointed by the *Config passed to Put() after calling Put() or all of them or something else ?

model/validation.go
```diff
@@ -65,7 +65,7 @@ func (km KindMap) Validate() error {
 }
 
 // ValidateConfig ensures that the config object is well-defined
-func (km KindMap) ValidateConfig(obj Config) error {
+func (km KindMap) ValidateConfig(obj *Config) error {
```
enricoschiattarella: Null check for obj ?

platform/kube/controller.go
```diff
@@ -16,93 +16,148 @@ package kube
...
 	namespace string,
-	resyncPeriod time.Duration) *Controller {
+	resyncPeriod time.Duration,
+) *Controller {
 	out := &Controller{
-		client:       client,
-		resyncPeriod: resyncPeriod,
-		queue:        NewQueue(1 * time.Second),
+		client: client,
+		queue:  NewQueue(1 * time.Second),
```
enricoschiattarella: Maybe a comment saying what this 1 sec interval is for ?

model/registry.go
kyessenov: Rewrote. It doesn't break thread-safety actually anymore (since objects are re-incarnated in controller.go), but it has no effect.

model/validation.go
kyessenov: Good catch

platform/kube/controller.go
kyessenov: Done

PR #19, opened 2016-12-21
Cache consistency tests and validation

kyessenov:
Kubernetes controllers should not rely on strong consistency of the config registry.
Added tests to validate some work flows on using cache + event handling together.


kyessenov:
Yeah, I looked at several implementations of the informer framework.
What I realize now is that it is difficult to work with multiple kinds (not just endpoints for kube-proxy) in a weakly-consistent world of etcd/kube-api.

PR #24, opened 2016-12-22
Import Amalgam8 proxy nanny

kyessenov:


proxy/envoy/config.go
```diff
@@ -0,0 +1,340 @@
...
+
+	return clusters
+}
+
+func buildWeightKey(service string, tags []string) string {
+	return fmt.Sprintf("%v.%v", service, BuildServiceKey("_", tags))
+}
+
+func buildRoutes() []Route {
+	var routes []Route
```
GregHanson: this line should be 
```
routes := []Route{}
```

otherwise the generated envoy.json file will be rejected when attempting the hot restart.  If the array is nil (like it is now), it will not generate the same json as an empty array

kyessenov:
Thanks. I addressed the comment and rebased the code.

PR #28, opened 2017-01-04
Implement services interface for kubernetes (no tags)

kyessenov:


model/service.go
```diff
@@ -44,15 +45,28 @@ type Endpoint struct {
 	// Address of the endpoint, typically an IP address
 	Address string `json:"ip_address,omitempty"`
 	// Port on the host address
-	Port int `json:"port,omitempty"`
+	Port int `json:"port"`
+	// Name of the port classifies ports for a single service
+	Name string `json:"name,omitempty"`
```
rshriram: What are we using the name field for, given that we have HTTP being designated as a protocol by itself.

model/service.go
```diff
@@ -44,15 +45,28 @@ type Endpoint struct {
...
+	// Protocol for the port: TCP, UDP (default is TCP)
+	Protocol Protocol `json:"protocol,omitempty"`
 }
 
+// Protocol defines network protocols for ports
+type Protocol string
+
+const (
+	ProtocolHTTP Protocol = "HTTP"
```
rshriram: refer to previous comment

platform/kube/controller.go
```diff
@@ -257,3 +257,62 @@ func (c *Controller) List(kind string, ns string) ([]*model.Config, error) {
...
+	return out
+}
+
+func convertProtocol(proto v1.Protocol) model.Protocol {
+	switch proto {
+	case v1.ProtocolTCP:
+		return model.ProtocolTCP
+	case v1.ProtocolUDP:
+		return model.ProtocolUDP
+	}
```
rshriram: if we are going to use HTTP as protocol, then its missing here.

platform/kube/controller.go
```diff
@@ -257,3 +257,62 @@ func (c *Controller) List(kind string, ns string) ([]*model.Config, error) {
...
+	return model.ProtocolTCP
+}
```
rshriram: Is the port structure validated somewhere else? Otherwise, we might have to validate right?

platform/kube/controller.go
frankbu: Also need HTTPS?

model/service.go
kyessenov: There could be multiple HTTP ports for a service. We have to further quality each port by a name to distinguish between them. I have not used this field yet, basically assuming there is only one port for a service.

model/service.go
kyessenov: Did you use port name to infer it's HTTP in amalgam8?

platform/kube/controller.go
kyessenov: Yes, both are missing since k8s only has UDP/TCP. UDP is not even supported by Envoy.
For TCP we need to start requiring annotations on the service definition for each port to specify which is it.

platform/kube/controller.go
kyessenov: Yes, services are assumed to be correct. What happens if it's not valid? We can't fail here, maybe we should keep track of black-listed services?

kyessenov:
Yes, I saw that code. I'm not comfortable with overloading port names with the protocol definition. How do you deal with status ports in a service? Those are also HTTP but can't be named http. Maybe use a prefix on the name at least?

model/service.go
rshriram: Yes.. And we did that based on how K8S had allowed the user to further qualify a TCP port. One example is here
http://kubernetes.io/docs/user-guide/services/#multi-port-services

Given the use of "name" to qualify the TCP port as HTTP/HTTPS etc., in Kubernetes, I assumed that your use of "name" field was for a similar intent.

platform/kube/controller.go
rshriram: http://kubernetes.io/docs/user-guide/services/#multi-port-services for reference

platform/kube/controller.go
rshriram: If we are pulling the service definition from K8S, then I guess we need not worry.

kyessenov: Yeah, that's my thinking. I'm trying to make it fail-safe for these core pieces like SDS so validation should cause a warning somewhere but not a failure.

platform/kube/controller.go
kyessenov: Taken into account!

model/service.go
kyessenov: Yes, but I also want to have duplicate HTTP ports (which can happen in k8s), so I check for a prefix and keep port name around.

PR #34, opened 2017-01-11
Define service tags

kyessenov:
This PR defines service tag as a label selector (`map[string]string`).
The code to compute endpoints for a service tag will be in a follow-up PR.

### DIFF
--- a/model/service.go
+++ b/model/service.go
@@ -180,7 +180,11 @@ func ParseServiceString(s string) *Service {
 func (t Tag) String() string {
 	labels := make([]string, 0)
 	for k, v := range t {
-		labels = append(labels, fmt.Sprintf("%s=%s", k, v))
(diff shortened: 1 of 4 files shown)


model/service.go
```diff
@@ -169,3 +176,36 @@ func ParseServiceString(s string) *Service {
 		Tags:      tags,
 	}
 }
+
+func (t Tag) String() string {
+	labels := make([]string, 0)
+	for k, v := range t {
+		labels = append(labels, fmt.Sprintf("%s=%s", k, v))
+	}
```
frankbu: would it make sense for the string representation of a key without value to be "key", instead of "key="? Then we could also allow simple tags like "v1", "prod", etc.

kyessenov: definitely, makes sense.

PR #38, opened 2017-01-13
Server-side configuration for Envoy with iptables

kyessenov:
Added traffic capture with an init container.

### DIFF
--- a/test/example/echo/echo.yaml
+++ b/test/example/echo/echo.yaml
@@ -2,9 +2,9 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: service1
+  name: a
   labels:
-    app: service1
+    app: a
 spec:
   type: NodePort
   ports:
@@ -12,18 +12,18 @@ spec:
     targetPort: 8080
     name: http
   selector:
-    app: service1
+    app: a
 ---
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
-  name: service1
+  name: a
 spec:
-  replicas: 3
+  replicas: 1
   template:
     metadata:
       labels:
-        app: service1
+        app: a
       annotations:
         pod.beta.kubernetes.io/init-containers: '[
           {
@@ -48,6 +48,23 @@ spec:
         imagePullPolicy: Always
         ports:
         - containerPort: 9999
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: test
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: test
+    spec:
+      containers:
+      - name: echo
+        image: gcr.io/google_containers/echoserver:1.4
+        ports:
+        - containerPort: 8080
 ---
 apiVersion: v1
 kind: Service
--- a/test/example/echo/echo.yaml
+++ b/test/example/echo/echo.yaml
@@ -2,9 +2,9 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: echoheaders
+  name: service1
   labels:
-    app: echoheaders
+    app: service1
 spec:
   type: NodePort
   ports:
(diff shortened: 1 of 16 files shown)


kyessenov:
Yes, I'd like to configure Envoy for all HTTP incoming traffic that does not originate in the same pod. Basically, server-side proxy. We should just scan the pod spec and create a listener for each container.
The client-side proxy is trickier. We want to capture all traffic originating in the pod (excluding the proxy) and then use destination Host/IP to make a decision. Can you share how you excluded the self-reference in your prototype?

enricoschiattarella:
The way I did it in the prototype (which was nginx-based at the time) was to use this rule:
'''
iptables -t nat -A OUTPUT -p tcp --dport 3306 -j REDIRECT --to-port 3307 -m owner '!' --uid-owner 999
'''
this applies only to traffic generated inside the pod (because it is in the OUTPUT chain) and excludes all traffic generated by a process owned by user with uid=999

So one way of doing it (possibly not the best one, though) would be to pick a specific UID, add the rule with in the init container, then in the proxy container make sure that the proxy runs with that uid. This solutin has the disadvantages that is prone to uid conflicts and possible abuse by an app that wants to bypass the proxy.

There are a couple of more solutions hinted by other people in the original proxy comparison doc. One entails using an system call to mark packets generated by the proxy and then match those packets (and exclude them) in iptables. This is more elegant but requires privilege for the proxy container and changes to the proxy code.

See if you can get the uid-based solution implemented, in the meanwhile I will take another look.

test/example/echo/echo.yaml
```diff
@@ -2,34 +2,54 @@
...
+            "name": "iptables",
+            "image": "gcr.io/istio-test/init:test",
+            "imagePullPolicy": "Always",
+            "securityContext": {
+              "privileged": true
+            },
+            "command": ["/bin/bash", "-c", "sudo iptables -t nat -A PREROUTING -p tcp --dport 8080 -j REDIRECT --to-port 9999"]
+          }
+        ]'
     spec:
```
rshriram: Dumb question: is this portable across kube versions (say 1.3, 1.4) etc. ?

test/example/echo/echo.yaml
```diff
@@ -2,34 +2,54 @@
...
         ports:
         - containerPort: 8080
+      - name: proxy
+        image: gcr.io/istio-test/runtime:test
+        args: ["proxy", "-s", "manager:8080", "--logtostderr", "-v", "2", "-p", "9999"]
+        securityContext:
+          privileged: true
+        imagePullPolicy: Always
+        ports:
+        - containerPort: 9999
```
rshriram: Do we need to run the proxy container as a privileged container? It may not sit well with security folks.
Secondly, I would add some config option in the code (and some additional config in proto) to optionally enable the IP tables trick in the proxy sidecar container. Because, the solution above works only with Envoy.
If we look at Nginx/HAproxy, we might need to generate additional configs to route the API call to localhost:port/foo/bar. Right?

Thoughts?

enricoschiattarella: In our current implementation, the iptables rules are programmed by an init container which starts in privileged mode, program the rules and exits. I think that this actually doesn't need full privilege, only CAP_NET_ADMIN, so it might be possible to tighten the security a bit more.
After that, the proxy and app container starts. The proxy listens on a single port > 1024 thanks to the new changes, so it doesn't need privilege.

As for the other proxies, haproxy already knows about SO_ORIGINAL_DST so it may not need further changes (needs to be tested). Nginx has a transparent mode which might be useful and I saw a patch for SO_ORIGINAL_DST (not in the official codebase). Linkerd has published its own solution, which looks quite convoluted.
Bottom line is that other proxies need to be looked at individually to determine what is needed.
For inbound traffic other non-iptables choices are possible but for outbound I don't know about any option that does not require iptables or changes to the app.

test/example/echo/echo.yaml
enricoschiattarella: This modifies iptables in the pod network namespace. Kubernetes only modifies the node networking namespace, so there are no conflicts or compatibility concerns.

test/example/echo/echo.yaml
moderation: We plan on disallowing containers running in privileged mode. We are moving some components out of containers into systemd services (where we have the experience and tooling to handle).

kyessenov: What about init containers? Is it OK to run privileged mode commands (`iptables`) in the init container?
@enricoschiattarella @rshriram The proxy doesn't need extra privileges. That was just a leftover from experimentation.

test/example/echo/echo.yaml
kyessenov: init containers require 1.1. TPR is a biggest constraint. They'll work properly in >1.5 and might need more manual plumbing in earlier versions.

test/example/echo/echo.yaml
rshriram: I was referring to the proxy container being marked as priviledged. But @kyessenov answered it..

kyessenov:
includes first cut at #36

PR #43, opened 2017-01-18
Client-side configuration for Envoy with iptables

kyessenov:


proxy/envoy/agent.go
```diff
@@ -62,10 +66,14 @@ mixer_options {
   mixer_server: "%s%s"
 }
 	`, EgressClusterPrefix, mixer))
+	if err := f.Chmod(0777); err != nil {
```
rshriram: This is super permissive for a JSON file. May be 0644?

proxy/envoy/config.go
```diff
@@ -33,6 +33,9 @@ func (conf *Config) WriteFile(fname string) error {
 	if err != nil {
 		return err
 	}
+	if err := file.Chmod(0777); err != nil {
```
rshriram: Same as above

proxy/envoy/config.go
```diff
@@ -98,7 +101,7 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service, mes
 		*/
 		Listeners: listeners,
 		Admin: Admin{
-			AccessLogPath: "/dev/stdout",
+			AccessLogPath: "/dev/null",
```
rshriram: Are the admin logs useless totally in our case?

proxy/envoy/config.go
```diff
@@ -170,7 +173,7 @@ func buildIngressListeners(instances []*model.ServiceInstance) ([]Listener, []Cl
 				Config: NetworkFilterConfig{
 					CodecType:  "auto",
 					StatPrefix: "ingress_http",
-					AccessLog:  []AccessLog{{Path: "/dev/stdout"}},
+					AccessLog:  []AccessLog{{Path: "/dev/null"}},
```
rshriram: I would rather keep it to either stdout or some file configurable by user. The monitoring reports that go the mixer are aggregate metrics. Until we have zipkin support for envoy, losing detailed access log seems not a good idea.

test/example/echo/echo.yaml
```diff
@@ -33,7 +33,10 @@ spec:
             "securityContext": {
               "privileged": true
             },
-            "command": ["/bin/bash", "-c", "sudo iptables -t nat -A PREROUTING -p tcp --dport 8080 -j REDIRECT --to-port 9999"]
+            "command": ["/bin/bash", "-c",
+              "sudo iptables -t nat -A PREROUTING -p tcp -j REDIRECT --to-port 5001",
+              "sudo iptables -t nat -A OUTPUT -p tcp -j REDIRECT --to-port 5002 -m owner \\! --uid-owner 1337"
+            ]
```
rshriram: Didn't you say that the owner marking has issues? A comment to that effect here would be helpful

proxy/envoy/agent.go
ayj: File creation and setting permissions can be done in one step, e.g. 
    os.OpenFile(path, O_CREATE | O_TRUNC | O_WRONLY, perms)

test/example/echo/echo.yaml
kyessenov: Yeah, it has issues, but it works for a first iteration. @enricoschiattarella is looking for a better solution.

proxy/envoy/config.go
kyessenov: No, it should be /dev/stdout (although, I think Mixer is going to monitor proxies themselves)

proxy/envoy/agent.go
kyessenov: Thanks, will fix permissions in the next iteration.

kyessenov:
I discovered securityContext: runAsUid which lets me get rid of permission stuff.

proxy/envoy/config.go
```diff
@@ -124,83 +112,126 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service, mes
...
 			Port:       port,
 			BindToPort: false,
 		}
 
+		ingressClusterName := fmt.Sprintf("%s%d", IngressClusterPrefix, port)
+		if len(lst.instances) > 0 {
+			clusters = append(clusters, Cluster{
+				Name:             ingressClusterName,
+				Type:             "static",
+				ConnectTimeoutMs: 1000,
```
rshriram: A TODO comment here to make this configurable?

proxy/envoy/config.go
```diff
@@ -124,83 +112,126 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service, mes
...
 		}
+	}
 
+	for port, lst := range ports {
 		listener := Listener{
 			Port:       port,
 			BindToPort: false,
 		}
 
+		ingressClusterName := fmt.Sprintf("%s%d", IngressClusterPrefix, port)
```
rshriram: ingressCluster seems wrong terminology to indicate a single app instance co-located with Envoy. May be "localAppInstance", where instance clearly indicates that its a singleton, local (to imply that its forwarding traffic to the app in same pod).

Also, won't this break in per-node Envoy deployment?

proxy/envoy/config.go
```diff
@@ -124,83 +112,126 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service, mes
...
+		}
+
 		// These two filters operate at different levels of the network stack.
 		// In practice, no port has two protocols used in the same set of instances, but we
 		// should be careful with not stepping on our feet.
-		if protocols[model.ProtocolTCP] {
+		// The order of the filter insertion is important.
+		if lst.instances[model.ProtocolTCP] {
 			listener.Filters = append(listener.Filters, NetworkFilter{
 				Type: "read",
```
rshriram: Are we assuming that the service is listening on only one TCP port?

proxy/envoy/config.go
```diff
@@ -124,83 +112,126 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service, mes
...
 				Name: "tcp_proxy",
 				Config: NetworkFilterConfig{
-					Cluster:    cluster.Name,
+					Cluster:    ingressClusterName,
 					StatPrefix: "ingress_tcp",
 				},
 			})
 		}
```
rshriram: What about egress clusters for TCP traffic?

proxy/envoy/config.go
```diff
@@ -124,83 +112,126 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service, mes
...
+				Domains: []string{"*"},
+				Routes:  []Route{{Cluster: ingressClusterName}},
+			})
+		}
+
+		// TODO: HTTPS protocol ingress and egress configuration
+
+		if len(hosts) > 0 {
 			listener.Filters = append(listener.Filters, NetworkFilter{
 				Type: "read",
```
rshriram: Once again, multi-port services?

rshriram: Its fine if its not there in this iteration. But lets just mark it as a TODO, so that we don't miss it and that everyone knows about the issue.

proxy/envoy/config.go
```diff
@@ -121,83 +112,126 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service, mes
...
 				Config: NetworkFilterConfig{
-					Cluster:    cluster.Name,
+					Cluster:    ingressClusterName,
 					StatPrefix: "ingress_tcp",
 				},
 			})
 		}
 
-		if protocols[model.ProtocolGRPC] || protocols[model.ProtocolHTTP2] || protocols[model.ProtocolHTTP] {
+		// HTTP-level filter is shared between ingress and egress traffic.
```
enricoschiattarella: This comments looks misleading to me. That may or may not be true depending on the destination ports ingress/egress traffic use. So the same filter chain might be shared between multiple ingress services traffic and/or egress or they may all end up having their own independent filter chain if they run on different ports.

enricoschiattarella: We should also add the options (not necessarily now) for the client to do TLS (terminated at Envoy) and/or use PROXY proto. These translate to different listener configs.

proxy/envoy/config.go
kyessenov: there is at most one service instance on a pod listening on a specific _port_.

proxy/envoy/config.go
kyessenov: TODO item left (needs Envoy TCP routing)

proxy/envoy/config.go
kyessenov: Done

proxy/envoy/config.go
kyessenov: Can you explain with an example what is the issue with multi-port service here?

proxy/envoy/config.go
rshriram: I like inbound and outbound. Seems to steer clear of k8s terms while keeping the semantics clear.

proxy/envoy/config.go
rshriram: I think I got confused initially in the absence of code comments. The logic is much clearer now. Thanks for the comments.

PR #44, opened 2017-01-19
Add a dummy kubectl config file for testing

lookuptable:
Currently platform/kube:go_default_test depends on a local config file,
but the file is gitignored. Thus we need to explicitly create that file
before doing `bazel build`. This commit fixes the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/istio/manager/44)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 /bazel-*
 contrib/cfgeditor/content.go
-platform/kube/config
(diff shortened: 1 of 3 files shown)


kyessenov:
You are right, one of the tests requires an access to the cluster. I changed README.md to only build non-test targets which should avoid having to symlink config file.

rshriram:
do we need this PR?

Issue #7, opened 2016-12-13
Add pre-submit checks and CI


kyessenov:
The tests in Manager require an access to a GKE cluster to run the integration test.
That means Jenkins/GKE is probably a better choice for CI than Travis.

sebastienvas:
But you still have some unittest that could easily run on Travis right ?

kyessenov:
Correct. But I'm also using Bazel for running Kubernetes tests. I'd like to have those in CI. So if you have a choice to get with Jenkins or Travis, can we go with Jenkins?

sebastienvas:
The plan is to go with both for now. Travis to prepare the PR for review and Jenkins to do the heavy lifting. We might drop Travis in the future.

kyessenov:
We dropped Travis since we very much depend on kubernetes to exercise code in Manager.
Closed by #64

PR #45, opened 2017-01-19
Update README.md

kyessenov:


README.md
```diff
@@ -15,7 +15,7 @@ If you have a question about the Istio Manager or have a problem using it, pleas
 
 We are using [Bazel 0.4.3](https://bazel.io) to build Istio Manager:
 
-    bazel build //...
+    bazel build //cmd/...
```
lookuptable: ok. Maybe we can create an issue to make sure we have a better solution in future.

kyessenov: yeah, see #7

PR #51, opened 2017-01-21
Use Bazel to build Docker images

kyessenov:
With this change we get deterministic and consistent docker builds for Manager and Proxy.
Try this:
    
    bazel run //docker:runtime

This change is [Reviewable](https://reviewable.io/reviews/istio/manager/51)


proxy/envoy/agent.go
```diff
@@ -44,39 +46,21 @@ type instance struct {
...
+		binary:     binary,
+		configRoot: configRoot,
+		cmdMap:     make(map[*exec.Cmd]instance),
+	}
 }
 
-func configFile(epoch int) string {
-	return fmt.Sprintf(EnvoyConfigPath+EnvoyFileTemplate, epoch)
+func configFile(config string, epoch int) string {
+	return fmt.Sprintf(config+EnvoyFileTemplate, epoch)
```
douglas-reid: any reason to not just have the EnvoyFileTemplate include a %s at the beginning and pass config as an arg here?

proxy/envoy/config.go
```diff
@@ -106,7 +118,9 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service, mes
 
 // buildListeners uses iptables port redirect to route traffic either into the pod or outside the pod
 // to service clusters based on the traffic metadata.
-func buildListeners(instances []*model.ServiceInstance, services []*model.Service) ([]Listener, []Cluster) {
+func buildListeners(instances []*model.ServiceInstance,
+	services []*model.Service,
```
douglas-reid: curious: why spread this out across lines?

proxy/envoy/agent.go
kyessenov: Makes sense. The config is always meant to be a path prefix, so maybe it's easier to enforce that with concatenation?

proxy/envoy/config.go
kyessenov: `lll` barked at me... I'll fix this.

PR #53, opened 2017-01-24
Integration test for proxy + manager

kyessenov:
This change is [Reviewable](https://reviewable.io/reviews/istio/manager/53)


test/integration/http-service.yaml.tmpl
```diff
@@ -42,15 +42,15 @@ spec:
...
       - name: proxy
-        image: gcr.io/istio-test/runtime:experiment
+        image: gcr.io/istio-test/runtime:$TAG
         securityContext:
           runAsUser: 1337
         args:
           - proxy
           - -s
-          - manager:8080
+          - m:8080
```
enricoschiattarella: should this 8080 be replaced with $PORT too ?

kyessenov: Manager port is fixed. I'm checking port cases for actual services (a and b here)

PR #55, opened 2017-01-24
Improvement to test infrastructure

kyessenov:
Added test client, server, better template, better test script, and packaging bazel rules.

This change is [Reviewable](https://reviewable.io/reviews/istio/manager/55)

### DIFF
--- a/docker/BUILD
+++ b/docker/BUILD
@@ -38,7 +38,7 @@ pkg_tar(
 
 docker_build(
     name = "runtime",
-    base = "@docker_ubuntu//:trusty",
+    base = "@docker_ubuntu//:xenial",
     entrypoint = ["/usr/local/bin/manager"],
     repository = "istio",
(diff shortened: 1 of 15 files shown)


sebastienvas:
The check script failed.

docker/BUILD
```diff
@@ -48,3 +61,12 @@ docker_build(
         ":proxy",
     ],
 )
+
+docker_build(
+    name = "app",
+    base = "@docker_ubuntu//:xenial",
+    entrypoint = ["/bin/server"],
```
rshriram: is it possible to run alpine or scratch images?

kyessenov: I'm building on xenial, so it was a path of least resistance in matching Envoy GLIBC versions.
We can spend effort in trying out alpine. It's not a big deal to use larger base as long as we reuse the base layer in docker.

PR #60, opened 2017-01-26
Fixes for HTTP routing

kyessenov:
Addressed the following HTTP routing capabilities:
- multiple HTTP ports in a service
- domain names can be "name" or "name.namespace"
- port is a part of the host name

---
This change is [Reviewable](https://reviewable.io/reviews/istio/manager/60)


model/service.go
```diff
@@ -25,10 +25,15 @@ import (
 type ServiceDiscovery interface {
 	// Services list declarations of all services and their tags
 	Services() []*Service
+
+	// GetService retrieves a service by name if it exists
+	GetService(name, namespace string) (*Service, bool)
```
rshriram: How about making it a bit more generic, such as GetService(name, namespace, tags) ?
Ideally, keeping in line with the ClusterIdentifier in the proto description, we should be using only name, tags as argument. (you would then tuck namespace into name field, there by hiding kube specific namespace concept from the model).

The model is supposed to be independent of the underlying platform. Namespace concept is kubernetes specific. It will break in Mesos/VM cluster or whatever. WDYT?

model/service.go
```diff
@@ -37,12 +42,28 @@ type ServiceDiscovery interface {
...
+
 	// Tags is a set of declared distinct tags for the service
 	Tags []Tag `json:"tags,omitempty"`
+
 	// Ports is a set of declared network service ports
-	Ports []Port `json:"ports,omitempty"`
+	Ports []*Port `json:"ports,omitempty"`
+
+	// Addresses lists service IPv4 addresses
+	Addresses map[string]bool `json:"addresses"`
```
rshriram: A bit confused. Why is this being added now, and why wasn't it added earlier?

model/service_test.go
```diff
@@ -21,25 +21,25 @@ var validServices = map[string]Service{
...
 	"svc:test:prod": {
 		Name:  "svc",
 		Tags:  []Tag{{"prod": ""}},
-		Ports: []Port{{Name: "test"}}},
+		Ports: []*Port{{Name: "test"}}},
 	"svc:http-test": {
 		Name:  "svc",
-		Ports: []Port{{Name: "http-test"}}},
+		Ports: []*Port{{Name: "http-test"}}},
 }
```
rshriram: Perhaps add a test for protocols as well?

platform/kube/controller.go
```diff
@@ -284,30 +286,56 @@ func (c *Controller) Services() []*model.Service {
 	return out
 }
 
+// GetService implements a service catalog operation
+func (c *Controller) GetService(name string, namespace string) (*model.Service, bool) {
```
rshriram: add tags here?

model/service.go
```diff
@@ -67,10 +92,13 @@ type ServiceInstance struct {
 
 // Port represents a network port
 type Port struct {
-	// Port value (may be distinct from the service port with the same name)
-	Port int `json:"port"`
-	// Name of the port classifies ports for a single service, optional
+	// Name of the port classifies ports for a single service, optional only for
+	// a single port service
 	Name string `json:"name,omitempty"`
+
```
rshriram: per offline discussion, this is once again a K8S specific abstraction, thats leaking into the generic Istio model of a service. I don't have a good solution for this at this hour. Pushing it down into platform/kube.go may make the code there ugly, but keeps interfaces clean. Keeping name field here is fine, as long as its a benign entry in the port spec, but there needs to be an API to retrieve ports by Protocol or Port or Name.

model/service.go
kyessenov: I added it in preparation for TCP services for which host field, the basis for HTTP routing, is not available. This adds support for "curl cluster_ip" instead of "curl cluster_name" for HTTP services.

model/service.go
kyessenov: There is `Services` method that is not supposed to do what you want. This is an optimized method to get a service by a name. The resulting service contains a list of tags.

model/service.go
kyessenov: I don't know how to handle multi-port services without names on ports.

model/service_test.go
kyessenov: Ok, WTAL.

platform/kube/controller.go
kyessenov: There is only one service for all service tags, no? v1 and v2 still belong to the same service?

kyessenov:
The current model contains just enough information to be able to configure proxies on Kubernetes. I'd argue the model should be extended if needed to support other platforms. We clearly require a naming scheme to address service instances. The name has to include:
- host name (split into name and namespace, but I'm ok collapsing it back into DNS name)
- port name (I'm ok following DNS SRV convention)
- service tag/cluster, which remains unclear.

rshriram:
Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/istio/manager/60)*

rshriram:
Mesos supports the concept of labels with names for services. Currebt mesos version has support for pods as well. So I don't think the current model is kubernetes specific unless the platform specific abstractions have leaked out. Mesos has dns srv as well. 
We don't need a host name. Just the IP and ports.  It would be good to periodically check against one non kubernetes platform (mesos) to make sure that we are not over specializing for kubernetes.

PR #68, opened 2017-01-31
[wip] Proxy cfg validation & route generation

kyessenov:
This change is [Reviewable](https://reviewable.io/reviews/istio/manager/68)


platform/kube/client.go
```diff
@@ -360,7 +361,7 @@ func modelToKube(km model.KindMap, obj *model.Config) (*Config, error) {
 	return out, nil
 }
 
-func protoToMap(msg proto.Message) (map[string]interface{}, error) {
+func protoToMap(msg proxycf.ProxyConfig) (map[string]interface{}, error) {
```
kyessenov: don't specialize this method - it's meant to be used for all proto messages (there is a descriptor field in each generated Go source)

platform/kube/client.go
```diff
@@ -303,18 +304,18 @@ func kindToAPIName(s string) string {
...
-			return nil, err
-		}
-	}
+	// var status proto.Message
+	// if config.Status != nil {
+	// 	status, err = mapToProto(schema.StatusMessageName, config.Status)
+	// 	if err != nil {
+	// 		return nil, err
+	// 	}
+	// }
```
kyessenov: leave this code as is, you can use a dummy proto for status (like the one in mock_config). I'm pretty sure we will need this later but it doesn't hurt now.

rshriram:
closing PR and splitting it into 2. Please see #80 and #79

Issue #6, opened 2016-12-12
Implement Ingress Controller for Envoy


louiscryan:
As a proof of concept of the abstraction model we should implement standard Ingress support for Envoy. Should not need any additional configuration artifacts beyond what K8S supports today.

kyessenov:
That sounds like a good target. I would add that Manager needs to support a regular proxy config, in addition to Istio Proxy, to fully reproduce an ingress controller. There is a difference in proxy topology for the edge proxy.

elevran:
@louiscryan @kyessenov trying to flesh out the scope, please weigh in so we're on the same page.

This is my understanding of the task, as worded:
Run the manager as an ingress controller to setup Envoy as a standard Ingress controller, only basic k8s Ingress functionality. Later extend to support full Istio functionality.

Since this issue was first written, things have evolved. Is the above still a correct rephrasing of what's needed?

louiscryan:
Yes, that's correct.

elevran:
Merged in #71 #177

zcahana:
Opened a bunch of follow-up issues, so that we can track and prioritize future work on the ingress controller:

- #214: Ingress controller: update ingress resource status
- #215: Ingress controller: support ingress class annotation
- #216: Ingress controller: TLS support
- #217: Ingress controller: support Istio routing rules
- #218: Ingress controller: integrate with Istio mixer
- #219: Ingress controller: evaluate kubernetes GenericController as a base implementation
- #220: Ingress controller: start proxy process regardless of existing ingress resources

zcahana:
Also note #151 which tracks some additional issues relating to ingress rules vs. Istio's rules.

PR #71, opened 2017-02-02
Envoy-based ingress controller

zcahana:
This is a first-pass on Envoy-based ingress controller (addressing #6).
I'm PR'ing this while still WIP to start gathering feedback and set this in the right direction.

What we have so far:
- Ingress controller running in the context of the manager (via `manager ingress` command).
- Extending the `platform/kube/Controller` to sync ingress resources.
- Adding ingress watcher that reconfigures Envoy based on the given ingresses.
- Envoy clusters made up of single upstream (service's DNS name).

What's still missing:
- Represent ingress rules in the Istio model, so that proxy/envoy isn't coupled with k8s objects.
- Support for path regex in ingress (under the limitations of what Envoy supports)
- Constructs envoy clusters using SDS + pod endpoints.
- Update ingress status with the proxy's "public" IP address (node IP?).
- TLS termination.
- Testing.


rshriram:
@zcahana could you please sign the cla?

zcahana:
@rshriram I applied to join the ibm-contributors group, waiting for it to get approved.

cmd/manager/main.go
```diff
@@ -122,6 +138,18 @@ func init() {
 	proxyCmd.PersistentFlags().StringVarP(&flags.proxy.MixerAddress, "mixer", "m", "",
 		"Mixer DNS address (or empty to disable Mixer)")
 	rootCmd.AddCommand(proxyCmd)
+
+	// TODO: Share these flags with proxyCmd (e.g., make both children of "manager envoy" command)?
+	ingressCommand.PersistentFlags().StringVarP(&flags.proxy.DiscoveryAddress, "sds", "s", "manager:8080",
+		"Discovery service DNS address")
+	ingressCommand.PersistentFlags().IntVarP(&flags.proxy.AdminPort, "admin_port", "a", 5000,
+		"Envoy admin port")
+	ingressCommand.PersistentFlags().StringVarP(&flags.proxy.BinaryPath, "envoy_path", "b", "/envoy_esp",
```
rshriram: Envoy's admin port is of little use to end user.  It is a readonly port. I don't even think this needs to be exposed.
Could you also add a TODO here saying that the hardcoded use of envoy must be changed to something more generic? e.g., 
``//TODO: Change to proxy_type: envoy|nginx|haproxy, and then make the rest (SDS, etc) as sub flags```

model/BUILD
```diff
@@ -4,6 +4,7 @@ go_library(
     name = "go_default_library",
     srcs = [
         "controller.go",
+        "ingress.go",
         "registry.go",
```
rshriram: Please remove ingress from the model. Model is platform agnostic representation of resources. Mesos may not even have ingress.

model/controller.go
```diff
@@ -34,6 +36,9 @@ type Controller interface {
 	// for a service.
 	AppendInstanceHandler(f func(*ServiceInstance, Event)) error
 
+	// AppendIngressHandler notifies about changes to ingress resources
+	AppendIngressHandler(f func(*v1beta1.Ingress, Event)) error
+
```
rshriram: May be add a type to the Config struct to detect if its an ingress or not. This way, ingress becomes kube specific.

model/ingress.go
```diff
@@ -0,0 +1,23 @@
...
+
+package model
+
+import "k8s.io/client-go/pkg/apis/extensions/v1beta1"
+
+// IngressRegistry provides access to ingress rules.
+type IngressRegistry interface {
+	// ListIngresses all ingress rules currently defined.
+	ListIngresses() []*v1beta1.Ingress
+}
```
rshriram: Please (re)move was you see fit.

proxy/envoy/ingress.go
```diff
@@ -0,0 +1,244 @@
...
+
+	"istio.io/manager/model"
+
+	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
+)
+
+type ingressWatcher struct {
+	agent   Agent
+	mesh    *MeshConfig
+	ingress model.IngressRegistry
```
rshriram: one favor: could you add a place holder or a TODO, so that #32 can be incorporated here easily as well?

proxy/envoy/ingress.go
```diff
@@ -0,0 +1,244 @@
...
+		return
+	}
+
+	// Add a short delay to de-risk a potential race condition in envoy hot reload code.
+	// The condition occurs when the active Envoy instance terminates in the middle of
+	// the Reload() function.
+	time.Sleep(256 * time.Millisecond)
+}
+
+func (w *ingressWatcher) generateConfig() (*Config, error) {
```
rshriram: is there a way to de-duplicate this functionality by moving code out of proxy/envoy/watcher.go(?) that generates Envoy configuration, and reusing it in both places? After all, that code just expects a list of services, pods, and rules. in the absence of rules, it behaves just like a user-space kube proxy (except operating on per-pod basis). Whatever is missing will not be configured.

rshriram: On second thought, lets add a TODO here as a reminder that we need to refactor this code. We could do that in a later PR.
Kind of similar to the IdentityProvider stuff in A8, in principle. What do you think?

proxy/envoy/ingress.go
```diff
@@ -0,0 +1,244 @@
...
+
+		// Ensure "catch-all" path appears last on the ordered slice of routes
+		if rLen := len(vhost.Routes); rLen > 1 && vhost.Routes[rLen-2].Prefix == "/" {
+			catchall := vhost.Routes[rLen-2]
+			vhost.Routes[rLen-2] = vhost.Routes[rLen-1]
+			vhost.Routes[rLen-1] = catchall
+		}
+
+		// TODO: Validate no duplicate path/routes within this virtualhost
+		// (e.g., due to 2 default backends defined in 2 ingresses in different namespaces)
```
rshriram: silly question: is ingress common to all namespaces? or specific to a namespace? IF latter, we don't need this right?

proxy/envoy/ingress.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	vhosts := vhostMapToSlice(vhostsMap)
+	clusters := clusterMapToSlice(clustersMap)
+
+	listeners := []Listener{}
+	listeners = append(listeners, Listener{
+		Port:           80,
+		BindToPort:     true,
+		UseOriginalDst: true,
```
rshriram: Set to false please. This does not apply to ingress. Its only for internal envoy instances.

proxy/envoy/ingress.go
```diff
@@ -0,0 +1,244 @@
...
+		clusterName := clusterName(namespace, backend)
+		cluster, ok := clustersMap[clusterName]
+		if !ok {
+			clustersMap[clusterName] = Cluster{
+				Name:             clusterName,
+				ServiceName:      clusterName,
+				Type:             "strict_dns", // TODO: Change to SDS
+				LbType:           DefaultLbType,
+				ConnectTimeoutMs: DefaultTimeoutMs,
+				Hosts:            clusterHosts(namespace, backend),
```
rshriram: don't we have to set the cluster `"features": "http2"`, since we would have Envoy backends? If this is for istio only, then by default (and for first cut) I think we should set to features to "http2", since all http communication happens over Envoy to Envoy. If not specified, Envoy falls back to http/1.1, nullifying a lot of benefits that we get out of http/2.

proxy/envoy/ingress.go
```diff
@@ -0,0 +1,244 @@
...
+		},
+	})
+
+	// TODO: 443 listener + TLS support
+
+	return &Config{
+		Listeners: listeners,
+		ClusterManager: ClusterManager{
+			Clusters: clusters,
+		},
```
rshriram: please set the `"local_cluster_name": "ingress"`. Local cluster name is needed for a variety of features in envoy (CDS, tracing, possibly RDS, etc.).

proxy/envoy/ingress.go
zcahana: Ingresses are defined within namespaces, but are normally realized by a single namespace-neutral ingress controller, which serves incoming traffic to all namespaces. This TODO refers to the need to make sure there are no collision between ingresses in different namespaces, which might yield invalid envoy configuration (e.g., a virtualhost with 2 similar routes, or 2 prefix-similar routes, etc).

proxy/envoy/ingress.go
zcahana: I think this is Istio-only for now. Will set "http2".

proxy/envoy/ingress.go
zcahana: Yes, there's some de-duplication and code reuse that can be done. Will be refactored in next iterations on this.

proxy/envoy/ingress.go
zcahana: Not sure I understand that comment...

model/BUILD
zcahana: So, we need something in the model (`RoutingRule`?) that the envoy ingress watcher can consume, and the kube ingresses converted to. Is there already any work in that direction?

model/controller.go
zcahana: I'm not sure I understand how the `Config` is designed to be worked with. Any reference or explanation can help.

proxy/envoy/ingress.go
```diff
@@ -0,0 +1,244 @@
...
+
+func (w *ingressWatcher) generateConfig() (*Config, error) {
+	// Map a domain name to a VirtualHost
+	vhostsMap := make(map[string]VirtualHost)
+
+	// Map a cluster name to a Cluster
+	clustersMap := make(map[string]Cluster)
+
+	// Function to add a rule for a particular host, path and backend.
+	// This will later be invoked for each ingress rule, as well as ingress default backend
```
kyessenov: We need to provide a service for a 404 backend. The service can be implemented by Envoy or default-http-backend from kubernetes.

proxy/envoy/ingress.go
kyessenov: I'll try to refactor current config to make it more re-usable. It handles things like enabling http2.

proxy/envoy/ingress.go
```diff
@@ -0,0 +1,244 @@
...
+			vhost.Routes[rLen-2] = vhost.Routes[rLen-1]
+			vhost.Routes[rLen-1] = catchall
+		}
+
+		// TODO: Validate no duplicate path/routes within this virtualhost
+		// (e.g., due to 2 default backends defined in 2 ingresses in different namespaces)
+	}
+
+	// Iterate on all ingresses, and create corresponding rule
+	for _, ingress := range w.ingress.ListIngresses() {
```
kyessenov: Please make the order of rules or ingresses canonical. Otherwise, the watcher will keep reloading Envoy due to some permutation in the config file.

kyessenov:
More comments for the TODOs:
- Constructs envoy clusters using SDS + pod endpoints.
Yes, we should avoid using kube DNS since we have SDS. Don't worry about CDS for now, it's easy to refactor for that later.
- Update ingress status with the proxy's "public" IP address (node IP?).
GCE LB ingress does this because it manages firewall rules and IP allocation. NGINX ingress does not AFAIK and you have to create a separate service of type `LoadBalancerIP` pointing to NGINX deployment. Let's do what NGINX ingress does.
- TLS termination.
Don't worry about it. We will do one big swoop to enable TLS termination.
- Testing.
Take a look at `test/integration/driver.go`. I think it has all the pieces to reuse.

model/BUILD
rshriram: +1. Model package is completely platform independent.

model/controller.go
zcahana: The ingress config seems to be a subset of the model's [`RouteRule`](https://github.com/istio/manager/blob/master/model/proxy/alphav1/config/cfg.pb.go#L153) (that's what you mean by `ProxyConfig`, right)? So, we can have the controller convert ingresses to such rules, and have the proxy watcher subscribe to these.

proxy/envoy/ingress.go
zcahana: I'd suggest to postpone merging the two watchers until we get a clearer picture of how much logic they share. At least for now, they are generating quite a different envoy configuration. The sidecar watcher shouldn't care / handle ingress-based configurations, whereas the ingress watcher shouldn't allow non-ingressed traffic.

kyessenov: Sounds good to me. We should share code for config generation but the controllers themselves should be independent of one another. Let's see how it works out and extract proper module interfaces later.

model/controller.go
rshriram: Yes. Please take a look at the Generate function in proxy/envoy/config.go. You can provide an empty array for upstreams and rules. It should give you an envoy configuration for services as hostnames and / as the default prefix. Along with SDS functionality. In next iteration (or if you want to hack the utility functions), you could generate Envoy route entries under a virtual host with specific path or prefixes. 

Bottom line. There is opportunity for code reuse. The config.go takes care of generating all the right listener blocks (with use_orig_dst, etc).

proxy/envoy/ingress.go
zcahana: Reading [here](https://lyft.github.io/envoy/docs/configuration/cluster_manager/cluster_manager.html), it seems that `local_cluster_name` is only needed for enabling zone aware routing. This is probably out of the scope of this PR.

cmd/manager/main.go
```diff
@@ -122,6 +138,18 @@ func init() {
...
+		"Envoy binary location")
+	ingressCommand.PersistentFlags().StringVarP(&flags.proxy.ConfigPath, "config_path", "e", "/etc/envoy",
+		"Envoy config root location")
+	rootCmd.AddCommand(ingressCommand)
```
zcahana: Yes, but should ingress be a subcommand of proxy? In essence, these are two different modes of operation (in-cluster service sidecar proxy vs. edge ingress proxy).

Most (all?) of the commonalities stem from the fact that both commands manage envoy, so maybe both should be a child command of an "envoy" command?

proxy/envoy/ingress.go
zcahana: This functionality is already exposed by k8s' ingresses, which allows the user to specify a default backend (service) for requests that don't match any ingress rule. This is implemented as an Envoy VirtualHost with a wildcard domain ("*"). If the user doesn't specify a default backend in its ingresses, then envoy we don't produce such wildcard VirtualHost, and fallback to Envoy's own 404.

zcahana:
I've rewritten the ingress controller implementation such that Ingress API objects are converted to Istio model rules (currently, same `config.RouteRule`, but with a different kind - `IngressRule`), which are then transformed to Envoy configuration. There's still only a little re-use of existing Envoy configuration code. This can be refactored later when the package stabilizes and the common parts become clearer.

Note that due to the major changes that already went to master, I've rebased my branch and restructured the commits so that they're more reviewable. The commits from the first round are thus gone :).

cmd/manager/main.go
kyessenov: Yeah, good point. I think we need a shared root, and then run "sidecar", "ingress", and "egress" as a subcommand. Actually, the flag parameter need to be turned into ProxyMeshConfig in the proto, but that's for later.

zcahana:
Jenkins' presubmit build quickly fails with the following error ([link](https://istio-testing.appspot.com/job/manager/job/presubmit/270/console)):
```
+ go get ./...
/var/jenkins_home/jobs/manager/jobs/presubmit/workspace/src/main@tmp/durable-3d3f8166/script.sh: 2: /var/jenkins_home/jobs/manager/jobs/presubmit/workspace/src/main@tmp/durable-3d3f8166/script.sh: go: not found
[Pipeline] }
[Pipeline] // dir
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
ERROR: script returned exit code 127
Finished: FAILURE
```

Not sure what's causing this... any chance it's some recent changes to istio-testing repo? @sebastienvas

sebastienvas:
Please rebase your PR.

zcahana:
thanks @sebastienvas , build now passes.

platform/kube/conversion.go
```diff
@@ -147,3 +155,111 @@ func modelToKube(km model.KindMap, k *model.Key, v proto.Message) (*Config, erro
...
+		Route: []*config.DestinationWeight{
+			{
+				Destination: destination,
+				Weight:      100,
+
+				// A temporary measure to communicate the destination service's port
+				// to the proxy configuration generator. This can be improved by using
+				// a dedicated model object for IngressRule (instead of reusing RouteRule),
+				// which exposes the necessary target port field within the "Route" field.
+				Tags: map[string]string{
```
rshriram: I don't think we do that anymore.

rshriram: I mean refer to ports in the route rule

zcahana: This is necessary for envoy routes which are generated from ingresses - we need to route traffic to the cluster corresponding with the service:port defined by the user in the ingress.

zcahana:
There's at least one bugfix on the way, and potentially another one. I'm currently running/debugging this locally.

platform/kube/conversion.go
zcahana: Actually, looking further into this, cluster names include the port name, and not the port number.
So, what happens when there's no port name specified?

kyessenov: That can only happen if there is only one port, and the upstream cluster becomes just the service name ("a.svc....") or service-tag combination ("a.svc...::env=prod")

kyessenov: Can you add a TODO item here? This needs to be in the next iteration of routing rules, we've been assuming rules apply per port, but ingress needs an explicit port. Moreover, this is a port name, so this has some implication for the route rule model.

platform/kube/controller.go
```diff
@@ -298,6 +399,22 @@ func (c *Controller) List(kind, namespace string) (map[model.Key]proto.Message,
 	return out, errs
 }
 
+func (c *Controller) listIngresses(kind, namespace string) (map[model.Key]proto.Message, error) {
```
zcahana: What is it that's not needed here? The `listIngresses` method gets called from `Controller.List(kind, namespace)` according to the actual kind. In practice, this is being used by the envoy's ingress watcher.

platform/kube/conversion.go
zcahana: This seems to be captured in #151.

zcahana:
With the latest round of commits, I've verified that basic ingress function works (i.e., ingress controller forwards ingress traffic to correct backend service according to defined routes). In practice, the backend service's reverse proxy is still blocking the request from hitting the actual service (due to #150), so this is still not usable end-to-end, but probably solid enough for first-pass --> merging.

platform/kube/controller.go
kyessenov: Oops my bad, I was referring to "kind" parameter.

zcahana: Well, you're right that the "kind" parameter isn't really needed here. I left it for uniformity with the other listX methods.

elevran:
Together with #177 - fixes #6

Issue #65, opened 2017-01-28
Proxy injection without k8s admission controller


ayj:
istio/manager/issues/57 is tracking long-term support for transparent proxy injection for per-pod case via k8s admission controller. We need an interim solution that does not require a special k8s build for managed k8s, e.g. GKE. This could simply consist of manually inserting additional pre-prepared containers into each pod spec, e.g. istio-init-container, istio-proxy-container.  There is precedence for this in other sidecar implementation, e.g. [linkerd](https://linkerd.io/getting-started/k8s/).

rshriram:
I think if we supply a standard YAML snippet that people have to take and stick into their k8s configs, it should suffice. We are not using ConfigMaps (cc @kyessenov ).. The only glitch here is TLS certs. I am not sure what is being worked out, but in the interim, I think there is #52 where the idea is that we will pull the TLS certs out of k8s secrets. The user has to tell us the name of the secret in the pod spec for sidecar.

kyessenov:
We have a basic template already:
https://github.com/istio/manager/blob/master/test/integration/http-service.yaml.tmpl
We just need to clean it up, and add reasonable defaults.
For TLS, there's Istio auth design under review, but we should help with simple HTTPS termination use cases that don't need managed certs and CAs.

ayj:
My plan was to do the following:

1) Create/update images as necessary for the init-container (w/embedded iptables setup vs. explicit spec command), proxy, and proxy agent. 
2) Update documentation to include snippet(s) that user needs to add to the specs (README.md or maybe create another file under a doc/ directory)
3) Look into simple HTTPS termination and update (1) and (2) accordingly in follow-up PR.

kyessenov:
Sounds good to me. iptables set-up can be improved to use cgroups instead of PID owner and also skip localhost to localhost traffic, if you want you can look into that.

ayj:
Will do. I took #36 since I'd be looking at the iptables stuff while doing #65.

ayj:
Addressed by #73

PR #73, opened 2017-02-02
Move proxy iptable configuration into istio init image

ayj:
Partially addresses #36 and #65


docker/prepare_proxy.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
kyessenov: Can you add `set -e` to make sure initialization fails if iptables fails for some reason?

PR #82, opened 2017-02-03
Introduce routing rule as TPR

kyessenov:
Re-work the glue for TPR since discovering an issue with multiple TPRs.
Changed the storage scheme so that the name encodes the kind (e.g. proxy rule a is called proxy-rule-a in kubernetes).
Included proto definitions and generated code for proxy-related config for faster iteration.
Will add a CLI to accommodate put/get operations for TPRs with validation in a follow-up PR.


model/proxy/alphav1/config/cfg.proto
```diff
@@ -0,0 +1,483 @@
...
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+import "google/protobuf/wrappers.proto";
+
+package istio.proxy.v1alpha.config;
+option go_package = "config";
```
rshriram: I didn't know that this option was possible! Can we change the name to routingcfg ?

model/proxy/alphav1/config/cfg.proto
```diff
@@ -0,0 +1,483 @@
...
+// on versions above). The proxy would choose the upstream cluster based on
+// various routing rules.
+
+// Applications address only the upstream service using without knowledge
+// of individual upstream clusters. The actual choice of the cluster is
+// determined by the proxy, enabling the application code to decouple
+// itself from the evolution of dependent services.
+
+// Proxy level global configurations go here
+message ProxyConfig {
```
rshriram: And this gets renamed to RoutingConfig?

model/validation.go
```diff
@@ -44,9 +45,9 @@ func IsDNS1123Label(value string) bool {
 }
 
 // Validate confirms that the names in the configuration key are appropriate
-func (k *ConfigKey) Validate() error {
+func (k *Key) Validate() error {
 	var errs error
-	if !kindRegexp.MatchString(k.Kind) {
+	if !IsDNS1123Label(k.Kind) {
```
rshriram: Doesn't this mean we need atleast two parts in the kind name ? route-rule.istio.io? If so the constants above need to be updated.

model/proxy/alphav1/config/cfg.proto
rshriram: On second thought, all mixer protos should also be changed to this config root package. Then it might be streamlined.

model/proxy/alphav1/config/cfg.proto
rshriram: Unless we decide to add more stuff into proxy config, such as envoy userid, etc.

platform/kube/kube_test.go
```diff
@@ -245,6 +240,48 @@ func TestServices(t *testing.T) {
...
 }
 
+func TestProxyConfig(t *testing.T) {
+	cl := makeClient(t)
+	ns := makeNamespace(cl.client, t)
+	defer deleteNamespace(cl.client, ns)
+
+	rule := &proxyconfig.RouteRule{
+		RouteRule: &proxyconfig.RouteRule_Http{
+			Http: &proxyconfig.HttpRouteRule{
```
rshriram: Please add dst_service here. We may need to do immediate validation in tpr code to see if the user provided service is valid or invalid

rshriram:
I signed it!

PR #86, opened 2017-02-03
Fix a bug with listing across all namespaces in a unit test

kyessenov:


model/registry.go
```diff
@@ -96,10 +96,10 @@ type IstioRegistry struct {
 	Registry
 }
 
-// RouteRules lists all rules
-func (i *IstioRegistry) RouteRules() []*proxyconfig.RouteRule {
+// RouteRules lists all rules in a namespace (or all rules if namespace is "")
```
rshriram: Don't we also need a UpstreamClusters() function (and similar in the tests as well?)

PR #88, opened 2017-02-03
Introduce a CLI for configuration registry

kyessenov:
Example:  
```
> cat example
dst_service_name: test
> cat example | bazel-bin/cmd/manager/manager config put route-rule  test  -c ~/.kube/config 
> manager config list route-rule -c ~/.kube/config 
Kind route-rule, Name: test, Namespace: default
======
dstservicename: test
routerule: null


```

### DIFF
--- a/model/registry.go
+++ b/model/registry.go
@@ -15,11 +15,7 @@
 package model
 
 import (
-	"encoding/json"
-	"reflect"
-
 	"github.com/golang/glog"
-	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
 
 	proxyconfig "istio.io/manager/model/proxy/alphav1/config"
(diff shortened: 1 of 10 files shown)


rshriram:
dst_Service_name has changed to destination

model/registry.go
```diff
@@ -65,6 +70,44 @@ type ProtoSchema struct {
...
+	pb := reflect.New(pbt.Elem()).Interface().(proto.Message)
+	err = jsonpb.UnmarshalString(string(str), pb)
+	if err != nil {
+		return nil, err
+	}
+
+	return pb, nil
+}
+
 const (
```
rshriram: Move these to to converter.go perhaps?

platform/kube/client.go
```diff
@@ -232,7 +232,8 @@ func (cl *Client) Get(key model.Key) (proto.Message, bool) {
 		return nil, false
 	}
 
-	out, err := mapToProto(cl.mapping[key.Kind].MessageName, config.Spec)
+	kind := cl.mapping[key.Kind]
+	out, err := kind.FromJSONMap(config.Spec)
 	if err != nil {
```
rshriram: I think we can definitely move those two to conversion.go

kyessenov: We need protobuf specific conversion stuff in `model/conversion.go`. `kube/conversion.go` is for k8s

model/registry.go
kyessenov: Yeah

kyessenov:
These examples should work:
```
bazel run //cmd/manager -- config list route-rule
echo destination: test | bazel-bin/cmd/manager/manager config put route-rule test
bazel run //cmd/manager -- config get route-rule test
```

PR #95, opened 2017-02-07
WIP: Add tls cert config (issue/52)

ayj:
TODO
* Create k8s secret to distribute certs. It wasn't clear to me yet whether to make this part of "Istio Manager" or "Istio Manager service". 

* Update proxy config generation to include certs. Assume that service port assignment is unique within the cluster  and only include listener ssl_context for non-local services. Otherwise, if service ports are reused, proxy listener expects TLS termination from local service.


rshriram:
What's the difference between istio manager and istio manager service ? Do you mean the proxy manager and the service that manages all proxies? 
If so, it goes into the local proxy manager imo. And we pull the certificates from the secret store. So users are free to use same certificate across all services. For this to happen, we need to ask users to specify the secret names in the proxy container config. And this seems a bit clunky... any other alternatives?

ayj:
Ignore my comment about difference between "Istio Manager" and "Istio Manager Service". I missed the spf13/cobra command inheritance and assumed that the rootCmd was a separate command (its not, its common between server and agent). 

Agreed that asking user to manually specify secrets is clunky, but we have the same problem with the init-container and proxy injection. Short term I don't see an alternative if we want to distribute certs via k8s secrets. We could pre-process k8s yaml files on the client via some helper script before submitting to API server, but it's arguable whether this provides a better UX via manually updating the spec.

Longer term this problem could hopefully be addressed some combination [Pod Injection Policy](https://github.com/kubernetes/community/pull/254) and [Proposal: Extensible Admission Control](https://github.com/kubernetes/community/pull/132),

ayj:
cc @lookuptable

sebastienvas:
Please rebase your PR.

PR #96, opened 2017-02-07
Header based routing.

ijsnellf:
Adds support for routing based on match headers.

### DIFF
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -393,6 +393,8 @@ func buildRoutes(svc *model.Service, rulesMap map[string][]*config.RouteRule,
 			route := Route{}
 			if httpRule != nil {
 				match := httpRule.GetMatch()
+
+				route.Headers = buildHeaders(match.GetHeaders())
 				route.Path, route.Prefix = buildPathAndPrefix(match)
 				if httpRule.WeightedClusters != nil {
 					route.WeightedClusters = buildWeightedClusters(httpRule.WeightedClusters,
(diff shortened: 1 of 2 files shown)


rshriram:
Gofmt errors.

proxy/envoy/config.go
```diff
@@ -411,6 +413,37 @@ func buildRoutes(svc *model.Service, rulesMap map[string][]*config.RouteRule,
 	return routes
 }
 
+func buildRouteHeaders(headerMatches map[string]*config.StringMatch) []Header {
```
rshriram: Rename to buildHeaderMatches and the other one to buildHeaderMatch as it can be reused in fault filters as well.

proxy/envoy/config.go
```diff
@@ -411,6 +413,37 @@ func buildRoutes(svc *model.Service, rulesMap map[string][]*config.RouteRule,
...
+	for name, stringMatch := range headerMatches {
+		header := buildRouteHeader(name, stringMatch)
+		headers = append(headers, header)
+	}
+	sort.Sort(HeadersByNameValue(headers)) // Guaranteed ordering.
+	return headers
+}
+
+func buildRouteHeader(name string, match *config.StringMatch) Header {
+	value := ""
```
ayj: It looks like `regex` should default to false? If so, you can take advantage of default values for both string and boolean types, e.g.

var value string
var regex bool

ijsnellf: You are right, that is supposed to be `false`.

proxy/envoy/config.go
ijsnellf: Done.

proxy/envoy/config.go
ijsnellf: Done.

PR #97, opened 2017-02-07
use http2 for envoy to envoy communication

rshriram:


rshriram:
Dont merge. Need to differentiate between egress services and internal services, and generate the http2 block accordingly.

proxy/envoy/config.go
```diff
@@ -549,8 +549,7 @@ func buildClusters(services []*model.Service) []Cluster {
 				LbType:           DefaultLbType,
 				ConnectTimeoutMs: DefaultTimeoutMs,
 			}
-			if port.Protocol == model.ProtocolGRPC ||
-				port.Protocol == model.ProtocolHTTP2 {
+			if port.Protocol != model.ProtocolTCP && port.Protocol != model.ProtocolUDP {
```
kyessenov: Please do not use != here. We'll be extending the set of protocols, and a check like this will cause unexpected failures.

rshriram: Agreed. But i think this code will still not work. Because we cannot distinguish between egress service clusters and non-egress clusters. For clusters inside k8s/mesos, i.e. within istio deployment, we have Envoy end to end, i.e., we can do http2 inside istio. But when going out, we cannot assume anything. So, we need to stick to http1.

kyessenov: Ok, we need to introduce additional properties to services first.

rshriram:
Closed in favor of #100

PR #98, opened 2017-02-07
fault injection filter for envoy config

rshriram:
Reads fault rules from the routing rule spec and sets up fault injection filters in Envoy.

### DIFF
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -455,11 +455,7 @@ func buildFaultFilter(internalUpstreamName string, faultRule *config.HttpFaultIn
 
 // buildAbortConfig builds the envoy config related to abort spec in a fault filter
 func buildAbortConfig(abortRule *config.HttpFaultInjection_Abort) *AbortFilter {
-	if abortRule == nil {
-		return nil
-	}
-	if abortRule.GetHttpStatus() == 0 {
-		// We do not support gRPC and http2 errors now.
+	if abortRule == nil || abortRule.GetHttpStatus() == 0 {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


proxy/envoy/config.go
```diff
@@ -410,7 +421,69 @@ func buildRoutes(svc *model.Service, rulesMap map[string][]*config.RouteRule,
...
+
+	return &AbortFilter{
+		Percent:    int(abortRule.Percent),
+		HTTPStatus: int(abortRule.GetHttpStatus()),
+	}
+}
+
+// buildDelayConfig builds the envoy config related to delay spec in a fault filter
+func buildDelayConfig(delayRule *config.HttpFaultInjection_Delay) *DelayFilter {
+	if delayRule == nil {
```
ijsnellf: Could make this more compact:
```golang
if delayRule == nil || delayRule.GetFixedDelay() == nil {
   return nil
}
```

proxy/envoy/config.go
```diff
@@ -410,7 +421,69 @@ func buildRoutes(svc *model.Service, rulesMap map[string][]*config.RouteRule,
...
+	if abortRule == nil {
+		return nil
+	}
+	if abortRule.GetHttpStatus() == 0 {
+		// We do not support gRPC and http2 errors now.
+		return nil
+	}
+
+	return &AbortFilter{
+		Percent:    int(abortRule.Percent),
```
ijsnellf: `abortRule.Percent` is a float in the range [0,100] not [0,1]?

rshriram:
So this simple test is not so simple, as we discussed earlier. Its an integration test framework with TPRs in place. And my plan was to do that as the next PR, once I manage to dis-entagle the test framework from driver.go and figure out the format for the TPRs.

So far, we do not provide *any* TPRs to istio. We just leverage existing service definitions in K8S. What we need to do is to define a new test application (not just a->b,t), e.g., a->b(v1),b(v2) or something.. And then submit TPRs and then validate.

proxy/envoy/config.go
rshriram: Yes, its float in range 0-100.

proxy/envoy/config.go
rshriram: done

proxy/envoy/config.go
kyessenov: please test it with a float and an int value. we do several translations and JS is funky about numbers.

PR #99, opened 2017-02-08
Wire service tags through SDS

kyessenov:


model/service.go
```diff
@@ -57,6 +57,33 @@ type Service struct {
...
+	return true
+}
+
+// TagList is a set of tags
+type TagList []Tag
+
+// Matches returns true if the input tag is a super set of one of the tags in the list
+// or if the tag list if empty
+func (tags TagList) Matches(that Tag) bool {
+	if len(tags) == 0 {
```
rshriram: This is an odd semantics for a matches function (returning true when one side is empty). While it may simplify code, it might make it hard to understand unless one remembers that empty can match with a non empty set. 

I suggest renaming. Like subsetOf here as well.

PR #100, opened 2017-02-08
Auto-upgrade HTTP1.1 to HTTP2

kyessenov:


rshriram:
Looking at this again, I don't even think we need this PR or my older PR #97 . Because neither of them change the envoy cluster configuration at all. (ultimately, its going to be http2 when the service proto is gRPC or Http2. Else no cluster.feature property should be defined).

e2e still passes without any issues. I don't understand where the bug #101 comes from.

rshriram:
can we close this? Its more of a bug isnt it?

PR #105, opened 2017-02-09
Routing proto revision

kyessenov:
Goal: 
- remove all mentions of clusters, tags, "source"
- align with existing service model.

### DIFF
--- a/model/proxy/alphav1/config/cfg.proto
+++ b/model/proxy/alphav1/config/cfg.proto
@@ -23,8 +23,8 @@ option go_package = "config";
 // Glossary & concepts
 //
 // Service is a unit of an application with a unique name that other services
-// can refer to the functionality being called. Service instances are
-// pods/VMs/containers that comprise the service.
+// use to refer to the functionality being called. Service instances are
+// pods/VMs/containers that implement the service.
 //
 // Service versions - In a continuous deployment scenario, for a given service,
 // there can be multiple sets of instances running potentially different
--- a/model/proxy/alphav1/config/cfg.proto
+++ b/model/proxy/alphav1/config/cfg.proto
@@ -188,12 +188,11 @@ message L4MatchAttributes {
 
 // Describes how to matches a given string (exact match, prefix-based match
 // or posix style regex based match). Match is case-sensitive.
+// It is a validation error to supply a regex for a proxy that does not support it.
 message StringMatch {
   oneof match_type {
     string exact = 1;
     string prefix = 2;
-    // posix style regex match. Throw error if user's chosen proxy does not
-    // support posix regex.
     string regex = 3;
   }
 }
@@ -266,7 +265,7 @@ message CircuitBreaker {
     // format seconds.nanoseconds
     double reset_timeout_seconds = 3;
 
-    // Maximum number of connections to a backend
+    // Maximum number of connections to a backend.
     int32 max_connections = 4;
 
     // Maximum number of pending requests to a backend.
@@ -278,7 +277,7 @@ message CircuitBreaker {
     // Minimum time the circuit will be closed.
     string sleep_window = 7;
 
-    //Number of 5XX errors before circuit is opened.
+    // Number of 5XX errors before circuit is opened.
     int32 http_consecutive_errors = 8;
 
     // Interval for checking state of hystrix circuit.
--- a/model/proxy/alphav1/config/cfg.proto
+++ b/model/proxy/alphav1/config/cfg.proto
@@ -119,31 +119,14 @@ message RouteRule {
   // Examples: "my-service.default.svc.cluster.local", "svc.cluster.local".
   string destination = 1;
 
+  // Match condition is optional.
+  // By default, the rule applies to HTTP traffic only.
+  MatchCondition match = 2;
+
   // Each routing rule is associated with one or more service version destinations
   // (see glossary in beginning of document). Weights associated with the service
   // version determine the proportion of traffic it receives.
-  repeated DestinationWeight route = 2;
-
-  // Identifies the service initiating a connection or a request by its name.
-  string source = 3;
-
-  // Identifies the source service version. The identifier is interpreted
-  // by the platform to match a service version for the source service.
-  //
-  // N.B. The map is used instead of pstruct due to lack of serialization support
-  // in golang protobuf library (see https://github.com/golang/protobuf/pull/208)
-  map<string, string> source_version = 4;
-
-  // Set of layer 4 match conditions based on the IP ranges
-  L4MatchAttributes tcp = 5;
-
-  // Set of layer 4 match conditions based on the IP ranges
-  L4MatchAttributes udp = 6;
-
-  // Set of HTTP match conditions based on HTTP/1.1, HTTP/2, GRPC request metadata,
-  // such as "uri", "scheme", "authority".
-  // The header keys are case-insensitive.
-  map<string, StringMatch> http = 7;
+  repeated DestinationWeight route = 3;
 
   // Precedence is used to disambiguate the order of application of rules for
   // the same destination service. A higher number takes priority.  If not
@@ -152,7 +135,33 @@ message RouteRule {
   //
   // Rule precedence can be used to order routes for a single destination, as
   // well as override existing routes with the same attribute condition.
-  int32 precedence = 8;
+  int32 precedence = 4;
+}
+
+// Match condition selects traffic for routing application.
+// The condition provides distinct set of conditions for each protocol with the
+// intention that conditions apply only to the service ports that match the protocol.
+message MatchCondition {
+  // Identifies the service initiating a connection or a request by its name.
+  string source = 1;
+
+  // Identifies the source service version. The identifier is interpreted
+  // by the platform to match a service version for the source service.
+  //
+  // N.B. The map is used instead of pstruct due to lack of serialization support
+  // in golang protobuf library (see https://github.com/golang/protobuf/pull/208)
+  map<string, string> source_version = 2;
+
+  // Set of layer 4 match conditions based on the IP ranges
+  L4MatchAttributes tcp = 3;
+
+  // Set of layer 4 match conditions based on the IP ranges
+  L4MatchAttributes udp = 4;
+
+  // Set of HTTP match conditions based on HTTP/1.1, HTTP/2, GRPC request metadata,
+  // such as "uri", "scheme", "authority".
+  // The header keys are case-insensitive.
+  map<string, StringMatch> http = 5;
 }
 
 // Each routing rule is associated with one or more service versions (see
--- a/model/proxy/alphav1/config/cfg.proto
+++ b/model/proxy/alphav1/config/cfg.proto
@@ -78,14 +78,14 @@ message Destination {
   // Load balancing policy
   LoadBalancing load_balancing = 3;
 
+  // Circuit breaker policy
+  CircuitBreaker circuit_breaker = 4;
+
   // Timeout policy. N.B. Applicable to the default service version only.
-  Timeout timeout = 4;
+  HTTPTimeout http_timeout = 5;
 
   // Retry policy. N.B. Applicable to the default service version only.
-  Retry retry = 5;
-
-  // Circuit breaker policy
-  CircuitBreaker circuit_breaker = 6;
+  HTTPRetry http_retry = 6;
 
   // L7 fault injection policy applies to L7 traffic
   HTTPFaultInjection http_fault = 7;
@@ -138,8 +138,10 @@ message RouteRule {
   // Set of layer 4 match conditions based on the IP ranges
   L4MatchAttributes udp = 6;
 
-  // Set of HTTP match conditions based on the request metadata
-  HTTPMatchAttributes http = 7;
+  // Set of HTTP match conditions based on HTTP/1.1, HTTP/2, GRPC request metadata,
+  // such as "uri", "scheme", "authority".
+  // The header keys are case-insensitive.
+  map<string, StringMatch> http = 7;
 
   // Precedence is used to disambiguate the order of application of rules for
   // the same destination service. A higher number takes priority.  If not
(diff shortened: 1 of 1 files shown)


kyessenov:
Responded to feedback from @rshriram:
- not all destination policies can be applied to service versions; for now, we apply policies to all service instances
- added fault injection policy back
- renamed to service version, origin to source
- clarified multi-port service situation; match attributes are specific to each protocol in the rule
- added precedence field to rules

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,162 @@ package istio.proxy.v1alpha.config;
...
-// and downstream service invoking the API call. The choice of the specific
-// upstream cluster will be determined by the routing rule.
-message RouteRule {
-  // Uniquely identifies the upstream service (not its versions) associated
-  // with this RouteRule. This is applicable for both Http and Tcp.
-  // Note: The name used here must match with the 'name' field in the
-  // ClusterIdentifier, that is used to uniquely identify a version of the
-  // upstream service.
+// Destination declares policies that determine how to handle traffic for a
+// destination service (load balancing policies, failure recovery policies such
```
rshriram: Kuat, thinking more about this, the problem with the current t destination is that its equivalent to envoy's virtual host. All routes under that host get the same policy. This might not be okay in several cases, especially in cases of retries. The retry policy is very much dependent on the API being called. Payments/checkcredit can be retried for any error code but payments/makepayment can't be retried if destination returns a 500 or example. That's why Envoy has retry policy per route entry. (Timeout as well). In your previous version, iirc destination was equivalent to the upstream cluster with tag based qualifiers and destination name. Right? Why was this removed here?

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,162 @@ package istio.proxy.v1alpha.config;
...
+  Timeout timeout = 3;
+
+  // Retry policy
+  Retry retry = 4;
+
+  // Circuit breaker policy
+  CircuitBreaker circuit_breaker = 5;
+
+  // L7 fault injection policy applies to L7 traffic
+  HTTPFaultInjection http_fault = 6;
```
rshriram: One of here? Either a http_fault or l4_fault. Not both

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,162 @@ package istio.proxy.v1alpha.config;
...
-  ClusterIdentifier src_cluster = 2;
+  // Set of layer 4 match conditions based on the IP ranges
+  L4MatchAttributes tcp = 4;
+
+  // Set of layer 4 match conditions based on the IP ranges
+  L4MatchAttributes udp = 5;
+
+  // Set of HTTP match conditions based on the request metadata
+  HTTPMatchAttributes http = 6;
+
```
rshriram: Also one of here? As per yesterday's discussion.

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,162 @@ package istio.proxy.v1alpha.config;
...
+  // Each routing rule is associated with one or more service versions (see
+  // glossary in beginning of document). Weights associated with the service
+  // version determine the proportion of traffic it receives.
+  repeated DestinationWeight versions = 7;
+
+  // Precedence is used to disambiguate the order of application of rules
+  // for the same destination service. A higher number takes priority.
+  int32 precedence = 8;
```
rshriram: This is nice. The rules format is evolving into the existing amalgam8 constructs @frankb. We can't escape precedence. 

I would also add a version field to the rule. So that it gives the end user the ability to rollback to a stable rule if they mess up something. this would certainly mean that manager needs to provide a rollback API. (Even if you don't want to do this now, can we just add this comment so that we can track this issue. Matter of fact, a CDN company I spoke to actually wanted this capability).

I would add twl more lines here.
 E.g. for http, routes under a virtual host need to be ordered. Here precedence is priority. 

Secondly, depending on the platform, the storage for the rules format may not support atomic replace (and concurrently will be a bigger mess). The easiest way to tackle this is to ask the user to create a new rule with higher priority and delete the old rule with lower priority. Priority essentially here becomes a rule version of some sort. E.g. send all traffic to helloworld/hello to v1. Then next rule is split traffic to helloworld/hello between v1 and v2. Both rules apply to same service and same route and virtual host. 

To provide atomic update semantics to the user, the platform must support a transactional storage and patch/edit semantics as well. K8s may have etcd but unsure if mesos has.

model/proxy/alphav1/config/cfg.proto
```diff
@@ -329,40 +235,40 @@ message RetryPolicy {
...
 
-// A minimal circuit breaker configuration for the upstream cluster.
-message CircuitBreakerPolicy {
+// A minimal circuit breaker configuration.
+message CircuitBreaker {
   message SimpleCircuitBreakerPolicy {
-    // (for an unhealthy upstream cluster) number of consecutive requests that
-    // should succeed before the upstream cluster is marked healthy.
+    // (for an unhealthy upstream) number of consecutive requests that
+    // should succeed before the upstream is marked healthy.
```
rshriram: Please add a TODO to add all Envoy CB features.

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,162 @@ package istio.proxy.v1alpha.config;
...
-
-// Applications address only the upstream service using without knowledge
-// of individual upstream clusters. The actual choice of the cluster is
-// determined by the proxy, enabling the application code to decouple
-// itself from the evolution of dependent services.
+//
+// Service instances - set of pods/VMs/containers where the service is running.
+// The members share one or more common attributes. For e.g., all pods in a
+// group could share a common set of labels, or be running the same version of
+// the application binary.
```
frankbu: This seems confusing. I would say service instances share only one thing - they are all implementing some variation of the same service. I also think that before defining the concept of Service instances, we need to clearly define the term "Service" itself. Specifically, explain that a service is a name (string) of some functionality (e.g., ServiceA) representing part of an application, and point out that clients use this name to refer to the functionality being called.

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,162 @@ package istio.proxy.v1alpha.config;
...
+// enforced when the service is opening a connection/sending a request via the
+// proxy to the destination.
+//
+// N.B. We may extend destination policies to apply to individual service
+// versions and not to the entire service. Proxy mesh may support only a subset
+// of policies to be applicable to individual service version destinations.
+message Destination {
+  // Service for which the service version is defined.
   string destination = 1;
-  oneof route_rule {
```
frankbu: Wouldn't "service" be a better that "destination" for this field?

model/proxy/alphav1/config/cfg.proto
kyessenov: Brought it back. I don't like the inconsistency that some policies apply per cluster while others apply per virtual host. Seems like Envoy-specific problem.

model/proxy/alphav1/config/cfg.proto
kyessenov: We don't apply L7 fault injection to L4 service ports and vice versa?
Service must declare all its ports and the protocol on each port.

model/proxy/alphav1/config/cfg.proto
kyessenov: Same thing here. If my service declares L4 and L7 ports, I shouldn't be writing two rules for them since destinations are likely to be the same.

model/proxy/alphav1/config/cfg.proto
kyessenov: There are no transactions in k8s, there's only optimistic concurrency. Hence, we should be dealing with revision at a higher-level than this proto. Every proto has its etcd revision number implicitly. That way we can provide optimistic apply/edit like kubctl. I don't think Manager can provide strong guarantees about rollouts though, given how k8s is structured.
The priority can be used for this too, and there's nothing wrong with that. I'll add a comment.

model/proxy/alphav1/config/cfg.proto
kyessenov: "service" is a keyword in protos. In the routing context, service name is the routing destination that the rules modify.

model/proxy/alphav1/config/cfg.proto
kyessenov: Yes, deleted the old phrasing.

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,165 @@ package istio.proxy.v1alpha.config;
...
-message RouteRule {
-  // Uniquely identifies the upstream service (not its versions) associated
-  // with this RouteRule. This is applicable for both Http and Tcp.
-  // Note: The name used here must match with the 'name' field in the
-  // ClusterIdentifier, that is used to uniquely identify a version of the
-  // upstream service.
+// Destination declares policies that determine how to handle traffic for a
+// destination service (load balancing policies, failure recovery policies such
+// as timeouts, retries, circuit breakers, etc).  Policies are applicable per
+// individual service versions. It is an error to define multiple policies for
```
frankbu: From the Destination structure, it looks more like Policies are applicable to all versions of a service. How are Policies applicable per individual service version, given that the Destination structure only identifies the service and does not include any specific version info?

model/proxy/alphav1/config/cfg.proto
rshriram: Isn't that the case?
And yes, we don't do L7 fault injection on L4. or vice versa. Envoy or nginx clearly separate stream connections (UDP/TCP) from HTTP and handle HTTP separately.

model/proxy/alphav1/config/cfg.proto
frankbu: Never mind. I see this has been fixed already.

model/proxy/alphav1/config/cfg.proto
frankbu: I see. Maybe it's not a big deal, but it just seems strange to have a field name "destination" inside the "Destination" structure. Given that service is a reserved word, I understand the problem, although I would prefer something like "service_name", myself, but that's just a personal nit, so feel free to ignore.

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,165 @@ package istio.proxy.v1alpha.config;
...
-// on versions above). The proxy would choose the upstream cluster based on
-// various routing rules.
-
-// Applications address only the upstream service using without knowledge
-// of individual upstream clusters. The actual choice of the cluster is
-// determined by the proxy, enabling the application code to decouple
-// itself from the evolution of dependent services.
+//
+// Service is a unit of an application with a unique name that other services
+// can refer to the functionality being called. Service instances are
```
frankbu: s/can/use to/

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,165 @@ package istio.proxy.v1alpha.config;
...
+// pods/VMs/containers that comprise the service.
```
frankbu: s/comprise/implement/   ?

model/proxy/alphav1/config/cfg.proto
kyessenov: We don't need oneof constraint since the fault injection policies don't apply to the same traffic. If you have L4 service, you write L4 fault policy; if you have both L4 and L7 ports, you write both fault policies?

model/proxy/alphav1/config/cfg.proto
kyessenov: Done

model/proxy/alphav1/config/cfg.proto
kyessenov: Done

model/proxy/alphav1/config/cfg.proto
kyessenov: Ack

model/proxy/alphav1/config/cfg.proto
kyessenov: Done

kyessenov:
I have a question about ingress: are we planning to represent ingress resource as a routing rule or should we keep them separate?

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,165 @@ package istio.proxy.v1alpha.config;
...
-  // (either end).
-
-  // TODO: This should be done on per-cluster basis.
-  L4FaultInjection fault = 3;
-}
+  // Service version destination identifier for the destination service.
+  //
+  // N.B. The map is used instead of pstruct due to lack of serialization support
+  // in golang protobuf library (see https://github.com/golang/protobuf/pull/208)
+  map<string, string> version = 1;
```
frankbu: Version should be = 2, and the following fields bumped by 1

model/proxy/alphav1/config/cfg.proto
```diff
@@ -21,230 +21,165 @@ package istio.proxy.v1alpha.config;
...
+//
+// Source - downstream client (browser or another service) calling the
+// proxy/sidecar (typically to reach another service).
+//
+// Destination - The remote upstream service to which the proxy/sidecar is
+// talking to, on behalf of the source service. There can be one or more
+// service versions for a given service (see the discussion on versions above).
+// The proxy would choose the version based on various routing rules.
+//
+// Applications address only the destination service using without knowledge of
```
frankbu: /using/d

kyessenov:
An attempt to capture ingress routing as well:
- add optional destination to destination-weight for ingress
- add "cluster.local" destination option for the route rule
- clarify that service names are fully qualified

kyessenov:
Made up examples:
```
destination: svc.cluster.local
http:
  uri:
    prefix: /service/v1/
route:
  - destination: service.default.svc.cluster.local
     version:
     env: staging
```

```
destination: a.default.cluster.local
route:
  - version:
       env: staging
     weight: 5
  - version:
       env: prod
     weight: 95
```

kyessenov:
Changes:
- add CB policies
- replace all uints with ints
- qualify policies by HTTP prefix when they only apply to HTTP traffic
- replace HTTP match attributes with just a map

PR #108, opened 2017-02-09
Refactor cluster generation

ijsnellf:
Switched to generating services based on routing rules instead of upstream definitions.


rshriram:
Did you try with helloworld? Also update branch

ijsnellf:
I have not tried with helloworld.

PR #110, opened 2017-02-10
Refactor config generation

kyessenov:


model/service.go
```diff
@@ -40,15 +40,14 @@ type ServiceDiscovery interface {
 // Service describes an Istio service
 type Service struct {
```
rshriram: Note to self. This file needs to be re-organized.. Structs first, interfaces later, for better code readability. And needs heavy commenting, mapping out the structs to platform features in kube and mesos.

proxy/envoy/resources.go
```diff
@@ -105,15 +121,20 @@ type Runtime struct {
...
-	Headers          []Header         `json:"headers,omitempty"`
-	TimeoutMS        int              `json:"timeout_ms,omitempty"`
-	RetryPolicy      *RetryPolicy     `json:"retry_policy,omitempty"`
+
+	Headers     Headers      `json:"headers,omitempty"`
+	TimeoutMS   int          `json:"timeout_ms,omitempty"`
+	RetryPolicy *RetryPolicy `json:"retry_policy,omitempty"`
+
+	// Special field to collect clusters
+	Clusters []Cluster `json:"-"`
```
ayj: It looks like `Clusters` is only ever used in buildHTTPRoute but never used anywhere. What was the intent for this field?

kyessenov: To create an AST of some sorts by inserting data into routes themselves.
These routes get passed around a lot, and we'd have to pass a pair of route, cluster, otherwise.

ayj: Ah, I missed the reference in config.go:buildOutboundFilters because github collapsed the config.go diff.

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+}
+
+// Port represents a network port
+type Port struct {
+	// Name ascribes a human readable name for the port object. When a
+	// service has multiple ports, the name field is mandatory (FIXME
+	// really? the scenario we are worried about is a kubernetes loophole,
+	// where a looney user can define two ports for a service such that
+	// both ports have the same port number but a different name This is a
+	// corner case and more like an abuse of kubernetes.
```
kyessenov: not sure what you are suggesting. it's the convention that names must be unique and non-empty for > 1 ports.

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+	// Virtual IP.
+	Address string `json:"address,omitempty"`
+
+	// Ports is the set of network ports where the service is listening for
+	// connections.  Depending on the platform, the port can be annotated
+	// with additional information such as the type of protocol used by the
+	// port (e.g., in Kubernetes, one can use the port.protocol field to
+	// specify whether the service is listening on TCP/UDP port, and the
+	// port.name field to indicate whether the service is listening on Http
+	// port or not.
```
kyessenov: missing )

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+
+	// FIXME are we obtaining port types like HTTP2, GRPC, etc., from port
```
kyessenov: this doesn't belong here

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
 	"strings"
 )
 
-// ServiceDiscovery enumerates Istio service instances
+
+// Service describes an Istio service (e.g., catalog.mystore.com:8080)
+type Service struct {
+	// Hostname of the service, e.g. "catalog.mystore.com"
+	// TODO need a FQDN check in rules validation
```
kyessenov: what do rules have to do with the service model? the rules are optional.

model/service.go
kyessenov: the point of a name is not human readability but uniqueness. it's ok to name a port by its value.

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+	// where a looney user can define two ports for a service such that
+	// both ports have the same port number but a different name This is a
+	// corner case and more like an abuse of kubernetes.
+
+	// FIXME detect this stuff in validation and eliminate such duplicates
+	Name string `json:"name,omitempty"`
+
+	// Port number where the service can be reached Does not necessarily
+	// map to the corresponding port numbers for the instances behind the
+	// service. See endpoint definition below.
```
kyessenov: missing punctuation

model/service.go
```diff
@@ -12,6 +12,138 @@
...
+// scenario has no bearing on the application code.
+//
+// The Istio proxy sidecar intercepts and forwards all requests/responses
+// between the application and the service.  The actual choice of the
+// service version is determined dynamically by the proxy sidecar process
+// (e.g., Envoy, nginx) based on the routing rules set forth by the
+// administrator. There are layer7 (http) and layer4 routing rules (see the
+// proxy config proto definition for more details).
+
+// Routing rules allow the proxy to select a version based on criterion
```
kyessenov: route rules do not belong here.

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+	// Port number where the service can be reached Does not necessarily
+	// map to the corresponding port numbers for the instances behind the
+	// service. See endpoint definition below.
+	Port int `json:"port"`
+
+	// Protocol to be used for the port. How the protocol associated with a
+	// port is identified depends heavily on the underlying platform. For
+	// e.g., in Kubernetes, the port definition allows the user to define a
+	// name for the port where the name can indicate various higher level
+	// protocols like Http, Http2, gRPC, etc.
```
kyessenov: do not describe implementation details here. this is going to change, there is no need to hardcode it.

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+	ProtocolHTTP2 Protocol = "HTTP2"
+	ProtocolHTTP  Protocol = "HTTP"
+	ProtocolTCP   Protocol = "TCP"
+	ProtocolUDP   Protocol = "UDP"
+)
+
+// Endpoint defines a IP:port associated with an instance of the service. This is the address to which the
+// request/connection from the caller will be routed to. If a service has multiple ports, then the same
+// instance IP is listening on multiple ports (one corresponding to each service port). However,
+// we do not group such ports here. Instead we use a one endpoint per service port.
```
kyessenov: this is confusing. grouping what by what?

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+
+	// Port declaration from the service declaration This is the port for
+	// the service associated with this instance (e.g.,
+	// catalog.mystore.com)
+	ServicePort *Port `json:"port"`
+}
+
+// Tag is non empty set of key=value pair attributes (e.g., Labels in
+// kubernetes/mesos) that are assigned to pods/VMs etc. Each version of a
+// service can be uniquely identified by a set of tags
```
kyessenov: do not describe tag by its implementation.
this is bound to change and shouldn't be fixed. tag is just a tag.

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+type Tag map[string]string
+
+// TagList is a set of tags
+
+// FIXME rename to something else for clarity. Its not clear why one needs
+// TagList when Tag by itself is a set of key=value pairs
+type TagList []Tag
```
kyessenov: convenient structure to provide extra functions for a list of tags. this is a golang pattern.

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+// the set of service instances for catalog.mystore.com are
+// catalog.myservice.com ->
+//      --> Endpoint(172.16.0.1:8888), Service(catalog.myservice.com), Tag(foo=bar)
+//      --> Endpoint(172.16.0.2:8888), Service(catalog.myservice.com), Tag(foo=bar)
+//      --> Endpoint(172.16.0.3:8888), Service(catalog.myservice.com), Tag(kitty=cat)
+//      --> Endpoint(172.16.0.4:8888), Service(catalog.myservice.com), Tag(kitty=cat)
+type ServiceInstance struct {
+	Endpoint Endpoint `json:"endpoint,omitempty"`
+	Service  *Service `json:"service,omitempty"`
+	Tag      Tag      `json:"tag,omitempty"`
```
kyessenov: we need to rework this structure so not sure you want to document it now.
there is a missing concept of tag matching.
tags form a lattice by inclusion and that's not documented.

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+}
+
+// ServiceDiscovery enumerates Istio service instances These interfaces
+// must be supported by underlying platform implementations in platform/*
```
kyessenov: do not hardcode paths. this is bound to change.

model/service.go
```diff
@@ -21,40 +153,183 @@ import (
...
+	HostInstances(addrs map[string]bool) []*ServiceInstance
 }
 
-// Tag describes an Istio service tag which provides finer-grained control
-// over the set of service endpoints.
-// Tag is a non-empty set of key-value pairs.
-type Tag map[string]string
 
 // SubsetOf is true if the tag has identical values for the keys
 func (tag Tag) SubsetOf(that Tag) bool {
```
kyessenov: this should be the abstract data type for tag - it's just a lattice.

model/service.go
rshriram: This is an example isn't it?

model/service.go
rshriram: Yes, but given that Tag by itslef should have been named as Tags (plural), this TagList is even more confusing. Is it a List of Maps?

model/registry.go
```diff
@@ -15,12 +15,28 @@
...
+	"sort"
+
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
 
 	proxyconfig "istio.io/manager/model/proxy/alphav1/config"
 )
 
+// The Registry describes a set of platform agnostic APIs that must be
+// supported by the underlying platform to store and retrieve routing
```
kyessenov: It's not *just* for routing rules.

model/registry.go
```diff
@@ -15,12 +15,28 @@
...
+// rules. The code in proxy/* uses these interfaces to retrieve the routing
```
kyessenov: We'll reorganize the source tree. Leaving code references like this is not good.

rshriram: Ooo. I missed these two.. my GH page didn't show these comments until few mins ago.. Lets fix in another PR

model/registry.go
rshriram: good point. will fix in another PR..

PR #118, opened 2017-02-11
Apply load balancing policy

kyessenov:
Example how to do policy application.

### DIFF
--- a/model/registry.go
+++ b/model/registry.go
@@ -149,6 +149,19 @@ func (i *IstioRegistry) Destinations(namespace string) []*proxyconfig.Destinatio
 	return out
 }
 
+// DestinationPolicies lists all policies for a service version.
+// Policies are not inherited by tag inclusion. The version must match precisely the policy version.
+func (i *IstioRegistry) DestinationPolicies(destination string, version Tag) []*proxyconfig.Destination {
+	out := make([]*proxyconfig.Destination, 0)
+	for _, value := range i.Destinations("") {
+		if value.Destination == destination && version.Equals(value.Version) {
+			out = append(out, value)
+		}
+	}
+	// TODO: sort destination policies
+	return out
+}
+
 // RouteRulePrecedence sorts rules by precedence (high precedence first)
 type RouteRulePrecedence []*proxyconfig.RouteRule
 
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -164,7 +164,13 @@ func build(instances []*model.ServiceInstance, services []*model.Service,
 		listeners = append(listeners, listener)
 	}
 	sort.Sort(ListenersByPort(listeners))
-	return listeners, clusters.Normalize()
+
+	clusters = clusters.Normalize()
+	for _, cluster := range clusters {
+		insertDestinationPolicy(config, cluster)
+	}
+
+	return listeners, clusters
 }
 
 // buildOutboundFilters creates route configs indexed by ports for the traffic outbound
--- a/proxy/envoy/policy.go
+++ b/proxy/envoy/policy.go
@@ -17,7 +17,10 @@
 
 package envoy
 
-import "istio.io/manager/model/proxy/alphav1/config"
+import (
+	"istio.io/manager/model"
+	proxyconfig "istio.io/manager/model/proxy/alphav1/config"
+)
 
 // TODO: apply fault filter by destination as a post-processing step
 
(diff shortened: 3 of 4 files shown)


model/registry.go
```diff
@@ -107,8 +107,8 @@ type IstioRegistry struct {
 	ConfigRegistry
 }
 
-// RouteRules lists all rules in a namespace (or all rules if namespace is "")
-func (i *IstioRegistry) RouteRules(namespace string) []*proxyconfig.RouteRule {
+// ListRouteRules lists all rules in a namespace (or all rules if namespace is "")
+func (i *IstioRegistry) ListRouteRules(namespace string) []*proxyconfig.RouteRule {
 	out := make([]*proxyconfig.RouteRule, 0)
 	rs, err := i.List(RouteRule, namespace)
```
rshriram: For future PRs, can we keep changes well scoped to the intent of the PR? Or at the very least, split into multiple commits, such that a commit that does renaming functions, etc., is on its own, while a commit that changes functionality is another. [Standard practice in kernel community]. Makes it easier to bisect issues and revert the offending commit.

proxy/envoy/config.go
```diff
@@ -161,7 +161,12 @@ func build(instances []*model.ServiceInstance, services []*model.Service,
 	_ = buildFaultFilters(nil, nil)
 	_ = buildMixerCluster(mesh)
 
-	return listeners, clusters.Normalize()
+	clusters = clusters.Normalize()
+	for _, cluster := range clusters {
+		insertDestinationPolicy(config, cluster)
+	}
+
```
rshriram: Sometime in a future PR, we should decouple the cluster building from the listener building. 
buildListeners->buildFilters
buildClusters
composeConfig(listeners,clusters,tls_context)

rshriram: This would help us in future, when Envoy gets a LDS as well.

proxy/envoy/policy.go
```diff
@@ -17,14 +17,37 @@
 
 package envoy
 
-import "istio.io/manager/model/proxy/alphav1/config"
+import (
+	"istio.io/manager/model"
+	pconfig "istio.io/manager/model/proxy/alphav1/config"
```
rshriram: Not a fan of this name as well. We could call it simply istiospec or meshspec

proxy/envoy/config.go
kyessenov: We have some logic to override routes, so it's easier to keep track of used clusters by including them as part of routes. For CDS, we can include clusters referenced across all rules, and only include inbound clusters as part of the local config.

model/registry.go
kyessenov: Good point. Will do. Renaming is disruptive but we don't want to get stuck with bad names, too.

model/registry.go
```diff
@@ -149,6 +149,19 @@ func (i *IstioRegistry) Destinations(namespace string) []*proxyconfig.Destinatio
...
 }
 
+// DestinationPolicies lists all policies for a service version.
+// Policies are not inherited by tag inclusion. The version must match precisely the policy version.
+func (i *IstioRegistry) DestinationPolicies(destination string, version Tag) []*proxyconfig.Destination {
+	out := make([]*proxyconfig.Destination, 0)
+	for _, value := range i.Destinations("") {
+		if value.Destination == destination && version.Equals(value.Version) {
+			out = append(out, value)
+		}
```
rshriram: clarification: Should there be "one" policy per destination/service version ? (i.e. one set of circuit breaker config, lb policy, etc).

proxy/envoy/policy.go
```diff
@@ -35,10 +38,29 @@ func insertMixerFilter(listeners []*Listener, mixer string) {
...
+		for _, policy := range config.DestinationPolicies(cluster.hostname, cluster.tag) {
+			if policy.LoadBalancing != nil {
+				switch policy.LoadBalancing.GetName() {
+				case proxyconfig.LoadBalancing_ROUND_ROBIN:
+					cluster.LbType = LbTypeRoundRobin
+				case proxyconfig.LoadBalancing_LEAST_CONN:
+					cluster.LbType = "least_request"
+				case proxyconfig.LoadBalancing_RANDOM:
+					cluster.LbType = "random"
+				}
```
rshriram: same question as above..

rshriram:
LGTM except for the multiple policies per destination cluster.

rshriram:
Merging for now and fixing that issue later.

PR #119, opened 2017-02-11
Parallelize the integration test's client requests

ayj:
Run all of the integration test's client requests in parallel. The downside to this approach (as-is) is the logs are harder to read since requests are interleaved. 

I'm also seeing that it takes a lot of request attempts (i.e. kube execs) before the test completes, but this seems to be a problem with the serialized version as well. 

```
2017/02/10 17:44:53 Making a request http://a/a from a (attempt 0)...
2017/02/10 17:44:53 kubectl exec a-3909435095-0c4fw -n istio-integration-1797x -c app client http://a/a
2017/02/10 17:44:54 Url=http://a/a
2017/02/11 01:44:54 Get http://a/a: dial tcp 10.19.252.162:80: getsockopt: connection refused
```

### DIFF
--- a/test/integration/driver.go
+++ b/test/integration/driver.go
@@ -19,6 +19,8 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -26,6 +28,7 @@ import (
 	"os/exec"
 	"regexp"
 	"strings"
+	"sync"
 	"text/template"
 	"time"
 
@@ -35,6 +38,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/ghodss/yaml"
+	"github.com/golang/sync/errgroup"
 	flag "github.com/spf13/pflag"
 
 	"istio.io/manager/model"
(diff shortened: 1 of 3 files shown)


ayj:
This leaks namespaces if the test fails. Will commit a fix for that shortly.

test/integration/driver.go
```diff
@@ -243,53 +249,65 @@ func dumpProxyLogs(name string) {
...
+
+							// Expected no match
+							if src == "t" && dst == "t" {
+								return nil
+							}
+
+							select {
+							case <-time.After(time.Second):
+								// try again
+							case <-ctx.Done():
```
kyessenov: Is the idea here is to interrupt the sleep if some other test fails?

ayj: Yup. The errgroup will cancel the group context if any of the goroutines returns with an error and the `select` here exits the goroutine request early when that happens.

andraxylia:
The "getsockopt: connection refused" message  indicates an issue. The requests were successful with serial attempts and images from 2 days ago. If you get this without your changes, it's a baseline issue that should be fixed before this code is merged.

rshriram:
More linter errors in driver.go

test/integration/driver.go
```diff
@@ -86,58 +74,94 @@ func main() {
...
+		log.Fatal(err)
+	}
+	if err := f.Close(); err != nil {
+		log.Fatal(err)
+	}
 
-	check(w.Flush())
-	check(f.Close())
+	if err := run("kubectl apply -f " + yaml + " -n " + namespace); err != nil {
+		log.Fatal(err)
```
rshriram: Don't you have to create different echo.yaml s one per namespace? Atleast, that would be useful from a debugging perspective, when I ctrl+c during the test, to see whats up with the failures, etc.

PR #121, opened 2017-02-11
Integration tests for weighted routing & header-based routing

rshriram:


rshriram:
The routing tests can be run concurrently with the reachability tests. But @ayj 's PR needs to rebase on top of this, so that the two sets checkBasicReachability and checkRouting can be run in tandem.

test/integration/driver.go
```diff
@@ -29,30 +30,41 @@ import (
...
+	tag         string
+	namespace   string
+	verbose     bool
+	client      *kubernetes.Clientset
+	istioClient *kube.Client
 )
 
-func cleanup() {
+func teardown() {
+	if err := istioClient.DeregisterResources(); err != nil {
```
kyessenov: Let's not de-register resources. TPRs definitions are global, so this will break any concurrent test.

test/integration/driver.go
```diff
@@ -126,18 +167,127 @@ func main() {
...
+	verifyRouting(pods, "hello", "world", 100, map[string]int{
+		"v1": 75,
+		"v2": 25,
+	})
+	log.Println("Success!")
+}
+
+func addRule(ruleConfig []byte, kind string, name string, namespace string) error {
+
+	if namespace == "" {
```
kyessenov: namespace is generated when it's empty (see line 102), so this check is unnecessary.

test/integration/driver.go
```diff
@@ -126,18 +167,127 @@ func main() {
...
+	var weightedRoute bytes.Buffer
+	w = bufio.NewWriter(&weightedRoute)
+
+	check(write("test/integration/rule-weighted-route.yaml.tmpl", map[string]string{
+		"destination": "world",
+		"namespace":   namespace,
+	}, w))
+
+	check(w.Flush())
+	check(addRule(weightedRoute.Bytes(), "route-rule", "weighted-route", namespace))
```
kyessenov: can we use the constant RouteRule for "route-rule"? This would prevent accidental failure in the future. It's probably right not to use the CLI interface and use API directly.

test/integration/driver.go
```diff
@@ -126,18 +167,127 @@ func main() {
...
+	v, err := istioKind.FromJSON(string(out))
+	if err != nil {
+		return fmt.Errorf("Cannot parse proto message from JSON: %v", err)
+	}
+
+	err = istioClient.Put(model.Key{
+		Kind:      kind,
+		Name:      name,
+		Namespace: namespace,
+	}, v)
```
kyessenov: I probably need to just open up a function in cmd/ but this can be done later.

PR #130, opened 2017-02-13
fix e2e.sh

rshriram:


ayj:
@kyessenov ,  why were these setup steps moved out of driver.go into a separate bash script?

kyessenov:
bazel can't run bazel in bazel :)
So I had to move out bazel commands out. But I really want to keep this script as small as possible!

rshriram:
I would have to investigate. And its frankly a waste of time to work around like this. The simplest alternative is to have two scripts or keep this script for Jenkins while folks use something else simpler for local development. Duplication will always be a pain.

PR #133, opened 2017-02-14
Fault filter

rshriram:
Will add testing in the next pass. I have to rework my test suite after @ayj 's changes.

### DIFF
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -119,28 +119,29 @@ func build(instances []*model.ServiceInstance, services []*model.Service,
 	inbound := buildInboundFilters(instances)
 
 	// merge the two sets of route configs
-	configs := make(RouteConfigs)
-	for port, config := range inbound {
-		configs[port] = config
+	routeConfigs := make(RouteConfigs)
+	for port, routeConfig := range inbound {
+		routeConfigs[port] = routeConfig
 	}
 
 	for port, outgoing := range outbound {
-		if incoming, ok := configs[port]; ok {
+		if incoming, ok := routeConfigs[port]; ok {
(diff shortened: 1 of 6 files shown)


proxy/envoy/config.go
```diff
@@ -119,28 +116,37 @@ func build(instances []*model.ServiceInstance, services []*model.Service,
...
+		sort.Sort(HostsByName(routeConfig.VirtualHosts))
+		clusters = append(clusters, routeConfig.clusters()...)
+
+		filters := buildFaultFilters(config, routeConfig)
+
+		filters = append(filters, Filter{
+			Type:   "decoder",
+			Name:   "router",
+			Config: FilterRouterConfig{},
+		})
```
kyessenov: any particular reason why create the chain here rather than just prepending the fault filter? you're deleting mixer filter in the process...

rshriram: I am prepending the fault filter. Look at line 142. It builds all fault filters. And then 144 adds the final route filter. I am not deleting the mixer filter at all. In fact, its created in line 78, right after the build() call.

rshriram: And in policy.go at line 27, its prepended to the list of filters, before the fault filter

rshriram: And the only reason I needed to do fault filter generation in the build function was because I needed to get hold of all the clusters per http_conn_filter, so that I can generate the appropriate fault filter blocks. I cannot do that outside the build function (it becomes messy).

kyessenov: Ok, let's get this in, but I want to treat fault filter and mixer filter uniformly. That means both should prepend its category of filters and not create a chain. We need to make it less messy to do such an operation.

PR #135, opened 2017-02-14
Cleanup namespaces properly when e2e test fails

ayj:
Fix for issues/134. defer() isn't called from log.Fatal (i.e. os.Exit). Wrap the main test
function in a helper function so deferred functions can be triggered
on errors properly.

### DIFF
--- a/test/integration/driver.go
+++ b/test/integration/driver.go
@@ -87,7 +87,12 @@ func init() {
 
 func main() {
 	flag.Parse()
+	if err := endToEndTest(); err != nil {
+		log.Fatal(err)
+	}
+}
 
+func endToEndTest() error {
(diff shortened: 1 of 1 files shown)


kyessenov:
@rshriram the test error is due to your recent change:
```
2017/02/14 06:23:54 Response time is 292.393201ms with status code 0
2017/02/14 06:23:54 Expected response time is 5s +/- 2s with status code 503
```
We need to clean up stale namespaces in the cluster. I can do that once you merge this in.

rshriram:
So, I don't understand.. My recent changes is making sure that new code does not break fault injection or version routing. And looking at the above message, it seems that fault injection is not working at all. What do you suggest we do?

PR #137, opened 2017-02-14
[experiment]: use save+import to create images with valid timestamps

ayj:


kyessenov:
Alternative approach: define tag to be (whoami)-(timestamp)?

ayj:
This save+import approach didn't work. See https://github.com/istio/manager/pull/140 which is similar to @kyessenov alternative suggestion.

PR #140, opened 2017-02-14
Remove e2e.sh and change default tag to <username>_<timestamp>

ayj:


ayj:
@rshriram, I factored out the docker specific steps into [prepareDockerImage](https://github.com/istio/manager/pull/140/files#diff-da1d2d7d8abb482d3c30de50806f3fe3R145).

You can still build standalone images from the command line, e.g.

```bazel build //docker:app```

You can also use your own images by using `docker save` to create layers for the app, init, and runtime images.

```
mkdir layers 
docker save <App ImageID> -o layer/app-layer.tar
docker save <Runtime ImageD> -o layer/runtime-layer.tar
docker save <Init ImageID> -o layer/init-layer.tar
bazel run //test/integration:integration -- --layers layers/ --hub `whoami`
```
 
Did you have something else in mind?

PR #141, opened 2017-02-14
Updates WORKSPACE proxy for submodule testing

sebastienvas:


WORKSPACE
```diff
@@ -1,3 +1,7 @@
+PROXY = "eb8bfccc391e528fba316c4e081515eb18583e5c"
+
+PROXY_SHA = ""
```
kyessenov: is this supposed to be empty?

sebastienvas: Yes there is no SHA for that particular commit. "" is the default, meaning no SHA.

sebastienvas:
It seems that the last proxy broke the build.

kyessenov:
I moved proxy SHA to PROXY variable so we can close this I think.

PR #144, opened 2017-02-15
Implement service node and cluster for envoy

kyessenov:
- uses downward api to propagate pod names
- rename CLI commands: server => discovery, proxy => proxy sidecar
- decrease drain time to 30s
- print test logs in jenkins

### DIFF
--- a/proxy/envoy/watcher.go
+++ b/proxy/envoy/watcher.go
@@ -15,7 +15,6 @@
 package envoy
 
 import (
-	"net"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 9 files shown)


proxy/envoy/watcher.go
```diff
@@ -38,46 +43,47 @@ type watcher struct {
...
 // NewWatcher creates a new watcher instance with an agent
 func NewWatcher(discovery model.ServiceDiscovery, ctl model.Controller,
-	registry *model.IstioRegistry, mesh *MeshConfig) (Watcher, error) {
-	addrs, err := hostIP()
-	glog.V(2).Infof("host IPs: %v", addrs)
-	if err != nil {
-		return nil, err
+	registry *model.IstioRegistry, mesh *MeshConfig, identity *ProxyNode) (Watcher, error) {
+	addrs := make(map[string]bool)
+	if identity.IP != "" {
```
ayj: What happens if the identify IP isn't set? Do we not need a fallback?

kyessenov: it means there is no co-located service instance (e.g. no localhost routes) which is OK.
this could happen for an ingress proxy.

PR #153, opened 2017-02-16
Restrict tests to its namespace

kyessenov:
I suspect we have a starvation issue where controllers overload some resource. This should reduce the load by limiting watchers to the test namespace.
Also fixed:
- marked driver.go manual to avoid building docker images automatically
- tried with recent envoy build and discovered it's broken for manager for unknown reason

### DIFF
--- a/test/integration/http-service.yaml.tmpl
+++ b/test/integration/http-service.yaml.tmpl
@@ -43,8 +43,6 @@ spec:
           - "{{.port1}}"
           - --port
           - "{{.port2}}"
-          - --version
-          - "unversioned"
         ports:
         - containerPort: {{.port1}}
         - containerPort: {{.port2}}
@@ -57,6 +55,8 @@ spec:
           - sidecar
           - -s
           - manager:8080
+          - -n
+          - {{.namespace}}
           - --logtostderr
           - -v
           - "2"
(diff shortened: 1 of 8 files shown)


test/integration/http-service.yaml.tmpl
```diff
@@ -43,8 +43,6 @@ spec:
           - "{{.port1}}"
           - --port
           - "{{.port2}}"
-          - --version
```
ayj: Intentional? I didn't see anything in the commit summary about removing versioned service.

kyessenov: it's not used

rshriram: I think if you remove this, the server will throw an error.. Its harmless anyway. Its used by routing tests.

kyessenov:
I did some refactoring in the driver - to allow parallel and sequential requests. I need sequential to understand envoy's problem. Let me know if you have more comments.

test/integration/http-service.yaml.tmpl
kyessenov: default value for version is "". i don't see any reference to "unversioned" anywhere in the tests.

kyessenov:
I think it's hard to reason about defer with log.Fatal so let's make namespace deletion explicit. This way we know for sure namespaces don't get leaked. I always thought making log.Fatal cause a panic  was a strange decision by golang designers.

PR #158, opened 2017-02-16
Revert back to e2e sh

ayj:
* Revert Remove e2e.sh and change default tag to <username>_<timestamp> (#140)
* Update bin/e2e.sh to default tag of <username>_<timestamp>
* Update bin/e2e.sh to pass unrecognized test arguments through to driver.go


test/integration/driver.go
```diff
@@ -83,16 +82,14 @@ func init() {
 		"kube config file or empty for in-cluster")
 	flag.StringVarP(&hub, "hub", "h", "gcr.io/istio-testing",
 		"Docker hub")
-	flag.StringVarP(&tag, "tag", "t", "",
-		"Docker tag (default <username>_YYYMMDD_HHMMSS formatted date)")
+	flag.StringVarP(&tag, "tag", "t", "test",
+		"Docker tag")
```
rshriram: Don't revert this. This doesn't match with the shell script. And we have abused test tags a lot already ;)

bin/e2e.sh
```diff
@@ -0,0 +1,34 @@
...
+
+set -ex
+
+if [[ "$hub" =~ ^gcr\.io ]]; then
+    gcloud docker --authorize-only
+fi
+
+for image in app init runtime; do
+    bazel run //docker:$image
+    docker tag istio/docker:$image $hub/$image:$tag
```
rshriram: What happened to that docker save trick? I thought it helped

test/integration/driver.go
ayj: driver.go no longers builds/tags/pushes the docker images. I moved this all back to bin/e2e.sh which can invoke 'bazel run' directly and avoid the issues we were seeing with missing layers/images in Jenkins.

bin/e2e.sh
ayj: No, the save+import thing never worked. Loading the intermediate layers (e.g. bazel-bin/docker/app-layer.tar) worked locally but failed in Jenkins.

test/integration/driver.go
ayj: I removed the default value and made the user specify tag explicitly to avoid mismatch with bin/e2e.sh. PTAL

PR #163, opened 2017-02-16
Categorize size of some bazel tests

ayj:
This gets rid of the following warning:

"There were tests whose specified size is too big. Use the
--test_verbose_timeout_warnings command line option to see which ones
these are.

WARNING: //proxy/envoy:go_default_test: Test execution time (0.1s
excluding execution overhead) outside of range for MODERATE tests.
Consider setting timeout="short" or size="small".
"


platform/kube/BUILD
```diff
@@ -43,6 +43,7 @@ filegroup(
 
 go_test(
     name = "go_default_test",
+    size = "small",
```
kyessenov: is 10 seconds a small test? i think bazel kills the test if it takes too long.

ayj: According to https://bazel.build/versions/master/docs/test-encyclopedia.html size="small" implies timeout of 60 seconds.

PR #165, opened 2017-02-16
Remove sleeps in integration test

ijsnellf:
Replaces the sleep inside the integration tests with Envoy restart epoch checks that wait until the epoch has been incremented.

Also re-enables fault tests, since this change should make them more reliable.

### DIFF
--- a/test/integration/driver.go
+++ b/test/integration/driver.go
@@ -323,14 +323,21 @@ func testRouting() error {
 	if err := w.Flush(); err != nil {
 		return err
 	}
-	if err := addRule(defaultRoute.Bytes(), model.RouteRule, "default-route", namespace); err != nil {
+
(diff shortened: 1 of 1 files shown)


test/integration/driver.go
```diff
@@ -432,6 +460,44 @@ func testRouting() error {
...
+		current, err := getRestartEpoch(pod)
+		if err != nil {
+			log.Printf("Could not obtain Envoy restart epoch for %s: %v", pod, err)
+		}
+
+		if current > start {
+			return nil
+		}
+		time.Sleep(1 * time.Second)
+		n++
```
ayj: 'n' is incremented twice per loop: once here and once in the 'for' post statement one line 465

ijsnellf: Fixed.

PR #176, opened 2017-02-18
Add default tag for mixer

mandarjog:
1. Add mixerTag
2. Only delete namespaces that we create
3. Add `--skipBuild` option to e2e.sh

### DIFF
--- a/bin/e2e.sh
+++ b/bin/e2e.sh
@@ -14,6 +14,8 @@ while [[ $# -gt 0 ]]; do
     case "$1" in
         -h) hub="$2"; shift ;;
         -t) tag="$2"; shift ;;
+        --mixerTag) mixerTag="$2"; shift ;;
+        --skipBuild) SKIP_BUILD=1 ;;
         -n) namespace="$2"; shift ;;
         --use_debug_image) debug_suffix="_debug" ;;
         *) args=$args" $1" ;;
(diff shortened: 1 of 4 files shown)


test/integration/driver.go
```diff
@@ -146,6 +152,24 @@ func setup() {
 
 }
 
+func assertExistsNamespace(cl *kubernetes.Clientset, ns string) {
+	if nl, err := cl.Core().Namespaces().List(v1.ListOptions{}); err != nil {
```
kyessenov: Use Get(name) call directly.

test/integration/driver.go
```diff
@@ -88,6 +90,8 @@ func init() {
 		"Docker hub")
 	flag.StringVarP(&tag, "tag", "t", "",
 		"Docker tag")
+	flag.StringVarP(&mixerTag, "mixerTag", "", "latest",
```
kyessenov: can we put a fixed (sha-based) tag? the test is supposed to be hermetic, and updates to mixer can accidentally break our tests.

test/integration/driver.go
```diff
@@ -772,10 +798,15 @@ func generateNamespace(cl *kubernetes.Clientset) (string, error) {
 		return "", err
 	}
 	log.Printf("Created namespace %s\n", ns.Name)
+	cleanupNameSpace = true
 	return ns.Name, nil
 }
 
 func deleteNamespace(cl *kubernetes.Clientset, ns string) {
+	// do not cleanup namespace that we did not create
+	if !cleanupNameSpace {
```
kyessenov: move this to teardown() call to deleteNamespace

bin/e2e.sh
```diff
@@ -21,16 +25,18 @@ done
 [[ ! -z "$hub" ]]       && args=$args" -h $hub"
 [[ ! -z "$namespace" ]] && args=$args" -n $namespace"
 
+args=$args" --mixerTag $mixerTag"
```
kyessenov: can you process mixerTag flag like in line 11-15? bin/e2e takes same arguments as driver and passes them along.

bin/e2e.sh
```diff
@@ -21,16 +25,18 @@ done
...
-    bazel run //docker:$image$debug_suffix
-    docker tag istio/docker:$image$debug_suffix $hub/$image:$tag
-    docker push $hub/$image:$tag
+	bazel run //docker:$image$debug_suffix
+	docker tag istio/docker:$image$debug_suffix $hub/$image:$tag
+	docker push $hub/$image:$tag
 done
 
-bazel run //test/integration -- $args --norouting
+bazel $BAZEL_ARGS run //test/integration -- $args --norouting
```
kyessenov: what does this do?

test/integration/driver.go
```diff
@@ -90,7 +91,7 @@ func init() {
 		"Docker hub")
 	flag.StringVarP(&tag, "tag", "t", "",
 		"Docker tag")
-	flag.StringVarP(&mixerTag, "mixerTag", "", "latest",
+	flag.StringVarP(&mixerTag, "mixerTag", "", "",
```
kyessenov: can you put SHA here too? I know it's duplication but I use both scripts e2e and driver.go for testing. driver.go bypasses docker image creation. We didn't find a good way to get rid of the shell script.

mandarjog:
Looks like this is blocked on
https://github.com/istio/istio-testing/pull/71

Issue #150, opened 2017-02-15
Generate envoy virtual host entry for pod IP and port


rshriram:
Ingress controllers like Nginx use the pod IP instead of service IP. In such a case, Envoy drops the incoming request as the host headers no longer match. Currently the proxy config generator creates virtual hosts only for all variations of service name and the service cluster IP. It needs to take the local pod IP into account as well.
cc @kyessenov

kyessenov:
Is it important to support pod IPs directly? There's a lot more churn in the pod registry, to the point we add this now, Envoy will be constantly restarting.

kyessenov:
Or is this just for inbound connection?

rshriram:
Its inbound.. from an ingress controller. Just to be able to match the pod's local IP, when its present in the host field.
I don't think we need to restart at all.

zcahana:
I'm hitting this as well when testing the envoy-based ingress controller in #71.

I'm assigning this to myself as I plan to work on it on Sunday (yes it's a working day in here...).
If anyone else wanna give it a go before, that's fine too.

zcahana:
Oh, I didn't notice this is already assigned to @kyessenov. So, if you don't get the time to work it and want me to roll it next week, let me know.

PR #177, opened 2017-02-19
Rewrite ingress traffic host header to destination service hostname

zcahana:
A complementary fix to #150/#160, where inbound service proxy expects the host header on incoming requests to match the local pods IP / service hostname.

For the ingress controller case, the host header is either set by the client (to a virtual host name != service host name) and forwarded as-is; or set by the Envoy ingress controller to the local host.

This PR utilizes Envoy's support for host rewrite (see [docs](https://lyft.github.io/envoy/docs/configuration/http_conn_man/route_config/route.html)) to set the host header to the target service host name.

Note: tested locally/manually to verify that I can hit a backend service exposed via an ingress.

### DIFF
--- a/proxy/envoy/ingress.go
+++ b/proxy/envoy/ingress.go
@@ -169,8 +169,9 @@ func (w *ingressWatcher) generateConfig() (*Config, error) {
 // buildIngressRoute translates an ingress rule to an Envoy route
 func buildIngressRoute(rule *config.RouteRule) *Route {
 	route := &Route{
-		Path:   "",
-		Prefix: "/",
+		Path:        "",
+		Prefix:      "/",
+		HostRewrite: rule.Destination,
 	}
 
 	if rule.Match != nil {
--- a/proxy/envoy/resources.go
+++ b/proxy/envoy/resources.go
@@ -124,10 +124,13 @@ type Runtime struct {
 
 // Route definition
 type Route struct {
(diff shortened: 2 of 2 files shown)


proxy/envoy/ingress.go
```diff
@@ -169,8 +169,9 @@ func (w *ingressWatcher) generateConfig() (*Config, error) {
 // buildIngressRoute translates an ingress rule to an Envoy route
 func buildIngressRoute(rule *config.RouteRule) *Route {
 	route := &Route{
-		Path:   "",
-		Prefix: "/",
+		Path:        "",
+		Prefix:      "/",
```
rshriram: Eliminate path ? Only one is needed.

kyessenov:
Also document the default backend behavior. Ingress controller is supposed to return 404 on a fallthrough case. We can use an abort filter or deploy a sink pod as a default backend.

proxy/envoy/ingress.go
zcahana: This is merely an explicit initialization for these values, the code below takes cares of assigning these fields according to the match clause. Ultimately, the `omitempty` kicks in and makes sure only one of them gets marshaled.

zcahana:
@kyessenov Envoy by default returns 404 when no route matches. No need to deploy a default backend, unless we want some custom behavior for fallthrough.

PR #184, opened 2017-02-22
Add envoy resource definitions for TCP network filters

ayj:
This should be a non-functional change that adds TCP network filter definitions. 

Next step is to update the config generation (e.g. buildOutboundFilters) to generate the listener and cluster structs for TCP routes, merge HTTP and TCP listener/clusters based on some order of precedence (e.g. HTTP always higher priority than TCP, use RouteRule precedence), and then update the integration test to test non-HTTP services.

### DIFF
--- a/proxy/envoy/resources.go
+++ b/proxy/envoy/resources.go
@@ -227,9 +227,9 @@ type HTTPFilterConfig struct {
 type TCPRoute struct {
 	Cluster           string   `json:"cluster"`
 	DestinationIPList []string `json:"destination_ip_list,omitempty"`
-	DestinationPorts  []string `json:"destination_ports,omitempty"`
+	DestinationPorts  string   `json:"destination_ports,omitempty"`
 	SourceIPList      []string `json:"source_ip_list,omitempty"`
-	SourcePorts       []string `json:"source_ports,omitempty"`
+	SourcePorts       string   `json:"source_ports,omitempty"`
 }
 
 // TCPRouteConfigs provides routes by port
@@ -267,7 +267,7 @@ type NetworkFilter struct {
 type Listener struct {
 	Port           int              `json:"port"`
 	Filters        []*NetworkFilter `json:"filters"`
-	BindToPort     bool             `json:"bind_to_port,omitempty"`
+	BindToPort     bool             `json:"bind_to_port"`
 	UseOriginalDst bool             `json:"use_original_dst,omitempty"`
 }
 
--- a/proxy/envoy/resources.go
+++ b/proxy/envoy/resources.go
@@ -47,8 +47,10 @@ const (
 	DefaultAccessLog = "/dev/stdout"
 	LbTypeRoundRobin = "round_robin"
 
-	// HTTPConnectionManager is the name of HTTP filter
+	// HTTPConnectionManager is the name of HTTP filter.
 	HTTPConnectionManager = "http_connection_manager"
+	// TCPProxyFilter is the name of the TCP Proxy network filter.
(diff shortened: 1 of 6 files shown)


proxy/envoy/resources.go
```diff
@@ -211,40 +213,72 @@ type AccessLog struct {
...
+	Name   string      `json:"name"`
+	Config interface{} `json:"config"`
 }
 
 // Listener definition
 type Listener struct {
 	Port           int              `json:"port"`
 	Filters        []*NetworkFilter `json:"filters"`
-	BindToPort     bool             `json:"bind_to_port"`
+	BindToPort     bool             `json:"bind_to_port,omitempty"`
```
kyessenov: bind_to_port = false will be omitted. please be careful with omitempty tags here.

proxy/envoy/resources.go
```diff
@@ -211,40 +213,72 @@ type AccessLog struct {
...
+	SourceIPList      []string `json:"source_ip_list,omitempty"`
+	SourcePorts       []string `json:"source_ports,omitempty"`
+}
+
+// TCPRouteConfigs provides routes by port
+type TCPRouteConfigs map[int]*TCPRouteConfig
+
+// TCPRouteConfig (or generalize as RouteConfig or L4RouteConfig for TCP/UDP?)
+type TCPRouteConfig struct {
+	Routes []TCPRoute `json:"routes"`
```
kyessenov: add stat_prefix (required)

proxy/envoy/resources.go
```diff
@@ -211,40 +213,72 @@ type AccessLog struct {
...
+	RouteConfig       *HTTPRouteConfig `json:"route_config"`
+	Filters           []HTTPFilter     `json:"filters"`
+	AccessLog         []AccessLog      `json:"access_log"`
+}
+
+// TCPRoute definition
+type TCPRoute struct {
+	Cluster           string   `json:"cluster"`
+	DestinationIPList []string `json:"destination_ip_list,omitempty"`
+	DestinationPorts  []string `json:"destination_ports,omitempty"`
```
kyessenov: this one is a string

kyessenov:
`RouteRule` is hard to implement since tcp_proxy lacks most of the features of HTTP filters. I don't think we landed weighted clusters, for example. Let's aim to get default kubeproxy-style routing first. The addition should not interfere with HTTP routes since each entry is based on a port declaration which have assigned protocols.

rshriram:
I agree. I would suggest that the generation of tcp proxy blocks be focused on generating the destination ip based on the service IP in k8s. Adding weighted routing support to tcp is trivial but that will still not solve the issue of how to handle tcp vs http services on same port unless we distinguish services by their dns resolved IPs.

proxy/envoy/resources.go
ayj: The stat_prefix is part of the [TCPProxyFilterConfig](https://lyft.github.io/envoy/docs/configuration/network_filters/tcp_proxy_filter.html#config-network-filters-tcp-proxy-route-config). The [TCPRouteConfig](https://lyft.github.io/envoy/docs/configuration/network_filters/tcp_proxy_filter.html#config-network-filters-tcp-proxy-route-config) is just an array of TCPRoutes.

ayj:
Thanks for the feedback. I'll ignore the RouteRule for the purposes of TCP network filter for now and add a check to verify no port conflict between TCP and HTTP listeners.

proxy/envoy/resources.go
ayj: Oops, thanks - reverted.

PR #187, opened 2017-02-22
[WIP] Use CDS and RDS

ijsnellf:
Use CDS to define clusters rather than the Envoy config file.


ijsnellf:
RDS needs to be implemented as well for CDS to work. Statically defined routes need statically defined clusters.

ijsnellf:
1st pass at RDS. Doesn't pass e2e tests. There is an issue with the RDS API.

kyessenov:
Closing since #266 got merged.

PR #188, opened 2017-02-22
Use fully qualified image name (FQIN) for known good mixer image flag

ayj:
The current bin/e2e.sh script assumes the known good prebuilt mixer image is stored in the same hub as the manager images, i.e. gcr.io/istio-testing. This doesn't work if user specifies alternate docker hubs, e.g. docker.io.

### DIFF
--- a/test/integration/driver.go
+++ b/test/integration/driver.go
@@ -62,7 +62,7 @@ var (
 	inClusterConfig  bool
 	hub              string
 	tag              string
-	mixerTag         string
+	mixerImage       string
 	namespace        string
 	verbose          bool
 	norouting        bool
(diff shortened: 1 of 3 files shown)


test/integration/driver.go
```diff
@@ -116,8 +116,8 @@ func setup() {
 	if tag == "" {
 		log.Fatal("No docker tag specified with -t or --tag")
 	}
-	if mixerTag == "" {
-		log.Fatal("No mixer tag specified with --mixerTag, 'latest?'")
+	if mixerImage == "" {
+		log.Fatal("No mixer tag specified with --mixerImage, 'latest?'")
```
ijsnellf: The log message needs to be updated.

PR #190, opened 2017-02-22
Enable routing tests and test refactoring

kyessenov:


platform/kube/client.go
```diff
@@ -111,7 +112,21 @@ func CreateRESTConfig(kubeconfig string) (config *rest.Config, err error) {
 
 // NewClient creates a client to Kubernetes API using a kubeconfig file.
 // Use an empty value for `kubeconfig` to use the in-cluster config.
+// If the kubeconfig file is empty, defaults to in-cluster config as well.
 func NewClient(kubeconfig string, km model.KindMap) (*Client, error) {
+	if kubeconfig != "" {
+		info, exists := os.Stat(kubeconfig)
+		if exists != nil {
+			return nil, fmt.Errorf("Empty kubernetes configuration file: %q", kubeconfig)
```
ayj: I don't think this error message is quite correct since it implies the file exists but is empty. Maybe something like "kubernetes configuration file %q does not exist"instead?

test/integration/reachability.go
```diff
@@ -0,0 +1,172 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/golang/sync/errgroup"
+)
+
+var (
+	// accessLogs is a mapping from app name to a list of request ids that should be present in it
+	accessLogs map[string][]string
```
ayj: nit: at least internally the golang convention seems to be to define the sync.Mutex above the fields it protects.

test/integration/reachability.go
```diff
@@ -0,0 +1,172 @@
...
+	accessMu   sync.Mutex
+)
+
+func testBasicReachability() error {
+	log.Printf("Verifying basic reachability across pods/services (a, b, and t)..")
+
+	accessLogs = make(map[string][]string)
+	for app := range pods {
```
ayj: Just a thought for another CL, but it would probably be good to move away from the use of global variables in these tests, particularly now that the test is split across multiple files. We can define a testInstance struct to carry fields around and make most of these test functions methods on that struct.

test/integration/driver.go
```diff
@@ -432,33 +229,37 @@ func getRestartEpoch(pod string) (int, error) {
...
+	addConfig(config, kind, name, namespace)
+	check(waitForNewRestartEpoch(envoy, epoch))
 }
 
 func write(in string, data map[string]string, out io.Writer) error {
 	tmpl, err := template.ParseFiles(in)
+	data["namespace"] = namespace
+	data["hub"] = hub
+	data["tag"] = tag
+
```
ayj: Shouldn't the 'err' check hug the template.parseFiles(), or do we need to unconditionally update the map regardless?

platform/kube/client.go
kyessenov: good catch.

test/integration/reachability.go
kyessenov: ok, will do

test/integration/reachability.go
kyessenov: Yeah, I was thinking about that. Can implement it while at it.

test/integration/driver.go
kyessenov: I think I should conditionally merge two maps: one default with hub/tag and then input map on top of it.

ayj:
LGTM

PR #196, opened 2017-02-23
Publish manager docker images via Jenkin's post submit hook

ayj:
A similar scheme can be adopted for istio/proxy and istio/mixer once this is approved. The docker-release script could be generalized further by adding '--images' flag in which case the release script could be common to all components (via istio-testing?). Alternatively, if the release-docker script is kept in manager, developers can use to release their own set of test images without dependency on istio/istio-testing.

### DIFF
--- a/bin/release-docker.sh
+++ b/bin/release-docker.sh
@@ -35,7 +35,7 @@ set -ex
 for image in $images; do
     bazel $BAZEL_ARGS run //docker:$image
     for tag in $tags; do
-        docker tag istio/docker:init $hub/$image:$tag
+        docker tag istio/docker:$image $hub/$image:$tag
         docker push $hub/$image:$tag
     done
 done
--- a/bin/release-docker.sh
+++ b/bin/release-docker.sh
@@ -36,5 +36,8 @@ for image in $images; do
     for tag in $tags; do
         bazel $BAZEL_ARGS run //docker:$image $hub/$image:$tag
         docker push $hub/$image:$tag
+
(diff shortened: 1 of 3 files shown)


sebastienvas:
Can I just re-use this and try to make it work ?

ayj:
Sure. 

For istio/mixer use `bazel run //docker:mixer`.
For istio/proxy use `bazel run --config=release //src/envoy/mixer:proxy`.

If we wanted to add a 'latest' tag for some releases you would add the following for each image/tag tuple.

```
docker tag $hub/$image:tag $hub/$image:latest
docker push $hub/$image:latest
```

bin/release-docker.sh
```diff
@@ -33,11 +33,9 @@ fi
 set -ex
 
 for image in $images; do
+    bazel $BAZEL_ARGS run //docker:$image
     for tag in $tags; do
-        bazel $BAZEL_ARGS run //docker:$image $hub/$image:$tag
+        docker tag istio/docker:init $hub/$image:$tag
```
kyessenov: init or $image?

ayj: Oops, this should be $image. @sebastienvas may end up taking over this PR anyway as there might need to be some additional changes in istio-testing to make this work properly. Keeping this PR for now as a reference.

sebastienvas:
Thanks Jason for this. I moved the method in istio-testing for every one to use and updated manager Jenkinsfile

PR #198, opened 2017-02-24
Introducing istioctl

kyessenov:
Many changes meant to improve usability of istioctl:
- add PUT/POST to config registry
- make istioctl standalone
- move ingress kind blacklisting down to client from controller
- add -f for file input
- always log to stderr
- update test to modify route rules
- add config flow doc


doc/configuration-flow.md
```diff
@@ -0,0 +1,54 @@
+# Overview of the configuration flow in Istio Manager
+
+Istio configuration consists of configuration objects. Each object has kind
+(for example, a routing rule), name and namespace uniquely identifying the
+object.  Istio Manager provides a REST-based interface for manipulating objects
```
mandarjog: This makes it sound like I can  do
`curl https://istio_manager/...  -X PUT -d "{}"`, if that is not the case this comment is misleading. 
If it *is* the case, we should show `curl` examples.

kyessenov: It is sort of the case, since `istioctl` essentially issues curl commands to kube API server. `curl` is difficult to show since we need to translate URL, add auth headers, and not do validation.

kyessenov: REST != HTTP, but I agree REST doesn't have a clear definition anymore.

rshriram:
@kyessenov do we need to expose put/post semantics to end user? How about we just do istioctl create, istioctl update, istioctl get, istioctl delete ?

cmd/root.go
```diff
@@ -0,0 +1,92 @@
...
+	Client *kube.Client
+
+	// RootCmd is the root CLI command
+	RootCmd = &cobra.Command{
+		Short: "Istio Manager",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) (err error) {
+			glog.V(2).Infof("Root flags: %#v", RootFlags)
+			Client, err = kube.NewClient(RootFlags.Kubeconfig, model.IstioConfig)
+			if err != nil {
+				Client, err = kube.NewClient(os.Getenv("HOME")+"/.kube/config", model.IstioConfig)
```
rshriram: shouldn't this move to istioctl instead of staying here?

cmd/root.go
```diff
@@ -0,0 +1,92 @@
...
+				return multierror.Prefix(err, "Failed to register Third-Party Resources.")
+			}
+			return
+		},
+	}
+)
+
+func init() {
+	RootCmd.PersistentFlags().StringVarP(&RootFlags.Kubeconfig, "kubeconfig", "c", "",
+		"Use a Kubernetes configuration file instead of in-cluster configuration")
```
rshriram: This as well.

doc/configuration-flow.md
```diff
@@ -0,0 +1,54 @@
...
+platform-specific persistent store, and then triggers a notification event at
+the remote proxy component.
+
+## User input
+
+Istio Manager provides a command line interface called `istioctl` that exposes
+the REST-based configuration store interface.  The tool validates that its YAML
+input satisfies the schema for the registered Istio configuration kind, and
+stores the validated input in the store.
+
```
rshriram: This is good. It might need some massaging, in terms of kind->type, etc. But the watering down of the message can be done later or in another doc that is for public consumption while we keep this for internal use (design doc).

doc/configuration-flow.md
```diff
@@ -0,0 +1,54 @@
...
+Istio configuration storage is platform-specific. Below, we focus on Kubernetes,
+since Istio Manager model is close to Kubernetes model of Third-Party
+Resources. In fact, `kubectl` could be used instead of `istioctl` if validation
+and patching semantics of Third-Party Resources was supported.
+
+Each Istio configuration object is stored as a Kubernetes Third-Party resource
+of kind `istioconfig`. The name of the resource is a combination of Istio kind
+and configuration name, and namespaces are shared between Kubernetes and Istio.
+Internally, this means Kubernetes assigns a key subset in `etcd` to Istio,
+exposes an HTTP endpoint for streaming updated to the store, and provides the
```
rshriram: nit: typo - s/updated/updates/

doc/configuration-flow.md
```diff
@@ -0,0 +1,54 @@
...
+also allows watching for changes to the configuration store. For example, a
+function can subscribe to receive object creation events for objects of a
+particular kind. The controller streams the events to the subscribers, and also
+maintains a local cache view of the entire store.
+
+The flow of configuration starts at the operator, continues to the
+platform-specific persistent store, and then triggers a notification event at
+the remote proxy component.
+
+## User input
```
rshriram: Can you number the sections so that its easier to map to the flow of configuration?

platform/kube/client.go
```diff
@@ -261,21 +261,49 @@ func (cl *Client) Get(key model.Key) (proto.Message, bool) {
 	return out, true
 }
 
-// Put implements registry operation
-func (cl *Client) Put(k model.Key, v proto.Message) error {
-	out, err := modelToKube(cl.mapping, &k, v)
+// Post implements registry operation
+func (cl *Client) Post(key model.Key, v proto.Message) error {
+	if key.Kind == model.IngressRule {
+		return fmt.Errorf("unsupported operation: cannot put a config element of kind '%s'", key.Kind)
```
rshriram: nit: cannot create Ingress rule through third party config
isn't that the intent?

test/integration/rule-content-route.yaml.tmpl
```diff
@@ -1,5 +1,5 @@
 destination: "{{.destination}}.{{.namespace}}.svc.cluster.local"
-precedence: 3
+precedence: 2
```
rshriram: why this change?

test/integration/driver.go
```diff
@@ -232,28 +231,34 @@ func getRestartEpoch(pod string) (int, error) {
...
 		Name:      name,
 		Namespace: params.namespace,
-	}, v))
+	}
+	if create {
+		check(istioClient.Post(key, v))
+	} else {
+		check(istioClient.Put(key, v))
+	}
 }
```
frankbu: I wonder if instead of having to CLI commands (POST and PUT), there could just be one (ADD) with "add or create" semantics? Or all of the above, but maybe that's overkill.

doc/configuration-flow.md
frankbu: A --debug option that displays the equivalent HTTP/curl request that the command is doing might be helpful for understanding

platform/kube/kube_test.go
```diff
@@ -133,9 +133,9 @@ func TestControllerCacheFreshness(t *testing.T) {
 	k := model.Key{Kind: mock.Kind, Name: "test", Namespace: ns}
 	o := mock.Make(0)
 
-	// put followed by delete
-	glog.Infof("Calling Put(%#v)", k)
-	if err := ctl.Put(k, o); err != nil {
+	// add and remove
+	glog.Infof("Calling Postt(%#v)", k)
```
frankbu: s/Postt/Post/

cmd/istioctl/main.go
```diff
@@ -0,0 +1,225 @@
...
+
+	key    model.Key
+	schema model.ProtoSchema
+
+	postCmd = &cobra.Command{
+		Use:   "post [kind] [name]",
+		Short: "Create a configuration object",
+		RunE: func(c *cobra.Command, args []string) error {
+			if len(args) != 2 {
+				return fmt.Errorf("Provide kind and name")
```
ayj: nit: lowercase error strings here and elsewhere unless the cobra command error handling doesn't prefix these errors with anything else (see https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

frankbu:
Not exactly an issue with this particular PR, but while looking at the doc part, I noticed that `proxy-injection.md` is not up to date. It's missing the new requirement to add POD_NAME/NAMESPACE/IP env variables.

cmd/root.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Client is a kubernetes client interface set-up by the root command
+	Client *kube.Client
+
+	// RootCmd is the root CLI command
+	RootCmd = &cobra.Command{
+		Short: "Istio Manager",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) (err error) {
+			glog.V(2).Infof("Root flags: %#v", RootFlags)
+			Client, err = kube.NewClient(RootFlags.Kubeconfig, model.IstioConfig)
```
ayj: Consider not using named result parameters to avoid declaring err var inside the function (see https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters)

cmd/root.go
```diff
@@ -0,0 +1,92 @@
...
+		"Use a Kubernetes configuration file instead of in-cluster configuration")
+	RootCmd.PersistentFlags().StringVarP(&RootFlags.Namespace, "namespace", "n", "",
+		"Select a Kubernetes namespace")
+
+	// carry over glog flags with new defaults
+	flag.CommandLine.VisitAll(func(gf *flag.Flag) {
+		switch gf.Name {
+		case "logtostderr":
+			err := gf.Value.Set("true")
+			// cannot handle error here since logging is not initialized
```
ayj: Can you use fmt.Printf here to print pre-log initialization errors?

platform/kube/kube_test.go
```diff
@@ -268,9 +268,9 @@ func TestProxyConfig(t *testing.T) {
 
 	key := model.Key{Kind: model.RouteRule, Name: "test", Namespace: ns}
 
-	err := cl.Put(key, rule)
+	err := cl.Post(key, rule)
```
ayj: (nit) These two lines could be combined, e.g.

if err := cl.Post(key, rule); err != nil {
    t.Errorf(...)
}

test/integration/driver.go
rshriram: I think with just one create command, it would be slightly confusing to the end user. Especially, given that kubernetes has `kubectl apply` (which provides update semantics). In the interest of clarity, it might be useful to explicitly have create, update, get and delete. What do you think?

kyessenov: Yes, we need PATCH semantics to make "apply" work, and there the difficulty is dealing with lists in protos. That's the path kubectl went on, when a similar question was brought up (why have "create" and "replace"). It's also nice to keep each command to one atomic operation.

cmd/istioctl/main.go
```diff
@@ -0,0 +1,225 @@
...
+
+var (
+	// input file name
+	file string
+
+	key    model.Key
+	schema model.ProtoSchema
+
+	postCmd = &cobra.Command{
+		Use:   "post [kind] [name]",
```

cmd/istioctl/main.go
```diff
@@ -0,0 +1,225 @@
...
+			v, err := readInput()
+			if err != nil {
+				return err
+			}
+			return cmd.Client.Post(key, v)
+		},
+	}
+
+	putCmd = &cobra.Command{
+		Use:   "put [kind] [name]",
```

cmd/root.go
kyessenov: it's shared with cmd/manager

cmd/root.go
kyessenov: it's shared with cmd/manager/main.go

doc/configuration-flow.md
kyessenov: done

doc/configuration-flow.md
kyessenov: done

platform/kube/client.go
kyessenov: yes, that's why we catch mutations here.

test/integration/rule-content-route.yaml.tmpl
kyessenov: testing PUT instead of POST. content and default route are named the same in the test driver, but the third rule uses precedence.

platform/kube/kube_test.go
kyessenov: done

cmd/istioctl/main.go
kyessenov: done, thanks

cmd/root.go
kyessenov: I think it's ok here, since Client is a global variable.

cmd/root.go
kyessenov: done

platform/kube/kube_test.go
kyessenov: done

doc/configuration-flow.md
kyessenov: -v 10 will print kube API server curl commands, so this is already supported.

kyessenov:
Thanks for the comments! I addressed them in the last commit.

platform/kube/client.go
rshriram: No I meant, change the error message to "cannot create Ingress rule through third party resource config"

kyessenov: done

PR #200, opened 2017-02-24
Jenkinsfile updates for codecov.

sebastienvas:


Jenkinsfile
```diff
@@ -59,12 +49,30 @@ def presubmit(gitUtils, bazel) {
...
     stage('Integration Tests') {
       sh('bin/e2e.sh -t alpha' + gitUtils.GIT_SHA + ' -d')
     }
   }
 }
+
+
+def postsubmit(gitUtils, bazel, utils) {
+  buildNode(gitUtils) {
+    stage('Bazel Tests') {
```
ayj: Do we need to re-run bazel tests as part of post-submit? My understanding of post-submit hook was that it could not fail so we can't reject the merge at this point. Also, the code coverage step essentially runs an instrumented version of the go tests anyway.

Jenkinsfile
```diff
@@ -11,35 +11,25 @@ def gitUtils = new GitUtilities()
...
-  if (utils.runStage('PRESUBMIT')) {
-    def success = true
-    utils.updatePullRequest('run')
-    try {
-      presubmit(gitUtils, bazel)
-    } catch (Exception e) {
-      success = false
-      throw e
-    } finally {
-      utils.updatePullRequest('verify', success)
```
ayj: Question for my own understanding, but are these 'run' and 'verify' updates now handled by istio-testing?

kyessenov: Just a general recommendation, you should put links to groovy istio-testing sources here. It'd help to demystify the Jenkins file.

sebastienvas: https://github.com/istio/istio-testing/blob/master/vars/pullRequest.groovy

PR #209, opened 2017-02-25
Add mixer filter back; check for mixer logs

kyessenov:


test/integration/mixer.yaml.tmpl
```diff
@@ -35,9 +35,5 @@ spec:
           - --serviceConfigFile=/etc/opt/mixer/serviceconfig.yml
           - --logtostderr
           - -v
-          - "4"
-      - name: debug
```
andraxylia: Can you add back the debug image, so we can disable iptables if needed?

test/integration/reachability.go
```diff
@@ -165,10 +165,35 @@ func (r *reachability) checkAccessLogs() error {
 			return nil
 		}
 
-		if n > budget {
-			return fmt.Errorf("exceeded budget for checking access logs")
+		time.Sleep(time.Second)
```
andraxylia: There is no need for sleep here.

test/integration/reachability.go
```diff
@@ -165,10 +165,35 @@ func (r *reachability) checkAccessLogs() error {
...
+			if !found {
+				break
+			}
+		}
+
+		if found {
+			return nil
 		}
 
 		time.Sleep(time.Second)
```
andraxylia: No need for sleep here.

test/integration/mixer.yaml.tmpl
kyessenov: there's no proxy or iptables here (it's not an app)

test/integration/reachability.go
kyessenov: yes, we need to wait between log fetching. otherwise, we just check 60 times in a second.

test/integration/reachability.go
kyessenov: same reason as above

PR #224, opened 2017-02-27
Add initial plumbing for TCP routing

ayj:
TODO

- Add TCP-only service step to integration 
- Resolve port sharing between TCP and HTTP based services


rshriram:
I believe @ensible is working on adding different test clients. Just FYI.

proxy/envoy/config.go
```diff
@@ -113,8 +113,8 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service,
 // build combines the outbound and inbound routes prioritizing the latter
 func build(instances []*model.ServiceInstance, services []*model.Service,
 	config *model.IstioRegistry, mesh *MeshConfig) ([]*Listener, Clusters) {
-	outbound := buildOutboundFilters(instances, services, config, mesh)
-	inbound := buildInboundFilters(instances)
+	outbound, tcpOutbound := buildOutboundFilters(instances, services, config, mesh)
```
rshriram: Rename to httpOutbound, etc. soon we will have udp

proxy/envoy/config.go
```diff
@@ -163,6 +180,43 @@ func build(instances []*model.ServiceInstance, services []*model.Service,
...
+
+		// Add TCP filters that share a port with HTTP filters.
+		if _, ok := tcpRouteConfigs[port]; ok {
+
+			// TCP and HTTP filter on the same port seems to cause
+			// envoy to assign all traffic to TCP filter regardless of
+			// the routing config. Temporarily drop TCP filters that
+			// share a port with HTTP (e.g. istio-egress) until this is
+			// root caused.
+
```
rshriram: To solve this, you need to use the cluster IP address of the service or range of IPs of the pods (latter is hard, former is s good approximation). I suggest adding a note with regard to this. And eliminate the comment below.

proxy/envoy/config.go
```diff
@@ -194,21 +249,26 @@ func buildOutboundFilters(instances []*model.ServiceInstance, services []*model.
 				host := buildVirtualHost(service, port, suffix, routes)
 				http := httpConfigs.EnsurePort(port.Port)
 				http.VirtualHosts = append(http.VirtualHosts, host)
+			case model.ProtocolTCP:
+				cluster := buildOutboundCluster(service.Hostname, port, nil)
+				route := buildTCPRoute(cluster, service.Address, port.Port)
+				config := tcpConfigs.EnsurePort(port.Port)
+				config.Routes = append(config.Routes, route)
```
rshriram: Routes? How does this apply to TCP?

proxy/envoy/resources.go
```diff
@@ -230,6 +230,34 @@ type TCPRoute struct {
...
+	for _, route := range that.Routes {
+		if !set[route.Cluster] {
+			out.Routes = append(out.Routes, route)
+		}
+	}
+	return out
+}
+
+// Clusters aggregates clusters across routes
+func (rc *TCPRouteConfig) clusters() []*Cluster {
```
rshriram: What about de duplication ?

proxy/envoy/config.go
rshriram: Ignore please

proxy/envoy/config.go
```diff
@@ -228,10 +288,15 @@ func buildInboundFilters(instances []*model.ServiceInstance) HTTPRouteConfigs {
 
 			http := httpConfigs.EnsurePort(instance.Endpoint.Port)
 			http.VirtualHosts = append(http.VirtualHosts, host)
+		case model.ProtocolTCP:
+			cluster := buildInboundCluster(instance.Endpoint.Port, port.Protocol)
+			route := buildTCPRoute(cluster, instance.Endpoint.Address, instance.Endpoint.Port)
```
rshriram: buildInboundTcp, buildInboundHttp,..

proxy/envoy/config.go
ayj: Will do

proxy/envoy/resources.go
ayj: All clusters  (HTTP, TCP, etc.) are deduped (i.e. normalized) at config.go:225.

proxy/envoy/config.go
ayj: My understanding is that build{inbound,outbound}Cluster is common for all protocols. Were you suggesting otherwise?

rshriram: Nope. I was referring to buildTCPRoute which is adding inbound filter right ?

ayj: buildTCPRoute builds the [L4 routes ](https://lyft.github.io/envoy/docs/configuration/network_filters/tcp_proxy_filter.html#config-network-filters-tcp-proxy-route) which is used later on to contruct the TCP proxy filter.

proxy/envoy/config.go
```diff
@@ -113,8 +113,8 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service,
 // build combines the outbound and inbound routes prioritizing the latter
 func build(instances []*model.ServiceInstance, services []*model.Service,
 	config *model.IstioRegistry, mesh *MeshConfig) ([]*Listener, Clusters) {
-	outbound := buildOutboundFilters(instances, services, config, mesh)
-	inbound := buildInboundFilters(instances)
+	outbound, tcpOutbound := buildOutboundFilters(instances, services, config, mesh)
+	inbound, tcpInbound := buildInboundFilters(instances)
 
 	// merge the two sets of route configs
 	routeConfigs := make(HTTPRouteConfigs)
```
kyessenov: (nit) might also rename to httpRouteConfigs...

proxy/envoy/config.go
kyessenov: I see, this is a pretty important issue. So essentially, if I have service A on TCP 80 and service B on HTTP 80, A filter catches all requests to B. Does that happen if I also put A's cluster IP for A's TCP filter?

proxy/envoy/config.go
```diff
@@ -163,6 +180,43 @@ func build(instances []*model.ServiceInstance, services []*model.Service,
...
+			// clusters = append(clusters, tcpConfig.clusters()...)
+			delete(tcpRouteConfigs, port)
+		}
+		listeners = append(listeners, listener)
+	}
+
+	for port, tcpConfig := range tcpRouteConfigs {
+		clusters = append(clusters, tcpConfig.clusters()...)
+		listener := &Listener{
+			Port: port,
```
kyessenov: We need listener merge function I think, to merge TCP and HTTP routes.

proxy/envoy/resources.go
```diff
@@ -230,6 +230,34 @@ type TCPRoute struct {
 	DestinationPorts  string   `json:"destination_ports,omitempty"`
 	SourceIPList      []string `json:"source_ip_list,omitempty"`
 	SourcePorts       string   `json:"source_ports,omitempty"`
+
+	// special value to retain dependent cluster definition for TCP routes.
+	clusterRef *Cluster
+}
+
+// Merge operation selects a union of two route configs prioritizing the first.
+func (rc *TCPRouteConfig) merge(that *TCPRouteConfig) *TCPRouteConfig {
```
kyessenov: bug: inbound cluster has prefix inbound: and outbound clusters have prefixes outbound. we need to use consistent names (service-based?)

kyessenov: I want to make sure we don't ping-pong TCP connections. If I'm inside A pod and open a TCP connection to A, I want inbound to be selected rather than outbound.

proxy/envoy/config.go
ayj: The commented out lines at 192-200 attempted to do a basic of merge for any TCP filter that shared a port with an HTTP by appending the TCP proxy filter after the  HTTP manager filter in the network filter array. The expectation was that envoy would stop iterating through the network filters when it found a matching HTTP filter and skip over the TCP filter but that doesn't seem to be happening despite the fact that the destination IP address shouldn't match anyway. It's quite possible I messed up the TCP route config, e.g. @rshriram's comment about using cluster IP or pod IP range.

proxy/envoy/resources.go
ayj: Is that a bug re: TCP routing or in general?

kyessenov: We need to add a field to TCP route with the destination service hostname, just like a virtualhost has a field "name". That way we merge based on that field, and prefer inbound routes over outbound routes.

ayj: We have access to destination service hostname via TCPRoute.clusterRef.hostname, but I'm not sure how they would be used to merge with HTTP since envoy manages listeners by port?

re: preferring inbound over outbound, I thought that would be handled by lines 136-149 which merges inbound/outbound routes which prefers inbound over outbound. Did you have something else in mind?

kyessenov: I think you just need to use `route.Cluster.Hostname` as keys rather than clusters. That way the the first (inbound) route takes priority.

proxy/envoy/config.go
ayj: @rshriram, re: your original comment, did you mean that we need to the generate the equivalent TCPProxy filter for each HTTP service? 

@kuat - yes, A filter seems to catch all traffic for the listener port regardless of the dst/src in the TCP route config. If a TCP Proxy filter is specified for a port, then the corresponding TCP route needs to match, otherwise the filter returns StopIteration and no additional filters are checked (and the traffic isn't forwarded). I didn't find a way to configure envoy to skip over the TCP proxy filter if dst/src didn't match and fallback to the HTTP connection pool mechanism.

PR #225, opened 2017-02-27
Add support for source-based routing

kyessenov:


model/config.go
```diff
@@ -161,6 +161,32 @@ func (i *IstioRegistry) RouteRules(namespace string) []*proxyconfig.RouteRule {
 	return out
 }
 
+// RouteRulesBySource selects routing rules from source to destination (sorted by precedence)
+func (i *IstioRegistry) RouteRulesBySource(namespace string, instances []*ServiceInstance) []*proxyconfig.RouteRule {
+	out := make([]*proxyconfig.RouteRule, 0)
+	for _, rule := range i.RouteRules(namespace) {
+		if rule.Match != nil {
```
ayj: (optional) you could invert the `rule.match != nil` check here and get rid of the extra `found` variable, e.g.

```golang
if rule.Match == nil {
    continue
}
for _, instance := range instances {
    if rule.Match.Source != "" && rule.Match.Source != instance.Service.Hostname {
        continue
    }
    var tags Tags = rule.Match.SourceTags
    if !tags.SubsetOf(instance.Tags) {
        continue
    }
    out = append(out, rule)
}
```

kyessenov: yeah, thanks. also need a break statement after append.

kyessenov: actually, we'd have to duplicate code for rule.Match == nil. maybe leave it as is?

model/config.go
```diff
@@ -146,7 +146,7 @@ type IstioRegistry struct {
 	ConfigRegistry
 }
 
-// RouteRules lists all routing rules in a namespace (or all rules if namespace is "")
+// RouteRules lists all unsorted routing rules in a namespace (or all rules if namespace is "")
 func (i *IstioRegistry) RouteRules(namespace string) []*proxyconfig.RouteRule {
```
rshriram: Why can't this be sorted by precedence and destination?

proxy/envoy/config.go
```diff
@@ -184,13 +184,28 @@ func buildOutboundFilters(instances []*model.ServiceInstance, services []*model.
 	suffix := sharedInstanceHost(instances)
 	httpConfigs := make(HTTPRouteConfigs)
 
+	// get all the route rules applicable to the instances
+	rules := config.RouteRulesBySource("", instances)
+
```
rshriram: If instances is the set of all destination services in k8s, then this function basically does that rulesByDestination did.

model/config.go
ayj: Whoops, yeah. Leave it as is and ignore my comment.

kyessenov:
Let's merge this, and send coverage tests in another PR.

test/integration/rule-content-route.yaml.tmpl
```diff
@@ -1,6 +1,9 @@
-destination: "{{.destination}}.{{.namespace}}.svc.cluster.local"
+destination: {{.destination}}.{{.namespace}}.svc.cluster.local
 precedence: 2
 match:
+  source: {{.source}}.{{.namespace}}.svc.cluster.local
+  source_tags:
+    version: v1
   http:
```
rshriram: This needs to go as a separate test to ensure that other services are unable to call service "worldl

proxy/envoy/config.go
```diff
@@ -184,13 +184,28 @@ func buildOutboundFilters(instances []*model.ServiceInstance, services []*model.
 	suffix := sharedInstanceHost(instances)
 	httpConfigs := make(HTTPRouteConfigs)
 
+	// get all the route rules applicable to the instances
+	rules := config.RouteRulesBySource("", instances)
```
rshriram: Where is this code expected to run? When running in a pod, this will be a single instance and not an array. This cannot run on the host as a daemon set. And even if it does, same concept applies.

model/config.go
```diff
@@ -161,6 +161,32 @@ func (i *IstioRegistry) RouteRules(namespace string) []*proxyconfig.RouteRule {
...
 
+// RouteRulesBySource selects routing rules from source to destination (sorted by precedence)
+func (i *IstioRegistry) RouteRulesBySource(namespace string, instances []*ServiceInstance) []*proxyconfig.RouteRule {
+	out := make([]*proxyconfig.RouteRule, 0)
+	for _, rule := range i.RouteRules(namespace) {
+		if rule.Match != nil {
+			found := false
+			for _, instance := range instances {
+				if rule.Match.Source != "" && rule.Match.Source != instance.Service.Hostname {
+					continue
```
rshriram: I don't think you need arrays here. It's just the pod filtering out rules before generating the config.

rshriram:
Please wait

model/config.go
kyessenov: We certainly need to sort route rules by all its fields. I documented precendece since that's how it works now.

model/config.go
kyessenov: It's for per-node case, but I don't think it makes any difference how many instances are behind the proxy.

proxy/envoy/config.go
kyessenov: It's in a pod. If there are two services pointing to the pod, or two services with distinct ports, we'd get two elements in the array, wouldn't we?

proxy/envoy/config.go
kyessenov: No, instances is co-located instances, not external destinations.

test/integration/rule-content-route.yaml.tmpl
kyessenov: We need two versions of world?

proxy/envoy/config.go
rshriram: That's true. We really need to rename these variables in future for better readability. May be call these as localSvcInstances ?

test/integration/rule-content-route.yaml.tmpl
rshriram: No. Two versions of hello, hello v1 can only call world v1, so on. Otherwise this test as is, is not checking whether service hello-v2 can access service world-v1 or not.

kyessenov: Yeah, we can do a follow-up test PR. I will get mocks set-up as well.

proxy/envoy/config.go
kyessenov: The top-level call takes instances and services, so I was just following that naming pattern. Pretend instance is always local instances throughout.

PR #228, opened 2017-02-28
Tests to improve coverage

kyessenov:


proxy/envoy/testdata/envoy.json.golden
```diff
@@ -0,0 +1,286 @@
+{
```
ayj: How was this file generated and how would it be refreshed/updated, e.g. new envoy features, schema changes? Maybe add a smaller helper program which uses the mocks you added but just saves the JSON to a file and skips any validation step?

kyessenov: `go test istio.io/manager/proxy/envoy/...`
`cp manager/proxy/envoy/testdata/envoy.json manager/proxy/envoy/testdata/envoy.json.golden`

I can also add a flag to the test to save it as golden instead. Would that be better?

ayj: Either would work. Maybe just add a comment if you chose to reuse the existing "go test && cp" method?

kyessenov: Sure, this definitely needs documenting somewhere. Let me actually write a doc on testing with general instructions and patterns for integration, goldens, etc.

PR #239, opened 2017-03-01
Initialize Envoy at watcher creation

### LINKED ISSUE #220: Ingress controller: start proxy process regardless of existing ingress resources
Currently the ingress controller's proxy is restarted every time a new configuration is created, which is triggered only when ingress or related API resources are created/modified. If there are no ingress resources at all, the proxy process is never started.

Instead, we can bootstrap and start the proxy with an initial configuration, so that requests that land on the ingress controller proxy provide some default response (e.g., 404).

zcahana:
Modify Envoy watchers (both ingress and sidecar) to `reload()` Envoy with configuration based on initial state, instead of waiting for the first event to arrive from the controller.

Note that this is currently done synchronously (blocking), to avoid racing with controller events lurking around the corner. This can be improved once we switch to a mechanism where reloads are linearized (e.g., by a single goroutine reloader).

Fixes #220.

### DIFF
--- a/proxy/envoy/ingress.go
+++ b/proxy/envoy/ingress.go
@@ -47,6 +47,14 @@ func NewIngressWatcher(discovery model.ServiceDiscovery, ctl model.Controller,
 		mesh:      mesh,
 	}
 
+	// Initialize envoy according to the current model state,
+	// instead of waiting for the first event to arrive.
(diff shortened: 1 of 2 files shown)


rshriram:
@zcahana merge when you wish to. LGTM

PR #240, opened 2017-03-01
Ingress controller: add support for ingress class annotation

### LINKED ISSUE #215: Ingress controller: support ingress class annotation
Ingress resources can have a `kubernetes.io/ingress.class` annotation (see [documentation](https://github.com/kubernetes/ingress/blob/master/examples/PREREQUISITES.md#ingress-class)), which, if present, indicates the name of the ingress controller which should own/process the ingress resource.

This is mainly relevant for clusters in which both a user-provided and a cloud-provider ingress controllers are run together.

zcahana:
Fixes #215.

### DIFF
--- a/platform/kube/controller_test.go
+++ b/platform/kube/controller_test.go
@@ -1,0 +1,60 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kube
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


platform/kube/controller_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
+	"k8s.io/client-go/pkg/util/intstr"
+)
+
+func TestIngressIgnored(t *testing.T) {
+	cases := []struct {
+		ingressClass string
+		shouldIgnore bool
+	}{
+		{ingressClass: "nginx", shouldIgnore: true},
```
rshriram: How do we ensure that when we chain ingress controllers, other controllers respect such annotations? I guess it's out of our hand.

zcahana: We cannot ensure that others are well-behaving, but we can ensure we are.
For what it's worth, GKE and Nginx (both community and Nginx Inc) ingress controllers respect it.

rshriram:
Please document before you merge!

rshriram:
Or after. May be stick an example ingress doc in istio/istio. We can organize later.

platform/kube/controller_test.go
kyessenov: Yes, this is the right way to separate controllers from acting on the same ingress resource.

zcahana:
Opened #241 for doc'ing the ingress controller stuff.

PR #245, opened 2017-03-02
Move routing proto to API repo and some minor updates

rshriram:
Move istio routing proto to API repo (the pb.go file is still here). Another major change is the move from // to /// to enable third party tools to generate markdown docs from the proto specs. 

This PR also removed stale fields from circuit breaker config, moved HTTP request timeout and retries from destination into route rule. Future PR will rename Destination to DestinationPolicy and then move fault injection into route rule.


kyessenov:
where's PR for new proto?

rshriram:
done

PR #257, opened 2017-03-04
Add TCP routing checks to e2e integration test

ayj:
Use existing HTTP server/client to verify TCP routing rules by setting
up k8s services with non-http port names (e.g. tcp, tcp-alternative).

This also fixes bug in inbound TCP route construction by adding
ServiceAddress:ServicePort and EndpointAddress:ServicePort in addition
to EndpointAddress:Port.

### DIFF
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -262,7 +262,7 @@ func buildOutboundFilters(instances []*model.ServiceInstance, services []*model.
 				http.VirtualHosts = append(http.VirtualHosts, host)
 			case model.ProtocolTCP:
 				cluster := buildOutboundCluster(service.Hostname, port, nil)
-				route := buildTCPRoute(cluster, service.Address, port.Port)
+				route := buildTCPRoute(cluster, []string{service.Address}, port.Port)
 				config := tcpConfigs.EnsurePort(port.Port)
 				config.Routes = append(config.Routes, route)
 			default:
@@ -303,9 +303,11 @@ func buildInboundFilters(instances []*model.ServiceInstance) (HTTPRouteConfigs, 
 			http.VirtualHosts = append(http.VirtualHosts, host)
 		case model.ProtocolTCP:
 			cluster := buildInboundCluster(service.Hostname, endpoint.Port, port.Protocol)
-			route := buildTCPRoute(cluster, endpoint.Address, endpoint.Port)
 			config := tcpConfigs.EnsurePort(port.Port)
-			config.Routes = append(config.Routes, route)
+			config.Routes = append(config.Routes,
+				buildTCPRoute(cluster, []string{endpoint.Address}, endpoint.Port),
+				buildTCPRoute(cluster, []string{service.Address, endpoint.Address}, port.Port),
+			)
 		default:
 			glog.Warningf("Unsupported inbound protocol %v for port %d", port.Protocol, port)
 		}
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -303,11 +303,13 @@ func buildInboundFilters(instances []*model.ServiceInstance) (HTTPRouteConfigs, 
 			http.VirtualHosts = append(http.VirtualHosts, host)
 		case model.ProtocolTCP:
 			cluster := buildInboundCluster(service.Hostname, endpoint.Port, port.Protocol)
+
+			// Omit destination port from TCP route since inbound TCP
+			// routes are already mapped to a listener on a specific
+			// port.
+			route := buildTCPRoute(cluster, []string{endpoint.Address, service.Address}, -1)
 			config := tcpConfigs.EnsurePort(port.Port)
-			config.Routes = append(config.Routes,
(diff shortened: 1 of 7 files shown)


proxy/envoy/config.go
```diff
@@ -303,9 +303,11 @@ func buildInboundFilters(instances []*model.ServiceInstance) (HTTPRouteConfigs,
 			http.VirtualHosts = append(http.VirtualHosts, host)
 		case model.ProtocolTCP:
 			cluster := buildInboundCluster(service.Hostname, endpoint.Port, port.Protocol)
-			route := buildTCPRoute(cluster, endpoint.Address, endpoint.Port)
 			config := tcpConfigs.EnsurePort(port.Port)
-			config.Routes = append(config.Routes, route)
+			config.Routes = append(config.Routes,
+				buildTCPRoute(cluster, []string{endpoint.Address}, endpoint.Port),
+				buildTCPRoute(cluster, []string{service.Address, endpoint.Address}, port.Port),
```
kyessenov: Is there are reason for endpoint.Address here? You can address by service_ip:service_port or endpoint_ip:endpoint_port, not a combination.

test/integration/reachability.go
```diff
@@ -43,19 +43,22 @@ func (r *reachability) run() error {
...
+	if err := r.checkAccessLogs(); err != nil {
 		return err
 	}
 
-	if err = r.checkMixerLogs(); err != nil {
+	if err := r.checkMixerLogs(); err != nil {
+		return err
+	}
+
+	if err := r.verifyTCPRouting(); err != nil {
```
kyessenov: can you move this line up near `makeRequests`? We'll need to check access logs eventually.

proxy/envoy/config.go
ayj: Without that combination the route from pod B to "http://a:{90,9090}" didn't work. Pod B's kubeproxy/iptables re-routes HTTP request to Pod A's endpoint IP:Port but the listener config (below) doesn't have a route with matching port for the listener port. Maybe there is any issue with how we're binding endpoint to services in controller.go?

```
"port": 90,
        "filters": [
          {
            "type": "read",
            "name": "tcp_proxy",
            "config": {
              "stat_prefix": "tcp",
              "route_config": {
                "routes": [
                  {
                    "cluster": "inbound:9090",
                    "destination_ip_list": [
                      "10.16.0.218/32"			       // endpoint address
                    ],
                    "destination_ports": "9090" // port doesn't match listener port so this route is never used.
                  },
                  {
                    "cluster": "inbound:9090",
                    "destination_ip_list": [
                      "10.16.0.218/32"			// endpoint address
                    ],		
                    "destination_ports": "90" 		 
                  },
```

test/integration/reachability.go
ayj: Sure, but as far as I could tell envoy did not generate HTTP access logs for HTTP traffic that was only handled at L4 (i.e. tcp_proxy).

kyessenov: We can do a check with stats or use mixer logs. Not necessary for now, just something to do later.

proxy/envoy/config.go
kyessenov: Listener port for inbound TCP connections is always `targetPort`, not the service port, since port rewriting happens in the node network namespace. In HTTP protocol, the host header still keeps the old service port. Take a look here: https://github.com/istio/manager/blob/master/proxy/envoy/config.go#L302. For example, consider a service A with port 80 and target port 8080, we'll generate an entry for listener port 8080 and HTTP route with virtual host "A:80". Does that make sense? I can explain more if it's still confusing.

ayj: Yes, that makes sense and matches my own expectation. But generating service:port endpoint:port routes from the existing model did not give the correct endpoint/port tuple for the endpoint address (see corrected example configuration below). Perhaps there is an issue with how I setup the port/targetPort in the k8s templates that differs from the http ports?

```
# Port 90 listener
{
  "cluster": "inbound:9090",
  "destination_ip_list": [
    "10.16.0.230/32"
  ],
  "destination_ports": "9090"  # Never matches from external-to-internal TCP traffic (e.g. B => A:90)
},
{
  "cluster": "inbound:9090",
  "destination_ip_list": [
    "10.19.242.187/32"
  ],
  "destination_ports": "90"
},
```

kyessenov: It might be that you have both 90 and 9090 and they get mixed up (one port receives another port's traffic). I'd expect `inbound:9090` cluster to be only on listener 9090. There's no need to use destination port setting since it's known to be 9090. So a TCP service with port 90, and target port 9090, should have a TCP listener on port 9090 that sends to cluster `inbound:9090` and a listener on port 90 that sends to cluster `outbound:.....:90`

ayj: Ya, looks like traffic was probably getting bounced back and forth between 90/9090. Removing the destination port from the routing rule as you suggested fixed the issue - thanks!

kyessenov:
(merging to rebase for RDS and Shriram's work)

PR #259, opened 2017-03-06
support parsing multiple documents in istioctl

esnible:
With this change `istioctl create` now expects YAML files that include name/type/spec.  An example YAML file in that format is provided for manual testing.

There was no existing test case to modify.  A test case for the istioctl commands is needed.


cmd/istioctl/main.go
```diff
@@ -20,6 +20,10 @@ import (
 	"io/ioutil"
 	"os"
 	"strings"
+    "reflect"
+    "errors"
+    "encoding/json"
+    "bytes"
```
rshriram: Need to run go fmt

cmd/istioctl/main.go
```diff
@@ -28,8 +32,22 @@ import (
 
 	"istio.io/manager/cmd"
 	"istio.io/manager/model"
+
+	// Note that to get this package I had to do
+	//     cd $GOPATH/src/k8s.io/client-go
+	//     git checkout v2.0.0
+	// because the default branch that `go get` uses couldn't find this package.
+	"k8s.io/client-go/pkg/util/yaml"
```
rshriram: No first person comments. I think this specific dependency needs to be encoded in bazel

cmd/istioctl/main.go
```diff
@@ -28,8 +32,22 @@ import (
...
 )
 
+type MetadataAndRule struct {
+	Type string
+	Name string
+	// Spec istio.proxy.v1alpha.config.RouteRule
+	Spec interface{}
+	Rule proto.Message
```
rshriram: If we know it's a rule, we don't have to name the struct accordingly. I suggest renaming this to something generic like InputDoc

cmd/istioctl/main.go
```diff
@@ -38,20 +56,33 @@ var (
...
+			}
+			for _, v := range varr {
+				fmt.Printf("About to setup using %v with name %v\n", v.Type, v.Name)
+				setup(v.Type, v.Name)
+				fmt.Printf("About to Post rule %v\n", v.Rule)
+				err = cmd.Client.Post(key, v.Rule)
+				if err != nil {
+					return err
+				}
+				fmt.Printf("Posted rule\n")
```
rshriram: I suggest removing the prior two fmt (they are for debugging) and use this one. Specifically we could say "created %v '%v'", type, name

cmd/istioctl/main.go
```diff
@@ -223,3 +254,66 @@ func readInput() (proto.Message, error) {
...
+		v := MetadataAndRule{}
+		err = yamlDecoder.Decode(&v)
+		if err == io.EOF {
+			break;
+		}
+		if err != nil {
+			fmt.Printf("cannot parse proto message: %v", err)
+			os.Exit(5)
+		}
+		fmt.Printf("Parsed, value=%v\n", v)
```
rshriram: Please remove these debug prints

cmd/istioctl/main.go
```diff
@@ -223,3 +254,66 @@ func readInput() (proto.Message, error) {
...
+	var yamlDecoder *yaml.YAMLOrJSONDecoder = yaml.NewYAMLOrJSONDecoder(reader, 512 * 1024)
+	for {
+		v := MetadataAndRule{}
+		err = yamlDecoder.Decode(&v)
+		if err == io.EOF {
+			break;
+		}
+		if err != nil {
+			fmt.Printf("cannot parse proto message: %v", err)
+			os.Exit(5)
```
rshriram: I suggest keeping the exit code to -1. Different exit codes have different meaning in the OS

test/integration/issue213.yaml
```diff
@@ -0,0 +1,39 @@
+type: route-rule
```
rshriram: this can go in as part of unit tests.

cmd/istioctl/istioctl_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"testing"
+
+	"istio.io/manager/cmd"
+)
+
+func TestCreateInvalidFile(t *testing.T) {
+	file = "does-not-exist.yaml"
+	err := postCmd.RunE(postCmd, []string{})
```
ayj: (style): You can limit the scope of 'err' and reduce the total lines of code by combining the 'err' assignment and check into a single line, e.g.

```
if err := postCmd.RunE(...); err != nil {
    t.Fatalf(...)
}
```

cmd/istioctl/main.go
```diff
@@ -38,20 +52,30 @@ var (
...
 			if err != nil {
 				return err
 			}
-			return cmd.Client.Post(key, v)
+			if len(varr) == 0 {
+				return errors.New("Nothing to create")
+			}
+			for _, v := range varr {
+				setup(v.Type, v.Name)
+				err = cmd.Client.Post(key, v.InputDoc)
```
ayj: See previous comment about combining assignment and check of `err` into single line.

cmd/istioctl/main.go
```diff
@@ -223,3 +247,64 @@ func readInput() (proto.Message, error) {
...
+		reader, err = os.Open(file)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	var varr []MetadataAndRule
+
+	// We store route-rules as a YaML stream; there may be more than one decoder.
+	var yamlDecoder *yaml.YAMLOrJSONDecoder = yaml.NewYAMLOrJSONDecoder(reader, 512*1024)
```
ayj: You can use [https://golang.org/ref/spec#Short_variable_declarations](https://golang.org/ref/spec#Short_variable_declarations) to simplify the code and avoid redeclaring the yamlDecode type, e.g.

```
yamlDecoder := yaml.NewYAMLOrJSONDecoder(reader, 512*1024)
```

cmd/istioctl/main.go
```diff
@@ -223,3 +247,64 @@ func readInput() (proto.Message, error) {
...
+	for {
+		v := MetadataAndRule{}
+		err = yamlDecoder.Decode(&v)
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			fmt.Printf("cannot parse proto message: %v", err)
```
ayj: It would be better to return the error and let the caller handle error handling, e.g. 

```
```

Similar comment applies to the other instances of os.Exit() below.

cmd/istioctl/main.go
```diff
@@ -28,8 +32,19 @@ import (
...
+
+	"k8s.io/client-go/pkg/util/yaml"
 )
 
+// Each entry in the multi-doc YAML file used by `istioctl create -f` MUST have this format
+type inputDoc struct {
+	Type string
+	Name string
+	// Spec is usually, but not always, a istio.proxy.v1alpha.config.RouteRule
+	Spec       interface{}
```
kyessenov: Can you add `json:spec,omitempty` annotations and such here and also make parsed spec private so we don't parse that field in the input by accident?

cmd/istioctl/testdata/four-route-rules.yaml
```diff
@@ -0,0 +1,39 @@
...
+name: test-v3
+spec:
+  destination: ratings.default.svc.cluster.local
+  precedence: 1
+  route:
+  - tags:
+      version: v3
+    weight: 25
+---
+type: route-rule
```
kyessenov: add a destination policy here too? I want to make sure we handle a variety of kinds?

cmd/istioctl/main.go
rshriram: I would also remove the comment above spec. Because user can provide route-rule or destination-policy via the CLI (and other types in future).

cmd/istioctl/testdata/four-route-rules.yaml
```diff
@@ -37,3 +37,18 @@ spec:
...
+  tags:
+    version: v1
+  http_fault:
+    headers:
+      Cookie:
+        regex: "^(.*?;)?(user=jason)(;.*)?$"
+    delay:
+      fixed_delay:
+        percent: 100
+        fixed_delay_seconds: 7.0
```
rshriram: This is no longer correct unfortunately. Please try this (convert to YAML):
```json
{
  "destination": "world.default.svc.cluster.local",
  "circuit_breaker": {
    "simple_cb": {
      "max_connections": 100,
      "sleep_window_seconds": 15.5,
      "http_max_requests": 100,
      "http_max_requests_per_connection": 100,
      "http_max_pending_requests": 100,
      "http_consecutive_errors": 10,
      "http_detection_interval_seconds": 30,
      "http_max_ejection_percent": 100
    }
  }
}
```

cmd/istioctl/istioctl_test.go
```diff
@@ -0,0 +1,33 @@
...
+		t.Fatalf("Did not fail looking for file")
+	}
+}
+
+func TestInvalidType(t *testing.T) {
+	file = "testdata/invalid-type.yaml"
+	if err := postCmd.RunE(postCmd, []string{}); err == nil {
+		t.Fatalf("Did not fail when presented with invalid rule type")
+	}
+}
```
rshriram: Is the four-route-rule.yaml used in the test case? I see invalid-type and does-not-exist.yaml

rshriram:
Please update the four-rules.yaml with correct rules and then i think this is good to go.

PR #261, opened 2017-03-06
sort TCP routes into stable order to avoid unnecessary envoy restarts

ayj:


proxy/envoy/resources.go
```diff
@@ -243,6 +243,50 @@ type TCPRoute struct {
...
+}
+
+func (r TCPRouteByRoute) Less(i, j int) bool {
+	if r[i].Cluster != r[j].Cluster {
+		return r[i].Cluster < r[j].Cluster
+	}
+
+	compare := func(a, b []string) bool {
+		lenA, lenB := len(a), len(b)
+		max := lenA
```
kyessenov: max is min of lenA and lenB

PR #266, opened 2017-03-07
Add RDS and CDS

kyessenov:
Large change due to code shuffle from agent to discovery service.
Added ProxyContext to capture local proxy information context.
Removed envoy epoch checks, since envoy doesn't restart anymore for HTTP rules sometimes.
Added golden tests for the discovery service.


proxy/envoy/discovery.go
```diff
@@ -87,27 +109,99 @@ func (ds *DiscoveryService) Run() {
...
+	}
+
+	if err := response.WriteEntity(ClusterManager{Clusters: clusters}); err != nil {
+		glog.Warning(err)
+	}
+}
+
+// ListRoutes responds to RDS requests, used by HTTP routes
+func (ds *DiscoveryService) ListRoutes(request *restful.Request, response *restful.Response) {
+	if serviceCluster := request.PathParameter(ServiceCluster); serviceCluster != IstioServiceCluster {
```
rshriram: This is effectively disabling all source based routing isn't it? Why can't we run the RDS and CDS on localhost next to Envoy itself so that the proxy agents know their parent services? This is where the explicit service affinity declaration comes in.

proxy/envoy/testdata/envoy-cb-policy.json.golden
```diff
@@ -311,74 +214,22 @@
...
-            "max_requests": 100
-          }
-        },
-        "outlier_detection": {
-          "consecutive_5xx": 10,
-          "interval_ms": 30000,
-          "base_ejection_time_ms": 15500,
-          "max_ejection_percent": 100
-        }
+        "lb_type": "round_robin"
```
rshriram: Why did you remove the circuit breakers ?

proxy/envoy/testdata/envoy-timeout-rule.json.golden
```diff
@@ -9,62 +9,10 @@
...
-                    "10.2.0.0:80",
-                    "10.2.0.0"
-                  ],
-                  "routes": [
-                    {
-                      "prefix": "/",
-                      "cluster": "outbound:world.default.svc.cluster.local:http",
-                      "timeout_ms": 30000,
-                      "retry_policy": {
-                        "retry_on": "5xx,connect-failure,refused-stream",
```
rshriram: And these as well

proxy/envoy/discovery.go
kyessenov: Proxies send `service-node` which is pod IP. Discovery computes routes for that IP (by using `HostInstances` call to the service registry). Source-based routing is already being tested by the end-to-end test, although no unit tests. I can add unit tests for more coverage for calls to `ListRoutes`.

proxy/envoy/testdata/envoy-cb-policy.json.golden
kyessenov: These are now part of CDS output. I need to add a unit test with the sample output from CDS.

proxy/envoy/testdata/envoy-timeout-rule.json.golden
kyessenov: This is now output of RDS. It turns out envoy config is identical for all three cases here, and we should be using RDS/CDS output instead.

proxy/envoy/config.go
```diff
@@ -102,59 +102,35 @@ func Generate(instances []*model.ServiceInstance, services []*model.Service,
 		},
 		ClusterManager: ClusterManager{
 			Clusters: clusters,
-			SDS: SDS{
-				Cluster:        buildSDSCluster(mesh),
+			SDS: &SDS{
+				Cluster:        buildDiscoveryCluster(mesh.DiscoveryAddress, "sds"),
+				RefreshDelayMs: 1000,
```
rshriram: Please add a TODO note here to make these three fields configurable (sds, cds, rds)

proxy/envoy/agent.go
```diff
@@ -34,9 +34,11 @@ type agent struct {
...
 	// Envoy config root
 	configRoot string
-	// serviceCluster is the first component of the local proxy identityi (cluster name)
+	// serviceCluster is the first component of the local proxy identity
+	// (cluster name, e.g. "istio-proxy")
 	serviceCluster string
-	// serviceNode is the second component of the local proxy identity (node name)
+	// serviceNode is the second component of the local proxy identity
+	// (node name, e.g. pod name)
 	serviceNode string
```
rshriram: please add a note here about how source based routing is done.. 
"we use pod IP to identify the source service that the pod belongs to and filter route rules accordingly" or whatever you see fit.

proxy/envoy/discovery.go
rshriram: Can you please add this comment to the function ListRoutes ?

PR #267, opened 2017-03-07
Enable buildifer checks

sebastienvas:
Run buildifer on bazel files and fixe issues found.

### DIFF
--- a/bin/check.sh
+++ b/bin/check.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -ex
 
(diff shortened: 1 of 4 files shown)


bin/check.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -ex
 
+go get github.com/bazelbuild/buildifier/buildifier
```
kyessenov: nit: this makes sense to move into bin/init.sh instead

sebastienvas: Done

PR #270, opened 2017-03-08
Add source based routing config gen test for more coverage

kyessenov:


proxy/envoy/testdata/rds-fault.json.golden
```diff
@@ -0,0 +1,62 @@
...
+     "world.default.svc.cluster",
+     "world.default.svc.cluster.local:80",
+     "world.default.svc.cluster.local",
+     "10.2.0.0:80",
+     "10.2.0.0"
+    ],
+    "routes": [
+     {
+      "prefix": "/",
+      "cluster": "outbound:world.default.svc.cluster.local:http:version=v1",
```
rshriram: One suggestion. Can we sync this version=v0 with the header value or (use arbitrary header value scooby: doo), so that it doesn't get confusing? I realize that this is how the integration tests are done, but if we are adding test cases, this might be an opportunity to fix. WDYT?

kyessenov: Done

PR #283, opened 2017-03-09
Extend test coverge for model package

ayj:
Increase test coverage of config.go portion of model package to 100%.


WORKSPACE
```diff
@@ -268,6 +268,18 @@ new_go_repository(
     importpath = "k8s.io/client-go",
 )
 
+new_go_repository(
+    name = "com_github_golang_mock",
+    commit = "bd3c8e81be01eef76d4b503f5e687d2d1354d2d9",
+    importpath = "github.com/golang/mock",
+)
+
+new_go_repository(
```
kyessenov: We already import `go-spew` above as part of k8s dependencies. Is this one necessary and if so, could you use the same SHA.

bin/check.sh
```diff
@@ -11,7 +13,9 @@ gometalinter --concurrency=4 --enable-gc --deadline=300s --disable-all\
   --enable=goconst\
   --enable=gofmt\
   --enable=goimports\
-  --enable=golint --exclude=.pb.go\
+  --enable=golint\
+  --exclude=.pb.go\
+  --exclude=_test.go\
```
kyessenov: Do we really not want to lint all tests?

WORKSPACE
ayj: Whoops, I missed that we already imported this. Removed my version.

bin/check.sh
ayj: I was trying to avoid adding special cases for specific auto-generating mock test files, but maybe it's not a big deal (yet)? I can add a specific exclusion for mock_config_test.go and we can generalize later based on some naming scheme, e.g. "--exclude=_autogen_test.go". Thoughts?

model/BUILD
```diff
@@ -14,18 +14,41 @@ go_library(
...
+
 go_test(
     name = "go_default_test",
     size = "small",
     srcs = [
         "config_test.go",
         "service_test.go",
+        "mock_config_test.go",
     ],
+    data = [":genmocks"],
```
kyessenov: I think bazel is smart enough to run this rule without specifying it as data dependency.

ayj: So it is, thanks! I thought I tried this combination but apparently not.

bin/check.sh
kyessenov: Can we name them as "*_test.gen.go" or something like that? We also need these as exclusions to code coverage.

ayj:
*_gen.test.go" would work. Files with _test suffix already seem to be
ignored for code coverage stats.

ayj:
I had to run "go get" before 'bazel build' to install mockgen. My first attempt at this failed (i.e. bin/init.sh before build stage) but the current form seems to work. cc @sebastienvas for heads up on Jenkinsfile incase there is a better way to do this or it causes instability issues even after presubmit passes.

kyessenov:
Can you add a note to `doc/build.md` about installing mockgen before running bazel build?

PR #287, opened 2017-03-09
Increase model package test coverage to 97.5%

ayj:
All tests are currently in config_test.go which is not ideal. Now that we have good test coverage in model package I'll work on a follow-up PR to move things around into more sensible file locations.

### DIFF
--- a/model/config_test.go
+++ b/model/config_test.go
@@ -246,36 +246,44 @@ func (r *testRegistry) shutdown() {
 var (
 	defaultNamespace = "default"
 
+	endpoint1 = NetworkEndpoint{
+		Address:     "192.168.1.1",
+		Port:        10001,
+		ServicePort: &Port{Name: "http", Port: 81, Protocol: ProtocolHTTP},
+	}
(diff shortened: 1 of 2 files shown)


model/config_test.go
```diff
@@ -499,3 +507,384 @@ func TestKeyString(t *testing.T) {
...
+		},
+	}
+	for _, c := range cases {
+		if got := c.tags.Validate(); (got == nil) != c.valid {
+			t.Errorf("%s failed: got valid=%v but wanted valid=%v: %v", c.name, got == nil, c.valid, got)
+		}
+	}
+}
+
+func TestTestEquals(t *testing.T) {
```
kyessenov: Typo? There are some tests in `service_test.go`, maybe worth combining them together in one file?

ayj: Yup, thanks. Agreed that many of these tests could be shuffled around and/or combined into one file. I was going to do that in a separate non-functional PR to make it easier to review.

PR #312, opened 2017-03-13
Support HTTPS

kyessenov:
Treat HTTPS as opaque TCP traffic.

Build fixes:
* organize init scripts
* document go1.8 as a requirement
* start errors with lower case 
* fix istioctl test dependency


rshriram:
Why do we need to go down this path? I thought the auth service was going to handle inter-service https. We need to do ingress level https for sure and I think @ijsnellf is taking a crack at it, with kube secrets based option as a first start

kyessenov:
It's two separate problems:
1) app doesn't trust Istio with https, we just route it as TCP
2) app chooses to use Istio auth (e.g. ingress), we implement Istio auth design

PR #314, opened 2017-03-13
Use $KUBECONFIG in istioctl

esnible:
If $KUBECONFIG defined, use it for --kubeconfig.  Don't silently fail if --kubeconfig or $KUBECONFIG is invalid.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -48,13 +48,23 @@ var (
 		Short: "Istio Manager",
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) (err error) {
 			glog.V(2).Infof("Root flags: %#v", RootFlags)
-			Client, err = kube.NewClient(RootFlags.Kubeconfig, model.IstioConfig)
-			if err != nil {
-				Client, err = kube.NewClient(os.Getenv("HOME")+"/.kube/config", model.IstioConfig)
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -48,13 +48,23 @@ var (
...
+			if RootFlags.Kubeconfig == "" {
+				if os.Getenv("KUBECONFIG") != "" {
+					glog.V(2).Infof("Setting configuration from KUBECONFIG environment variable")
+					RootFlags.Kubeconfig = os.Getenv("KUBECONFIG")
+				}
+			}
+
 			Client, err = kube.NewClient(RootFlags.Kubeconfig, model.IstioConfig)
-			if err != nil {
+			if err != nil && RootFlags.Kubeconfig == "" {
```
kyessenov: Is the second predicate necessary? Why not fallback to .kube/config even if KUBECONFIG is wrong?

esnible: With the implementation before this change, it is possible to say `istioctl --kubeconfig ~/myconfig.yaml list route-rule`.  This is a problem if the developer really wants to keep a default ~/.kube/config and some alternatives for other clusters and makes a typo.  We don't want the behavior on a typo to be silently switching to a completely different default cluster.

PR #318, opened 2017-03-14
Run istio containers in priviledged mode

andraxylia:


kyessenov:
Why privileged mode? We just need it for testing, right?

andraxylia:
There is no guarantee envoy will no longer crash in the future, and while we are still in development mode, it is important to enable core dumps and get our hands on the core files.

ayj:
Do we only need privileged mode to set kernel.core_pattern? If so, can we achieve the result via init-container and modified prepare_proxy.sh (and adding equivalent for non-side car proxy deployments?).

kyessenov:
I don't think we should be using elevated security in our examples (demo or here). A catastrophic failure of Envoy should be dealt with health checking and whole pod termination. Maybe add a debug variable to the template so it's clear that it's purely for debugging.

andraxylia:
Will change prepare_proxy.

PR #322, opened 2017-03-14
Fix generated file to be deployable if different namespaces

andraxylia:


andraxylia:
This does not help with testing in multiple namespaces, but it helps if we need to use the generated files, since instead of generating something like -n andra_namespace, it will generate -n $(POD_NAMESPACE), which will later work in any namespace without modification. It will help avoid errors in the future if we re-use the generated files.

test/integration/manager-discovery.yaml.tmpl
```diff
@@ -27,9 +27,14 @@ spec:
       - name: manager
         image: {{.hub}}/runtime:{{.tag}}
         imagePullPolicy: Always
-        args: ["discovery", "-v", "2", "-n", "{{.namespace}}"]
+        args: ["discovery", "-v", "2", "-n", "$(POD_NAMESPACE)"]
         ports:
         - containerPort: 8080
+        env:
```
kyessenov: please remove this. discovery can be run in a separate namespace, and using downward API here is not appropriate.

andraxylia: Ok, will remove this. For clarification, -n flag is used to filter the k8s namespaces the manager watches for events as pod, svc creation/deletion etc. This should be the namespace  (or a list of namespaces) where the apps are deployed, and not the namespace where istio infra is deployed.

andraxylia:
For clarification, -n flag is used to filter the k8s namespaces the manager watches for events as pod, svc creation/deletion etc. This should be the namespace (or a list of namespaces) where the apps are deployed, and not the namespace where istio infra is deployed.

PR #330, opened 2017-03-15
Add more logging to e2e tests

kyessenov:
- cleanup rules in the test to make it idempotent
- increase verbosity for debug image

### DIFF
--- a/proxy/envoy/config_test.go
+++ b/proxy/envoy/config_test.go
@@ -19,9 +19,11 @@ import (
 	"io/ioutil"
 	"reflect"
 	"sort"
+	"strings"
 	"testing"
 
 	"github.com/golang/protobuf/proto"
+	"github.com/pmezard/go-difflib/difflib"
 
 	"istio.io/manager/model"
 	"istio.io/manager/test/mock"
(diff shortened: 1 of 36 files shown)


test/client/client.go
```diff
@@ -22,16 +22,19 @@ import (
...
 
 var (
-	count int
+	count   int
+	timeout int
 )
 
 func init() {
 	flag.IntVar(&count, "count", 1, "Number of times to make the request")
+	flag.IntVar(&timeout, "timeout", 60, "Timeout in seconds")
```
ayj: You could use a native time.Duration flag here, e.g.

```
var (
    timeout time.Duration
)

func init() {
    flag.DurationVar(&timeout, "timeout", 60 * time.Second, "Timeout")
}


client := http.Client{
    Timeout: timeout,
}
```

proxy/envoy/config_test.go
```diff
@@ -241,20 +270,8 @@ func testConfig(r *model.IstioRegistry, instance, envoyConfig, testCase string,
...
-	expected, err := ioutil.ReadFile(envoyConfig + ".golden")
-	if err != nil {
-		t.Fatalf(err.Error())
-	}
-
-	// TODO: use difflib to obtain detailed diff
-	if string(expected) != string(data) {
-		t.Errorf("Envoy config %q differs from master copy for %q", envoyConfig, testCase)
-	}
+	compareJSON(envoyConfig, t)
```
ayj: It might be better to return the error (and diff) and handle logging/t.Errorf in the caller, e.g.

```golang
if diff, eq := compareJSON(envoyConfig); !eq {
    t.Errorf("Failed validating golden artifact: %v", diff)
}
```

test/integration/driver.go
```diff
@@ -77,20 +73,16 @@ var (
...
-	flag.StringVar(&params.mixerImage, "mixerImage", "gcr.io/istio-testing/mixer:"+mixerTag,
+	flag.StringVar(&params.hub, "hub", "gcr.io/istio-testing", "Docker hub")
+	flag.StringVar(&params.tag, "tag", "", "Docker tag")
+	flag.StringVar(&params.mixerImage, "mixer", "gcr.io/istio-testing/mixer:"+mixerTag,
 		"Mixer Docker image")
-	flag.StringVarP(&params.namespace, "namespace", "n", "",
+	flag.StringVar(&params.namespace, "n", "",
 		"Namespace to use for testing (empty to create/delete temporary one)")
-	flag.StringVarP(&kubeconfig, "config", "c", "platform/kube/config",
+	flag.StringVar(&params.kubeconfig, "c", "platform/kube/config",
```
rshriram: WE should be following same convention as kubectl.. -kubeconfig, KUBECONFIG and .kube/config

kyessenov: Yeah, we need to unify all the code for discovering kube config file. This is really meant to be used as a test though (look at the hardcoded platform/kube/config).

PR #337, opened 2017-03-16
Add postsubmit test run with debug mode

kyessenov:


test/client/client.go
```diff
@@ -17,78 +17,92 @@
...
 		}
 
-		fmt.Printf("StatusCode=%d\n", resp.StatusCode)
-		_, err = io.Copy(os.Stdout, resp.Body)
+		log.Printf("[%d] StatusCode=%d\n", i, resp.StatusCode)
+
+		data, err := ioutil.ReadAll(resp.Body)
 		if err != nil {
-			log.Println(err.Error())
+			return err
```
andraxylia: Add resp.Body.Close(), or maybe defer the close somewhere.

kyessenov: Thanks, I removed defer because a linter complained, but then I missed this case. PTAL.

PR #347, opened 2017-03-17
Use external cluster for postsubmit as well

sebastienvas:


sebastienvas:
@kyessenov Sorry about this. You need to do it for every new node.

sebastienvas:
rebuild manager/presubmit

PR #349, opened 2017-03-17
Limit postsubmit action.

sebastienvas:
Currently for postsubmit we also run the presubmit part. Modifying this
to limit postsubmit to:
- Code Coverage (To update the reference)
- Integrations tests
- Docker Push


sebastienvas:
Done

PR #354, opened 2017-03-17
Add service attributes to Mixer

kyessenov:
Supply source and target service names to Mixer.

### DIFF
--- a/proxy/envoy/policy.go
+++ b/proxy/envoy/policy.go
@@ -18,6 +18,7 @@
 package envoy
 
 import (
+	"sort"
 	"strings"
 
 	"istio.io/manager/model"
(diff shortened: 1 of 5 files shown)


proxy/envoy/policy.go
```diff
@@ -18,39 +18,58 @@
...
-		for ip := range context.Addrs {
-			ips = append(ips, ip)
-		}
-		id := strings.Join(ips, ",")
+	if context.MeshConfig.MixerAddress == "" {
+		return
+	}
+
+	// join IPs with a comma
+	ips := make([]string, 0)
```
rshriram: Why would we have multiple ?

kyessenov: The API just takes a list of IP addresses. Never happens in per-pod case.

kyessenov: Same thing for service instances except there instances are per-port (so you'd get several structs with same service)

Issue #251, opened 2017-03-03
Make `MeshConfig` a configuration resource


kyessenov:
The flags that are fed into Manager runtime docker images should become `ProxyMeshConfig` as defined by the configuration proto. These flags are:
- address of discovery services
- address of the mixer service
- ports for iptables rewriting

A workflow would be: discovery service registers itself by updating the proto, the agents register the update and start using the discovery service.

PR #361, opened 2017-03-20
Add init script parameters

kyessenov:
- parametrize the script in preparation of using `configmap`s to store global configs
- change proxy port from 5001 to 15001 and admin port from 5000 to 15000 to prevent port collisions


andraxylia:
What issue/feature is this related to? Could you add more in description?

kyessenov:
#251

PR #363, opened 2017-03-21
add istioutil for client-side proxy injection into k8s resource files

ayj:
Add a utility for client-side proxy injection into kubernetes resource files. Options include ability to override the default image, manager/mixer ports, and logging verbosity. Files with multiple YAML documents are supported and non pod-based resources are left unmodified. 

This is intended to be short lived until a proper admission controller is written at which time this tool should be deleted.

```
$ istioutil inject --help
Inject istio runtime into existing kubernete resources

Usage:
   inject [flags]

Flags:
      --discoveryPort int     Manager discovery port (default 8080)
  -f, --filename string       Unmodified input kubernetes resource filename
      --initImage string      Istio init image (default "gcr.io/istio-testing/init:latest")
      --mixerPort int         Mixer port (default 9091)
  -o, --output string         Modified output kubernetes resource filename
      --runtimeImage string   Istio runtime image (default "gcr.io/istio-testing/runtime:latest")
      --sidecarProxyUID int   Sidecar proxy UID (default 1337)
      --verbosity int         Runtime verbosity (default 2)
```


ayj:
I'll add a link to doc/proxy-injection.md from istio/istio/doc in a follow-up PR. Alternatively, we could move this utility to istio/istio and add additional subcommands to generate mixer and manager resource files from templates with overrides for image name until the base definitions settle down.

ayj:
re: tests, I originally considered diffing `got` vs. `want` YAML files but didn't think the ordering of content would be consistent re: golang maps. However, verifying the unmarshalled version should work (i.e. PodTemplateSpec).

Moving the majority of the code to a helper package should be fine. Perhaps platform/kube/inject since its k8s specific and presumably our AC code would eventually live under platform/kube as well?

kyessenov:
In the longer run, injection should be inside `platform/kube` but `cmd` is coupled with k8s so you can choose either place.  Not sure how to deal with comparing to test templates...

ayj:
Updated the PR based on the comments here and discussion in slack. I made the following changes:

- Moved `inject` subcommand into istioctl as `kube-inject` as it's one less binary to worry about shipping around. We can deprecate the command once we have a proper admission controller.
- Added tests for common resource types.
- Added version info into resource annotation so we can track which version of istioctl was used to inject the proxy.
- Updated documentation. I'm working on updating the istio/istio documentation and examples in parallel and will likely leave the duplicate version in manager for completeness.

doc/proxy-injection.md
```diff
@@ -1,94 +1,66 @@
...
+$ istioutil inject --help
+Inject istio runtime into existing kubernete resources
+
+Usage:
+   inject [flags]
+
+Flags:
+      --discoveryPort int     Manager discovery port (default 8080)
+  -f, --filename string       Unmodified input kubernetes resource filename
+      --initImage string      Istio init image (default "docker.io/istio/init_debug:latest")
```
rshriram: Do we want debug image?

doc/proxy-injection.md
```diff
@@ -1,94 +1,66 @@
...
+      --mixerPort int         Mixer port (default 9091)
+  -o, --output string         Modified output kubernetes resource filename
+      --runtimeImage string   Istio runtime image (default "docker.io/istio/runtime_debug:latest")
```
rshriram: Note for future. This needs to be reworded. Runtime is super confusing.

doc/proxy-injection.md
ayj: I added debug images while the system was fragile. Based on conversations in slack developers always seemed to end up using those to debug problems. We can switch to non-debug versions if people feel strongly about it, but I figured for the next few weeks we use debug versions.

doc/proxy-injection.md
ayj: Agreed. We can make the change here when the rest of the manager naming is cleaned up and jenkin images are pushed with new names.

ayj:
> Please see comments for docs. Also I suggest you test this with bookinfo demo once to save us from any pain. Echo is already validated through manager tests.

Will do. I'll address that as part of https://github.com/istio/istio/pull/75

doc/proxy-injection.md
kyessenov: The biggest concern is the size. Debug envoy build binary is over 150MB in size.

doc/proxy-injection.md
```diff
@@ -1,94 +1,66 @@
...
-  image: docker.io/istio/runtime:latest
-  imagePullPolicy: Always
-  securityContext:
-    runAsUser: 1337
-  args:
-    - proxy
-    - -s
-    - manager:8080
-```
+    istioctl kube-inject -f depoyment.yaml | kubectl appy -f -
```
kyessenov: typo: apply

doc/proxy-injection.md
```diff
@@ -1,94 +1,66 @@
 # Proxy sidecar injection
 
+## Automatic injection
+
+Istio's goal is transparent proxy injection into end-user deployments
+with minimal effort from the end-user. Ideally, a kubernetes admission
+controller would rewrite specs to include the necessary init and proxy
+containers before they are committed, but his currently requires
```
kyessenov: "this"

doc/proxy-injection.md
```diff
@@ -1,94 +1,66 @@
...
-        imagePullPolicy: Always
-        securityContext:
-          runAsUser: 1337
-        args:
-          - proxy
-          - -s
-          - manager:8080
----
-```
+    kubectl get deployment -o yaml | istioctl kube-inject -f - | kubectl apply -f -
```
kyessenov: Have you tried `kubectl edit`?

doc/proxy-injection.md
```diff
@@ -1,94 +1,66 @@
...
+added as necessary.
+
+Unsupported resources are left unmodified so, for example, it is safe
+to run `istioctl kube-inject` over a single file that contains multiple
+Service, ConfigMap, and Deployment definitions for a complex
+application.
 
-Istio's goal is transparent proxy injection into end-user deployments with minimal effort from the end-user. Ideally, a kubernetes admission controller would rewrite specs to include the necessary init and proxy containers before they are committed, but his currently requires upstreaming changes to kubernetes which we would like to avoid for now. Instead, it would be better if a dynamic plug-in mechanism existed whereby admisson controllers could be maintained out-of-tree. There is no platform support for this yet, but a proposal has been created to add such a feature (see [Proposal: Extensible Admission Control](https://github.com/kubernetes/community/pull/132/)).
+The Istio project is continually evolving so the low-level proxy
+configuration may change unannounced. When it doubt re-run `istioctl kube-inject`
```
kyessenov: "in doubt'

.gitignore
```diff
@@ -5,6 +5,7 @@ vendor
 .idea*
 *.iml
 echo.yaml
```
kyessenov: remove "echo.yaml" since it's already ignored by a wildcard.

platform/kube/inject/testdata/frontend.yaml.injected
```diff
@@ -0,0 +1,77 @@
...
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: POD_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
+        image: docker.io/istio/runtime:latest
+        imagePullPolicy: Always
+        name: proxy
+        resources: {}
```
kyessenov: Is this some artifact of YAML serialization?

ayj: Yes. The same appears in the output of `kubectl get -f <file> -o yaml>` with an unmodified resource file.

doc/proxy-injection.md
ayj: How do you mean? `kubectl edit` is interactive so we'd need to modify KUBE_EDITOR to use `istioctl kube-inject`. On the other hand, `kubectl edit` in interactive mode does show the resource with injected proxy.

```
$ KUBE_EDITOR="grep 'alpha.istio.io/sidecar: injected'" kubectl edit -f in.yaml
        alpha.istio.io/sidecar: injected
```

PR #365, opened 2017-03-21
Allow init container to set kernel core pattern

### LINKED ISSUE #364: core pattern has no effect in init script
Seems like the core pattern setter has no effect in the init script:
`kubect logs ... -c iptables`
`sysctl: setting key "kernel.core_pattern": Read-only file system`

andraxylia:
Fixes #364

### DIFF
--- a/test/integration/app-proxy-manager-agent.yaml.tmpl
+++ b/test/integration/app-proxy-manager-agent.yaml.tmpl
@@ -40,7 +40,7 @@ spec:
             "image": "{{.hub}}/init:{{.tag}}",
             "args": ["-p", "15001", "-u", "1337"],
             "imagePullPolicy": "Always",
-            "securityContext": { "capabilities" : { "add" : ["NET_ADMIN"] } }
+            "securityContext": { "capabilities" : { "add" : ["NET_ADMIN"] }, "privileged": true }
           }]
     spec:
       containers:


test/integration/app-proxy-manager-agent.yaml.tmpl
```diff
@@ -40,7 +40,7 @@ spec:
             "image": "{{.hub}}/init:{{.tag}}",
             "args": ["-p", "15001", "-u", "1337"],
             "imagePullPolicy": "Always",
-            "securityContext": { "capabilities" : { "add" : ["NET_ADMIN"] } }
+            "securityContext": { "capabilities" : { "add" : ["NET_ADMIN"] }, "privileged": true }
```
ayj: Is there a more specific set of capabilities that would work instead of full privileges? Maybe CAP_SYS_ADMIN (untested)?

kyessenov: Let's remote sysctl command from init container and put it into debug container. Core dumps are only useful if you are willing to debug them.

ayj: It might not be the case that the user who is able to reproduce the bug is capable of debugging the problem. There might also be bugs that are hard to reproduce again with debug image to capture the core dump.

kyessenov: We should also do `ulimit -c unlimited` since that's not guaranteed.

andraxylia:
Fixed in kube inject tool by adding a config to enable core dump #411

Issue #336, opened 2017-03-15
Proxy agent should retry restarting the proxy


kyessenov:
Due to arbitrary duration of the proxy init phase, hot restart protocol can fail to start the proxy. The agent should attempt to restart the proxy with an exponential backoff.

PR #366, opened 2017-03-21
Improved and resilient proxy agent

kyessenov:
- Generalize the proxy agent and add a retry budget with an exponential back-off

- Provide 100% test coverage for the agent

- Refactor how watcher uses an agent

- Fix minor inconveniences I discovered:
  * do not use `NodePort` in the test since it must be unique per cluster
  * use `<-chan` for closing channel
  * populate all flags with default mesh values


kyessenov:
#336

proxy/agent.go
```diff
@@ -0,0 +1,198 @@
...
+
+// Agent manages the restarts and the life cycle of a proxy binary.  Agent
+// keeps track of all running proxy processes and their configurations.  Hot
+// restarts are performed by launching a new Envoy process with a strictly
+// incremented restart epoch.  This matches Envoy semantics for restart epochs:
+// To successfully launch a new Envoy process that will replace the running
+// Envoy processes, the restart epoch of the new process must be exactly 1
+// greater than the highest restart epoch of the currently running Envoy
+// processes. The initial epoch is 0.
+//
```
rshriram: Just a small nit. if this file is agnostic of proxy implementations, we should not mention Envoy here. But if its specific to envoy, then we could push the interface down into proxy/envoy

proxy/agent.go
```diff
@@ -0,0 +1,198 @@
...
+// Envoy hot restart documentation:
+// https://lyft.github.io/envoy/docs/intro/arch_overview/hot_restart.html
+type Agent interface {
+	// Apply sets the desired configuration for the proxy.  Agent compares the
+	// current active configuration to the desired state and initiates a restart
+	// if necessary. If the restart fails, the agent attempts to retry with an
+	// exponential back-off.
+	Apply(config interface{})
```
rshriram: suggest renaming to UpdateConfig ?

kyessenov:
Would appreciate if someone with good understanding of goroutines looked at `agent.go`.

proxy/agent.go
```diff
@@ -0,0 +1,198 @@
...
+
+// NewAgent creates a new proxy instance agent for a "run" function is a
+// blocking call to start the proxy, "cleanup" function is called after "run"
+// finishes.  Both functions take the epoch as an argument. Retry is budgeted
+// and uses an exponential back-off from the initial delay.
+func NewAgent(run func(interface{}, int) error, cleanup func(int),
+	initialRetryBudget int, initialRetryDelay time.Duration) Agent {
+	return &agent{
+		run:                run,
+		cleanup:            cleanup,
```
rshriram: Why can't we have cleanup as part of the agent interface?

proxy/agent.go
rshriram: Alternatively, we could rename the epoch to version (config file version [not schema version]). This would then be generic to different proxies

proxy/agent.go
```diff
@@ -0,0 +1,198 @@
...
+	current interface{}
+
+	// channel for posting desired configurations
+	config chan interface{}
+
+	// channel for proxy exit notifications
+	status chan exitStatus
+
+	// retry budget count
+	retryBudget int
```
rshriram: is this the current number of retries for the specific epoch? if so, I would rename this to retries and the initialRetryBudget to maxRetries. and retryDelay to initialRetryInterval. Although, i don't think we need that field to be configurable. Too many configuration parameters. We could just set a standard 5s initial retry delay.

proxy/agent.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// total budget available
+	initialRetryBudget int
+	initialRetryDelay  time.Duration
```
rshriram: time between retries or total time across all retries? Given that this is about process management, i think you are referring to time between retries.

proxy/agent.go
```diff
@@ -0,0 +1,198 @@
...
+	// Throttle processing up to smoothed 1 qps with bursts up to 10 qps
+	rateLimiter := flowcontrol.NewTokenBucketRateLimiter(float32(1), 10)
+
+	// Set default delay to a long duration - reconciliation is a no-op in the regular case.
+	// For permanent errors, reconciliation is still attempted once after the default delay.
+	defaultDelay := 1 * time.Hour
+	delay := defaultDelay
+
+	for {
+		rateLimiter.Accept()
```
rshriram: are we getting incremental updates or are we getting the entire config as such? If incremental updates, then this code sounds fine. But if each update is the entire config, then don't we need to just skip the earlier configs and retain just the most recent config?

proxy/agent.go
```diff
@@ -0,0 +1,198 @@
...
+
+			// schedule a retry for a transient error
+			if status.err != nil && !reflect.DeepEqual(a.desired, a.current) {
+				if a.retryBudget > 0 {
+					delay = a.initialRetryDelay * (2 << uint(a.initialRetryBudget-a.retryBudget))
+					a.retryBudget = a.retryBudget - 1
+					glog.V(2).Infof("Schedule retry after %v (budget %d)", delay, a.retryBudget)
+				} else {
+					glog.Warningf("Permanent error: budget exhausted trying to fulfill the desired configuration")
+					delay = defaultDelay
```
rshriram: So what do we do in this case? Thinking out aloud, it depends on the error. what is the point of running the agent if we know that we couldn't apply the config? For all we know, we could have N instances of envoy running or something failing. Or the error may resolve itself.

proxy/agent.go
```diff
@@ -0,0 +1,198 @@
...
+				return
+			}
+		}
+	}
+}
+
+func (a *agent) reconcile() {
+	// check that the config is current
+	if reflect.DeepEqual(a.desired, a.current) {
+		glog.V(2).Info("Desired configuration is already applied")
```
rshriram: May be have a md5 hash or something to indicate the config version.. or print out the current epoch as well. I would put this as V(3), as this message is bound to flood the debug dump.

proxy/envoy/resources.go
```diff
@@ -24,27 +24,53 @@ import (
...
 var (
 	// DefaultMeshConfig configuration
 	DefaultMeshConfig = &MeshConfig{
-		DiscoveryAddress: "manager:8080",
-		MixerAddress:     "mixer:9091",
-		ProxyPort:        15001,
-		AdminPort:        15000,
-		BinaryPath:       "/usr/local/bin/envoy",
-		ConfigPath:       "/etc/envoy",
+		DiscoveryAddress:          "manager:8080",
```
rshriram: istio-manager

proxy/envoy/resources.go
```diff
@@ -24,27 +24,53 @@ import (
...
+		MixerAddress:              "mixer:9091",
```
rshriram: istio-mixer

rshriram:
@kyessenov my comments pertain to just the logic. I suggest asking someone else for the go routine stuff.

ayj:
I'll take a look at the goroutine stuff later tonight if someone else doesn't first.

proxy/agent.go
kyessenov: I prefer "epoch". The comment serves as an example of how to use the agent, I'll make that more clear.

proxy/agent.go
kyessenov: Yes. I wanted it to mean "drive towards the config state I supply". Update seems to imply it's immediate and guaranteed.

proxy/agent.go
kyessenov: cleanup is specific to the proxy implementation.

proxy/agent.go
kyessenov: It's configurable for the purpose of testing.

proxy/agent.go
kyessenov: Between retries (times 2 for back-off).

proxy/agent.go
kyessenov: The update is the entire config. Channels block on sender side, and it's expected that the config is the freshest, not at the time of the event (see comment in the watcher). Our model controller provides this guarantee.

proxy/agent.go
kyessenov: We'd need to update some monitoring status to say that the config permanently fails and the system is inconsistent.

proxy/agent.go
kyessenov: Config is being printed as a whole already (see `watcher.go`).

proxy/envoy/resources.go
kyessenov: This should be done in another PR.

proxy/envoy/resources.go
kyessenov: Again, this is for another PR.

kyessenov:
@rshriram I'll take another pass over documentation, which seems to be the main concern in your review. I will wait for the code review comments.

proxy/agent.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (a *agent) Apply(config interface{}) {
+	a.config <- config
+}
+
+// Run executes the reloads until stop channel closes
+func (a *agent) Run(stop <-chan struct{}) {
+	glog.V(2).Info("Starting proxy agent")
+
+	// Throttle processing up to smoothed 1 qps with bursts up to 10 qps
```
ayj: Do we really need to update configuration 10 times per second? Do we have any idea of expectations for routing traffic after a new service is added, i.e. how fast do adopt new configuration?

proxy/agent.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (a *agent) reconcile() {
+	// check that the config is current
+	if reflect.DeepEqual(a.desired, a.current) {
+		glog.V(2).Info("Desired configuration is already applied")
+		return
+	}
+
+	// discover and increment the latest running epoch
+	epoch := a.latestEpoch() + 1
```
ayj: It looks like reconcile() will increment `epoch` regardless of whether the previous envoy instance started successfully or not. If so, does envoy handle this case correctly or does it assume epoch+1 is only valid for already running instances?

proxy/agent_test.go
```diff
@@ -0,0 +1,226 @@
...
+// in between.
+func TestApplyThrice(t *testing.T) {
+	stop := make(chan struct{})
+	good := "good"
+	bad := "bad"
+	applied := false
+	var a Agent
+	start := func(config interface{}, epoch int) error {
+		if config == bad {
+			return fmt.Errorf(bad)
```
ayj: nit: `return errors.New(bad)` here and elsewhere where string formatting is not needed.

proxy/agent.go
kyessenov: High QPS is to prevent blocking in the agent. Otherwise, the sender, which is the k8s controller in this case, would be blocked for each configuration update, even if the configuration does not change.

proxy/agent.go
kyessenov: The typical bad case that this is trying to solve is: epoch 0 starts, epoch 1 starts and immediately crashes since epoch 0 has not initialized. The correct next epoch should be 1 -- it's always 1+last epoch. The last epoch belongs to whichever epoch owns the sockets, which  is always latest successfully started proxy. It could also happen that you start epoch 0, then epoch 1, then epoch 2, and then epoch 1 crashes. In that case, epoch 2 should crash as well, and it doesn't we need to fix Envoy.

proxy/agent_test.go
kyessenov: Thanks.

proxy/agent.go
geeknoid: Apply also seems pretty definitive. Perhaps ScheduleConfigUpdate?

kyessenov: Done.

proxy/agent.go
kyessenov: Done.

proxy/agent_test.go
kyessenov: Done

kyessenov:
@rshriram PTAL, I updated documentation mostly with minimal changes to the code.

kyessenov:
ingress test is very flaky...

PR #367, opened 2017-03-22
Resolve no ingress rule bug

ijsnellf:
Removes `omitempty` from the cluster_manager clusters field. Resolves https://github.com/istio/manager/issues/307

### DIFF
--- a/proxy/envoy/resources.go
+++ b/proxy/envoy/resources.go
@@ -596,7 +596,7 @@ type RDS struct {
 
 // ClusterManager definition
 type ClusterManager struct {
-	Clusters []*Cluster `json:"clusters,omitempty"`
+	Clusters []*Cluster `json:"clusters"`
 	SDS      *SDS       `json:"sds,omitempty"`
 	CDS      *CDS       `json:"cds,omitempty"`
 }
--- a/test/integration/ingress-proxy.yaml.tmpl
+++ b/test/integration/ingress-proxy.yaml.tmpl
@@ -8,17 +8,10 @@ spec:
   ports:
   - name: https
     port: 443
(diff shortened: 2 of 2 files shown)


zcahana:
Fix looks good, but build failed. I couldn't tell what exactly caused the error... any idea?

zcahana:
Btw, would be nice if we can validate (even manually) that this fix indeed produces a valid Envoy configuration (instead of moving to the next error).

ijsnellf:
I am hitting this failure in the check:
```I0322 14:20:03.054801    7798 driver.go:320] kubectl apply -f istio-ingress-ingress-proxy.yaml -n istio-integration-6rg7c
deployment "ingress" created
The Service "ingress" is invalid: spec.ports[0].nodePort: Invalid value: 32443: provided port is already allocated```

This is probably because setting up the node port is a node-wide operation, so it can interfere with other jobs. The tests don't use the node port, so I will remove it.

zcahana:
Do we have concurrent tests running on the same cluster?

ijsnellf:
My understanding was that the tests are run on a gcloud cluster.

ijsnellf:
^^^I did manually verify that this change fixes the bad schema issue when there are no ingress rules.

zcahana:
LGTM, given that the nodePort isn't used in the test. Does it mean that the ingress service is hit from within the cluster, or isn't exercised at all in the test?

ijsnellf:
It is hit from within the cluster: https://github.com/istio/manager/blob/master/test/integration/reachability.go#L298. We are only testing that the Ingress controller is proxying correctly, not that it is exposed externally.

PR #374, opened 2017-03-22
Use istio.io/manager/platform/kube/inject in integration tests.

ayj:
Use the inject package to dynamically inject the runtime proxy into the
integration tests YAML resource files. Remove the istio-specific app
template and use the same app.yaml.tmpl for istio and non-istio
services.


test/integration/driver.go
```diff
@@ -136,18 +137,18 @@ func setup() {
...
 
 	// deploy istio-infra
-	check(deploy("http-discovery", "http-discovery", managerDiscovery, "8080", "80", "9090", "90", "unversioned"))
-	check(deploy("mixer", "mixer", mixer, "8080", "80", "9090", "90", "unversioned"))
-	check(deploy("istio-egress", "istio-egress", egressProxy, "8080", "80", "9090", "90", "unversioned"))
-	check(deploy("istio-ingress", "istio-ingress", ingressProxy, "8080", "80", "9090", "90", "unversioned"))
+	check(deploy("http-discovery", "http-discovery", managerDiscovery, "8080", "80", "9090", "90", "unversioned", false))
+	check(deploy("mixer", "mixer", mixer, "8080", "80", "9090", "90", "unversioned", false))
+	check(deploy("istio-egress", "istio-egress", egressProxy, "8080", "80", "9090", "90", "unversioned", false))
+	check(deploy("istio-ingress", "istio-ingress", ingressProxy, "8080", "80", "9090", "90", "unversioned", false))
```
kyessenov: nit: I think we can collapse all these files into one deployment and not use `deploy` function, just like in the demo

test/integration/driver.go
```diff
@@ -201,7 +201,30 @@ func deploy(name, svcName, dType, port1, port2, port3, port4, version string) er
 		return err
 	}
 
-	if err := w.Flush(); err != nil {
+	writer := bufio.NewWriter(f)
+	if injectProxy {
+		p := &inject.Params{
+			InitImage:        fmt.Sprintf("%s/init:%s", params.hub, params.tag),
+			RuntimeImage:     fmt.Sprintf("%s/runtime:%s", params.hub, params.tag),
+			RuntimeVerbosity: 2,
```
kyessenov: can you add a verbosity field at the top since it's used in two places, here and in `write` function?

test/integration/driver.go
kyessenov: this should be done in a separate PR

PR #376, opened 2017-03-23
Define secret registry

kyessenov:
Basic interface to access secrets to prevent the use of k8s library directly in the proxy config gen.
Also, fixes a couple of bugs with ingress: default namespace and retries.


rshriram:
Can you add some description to the PR? What is this doing and how does this relate to auth pr?

PR #380, opened 2017-03-23
Use consistent naming for istio components

ayj:
Documentation and examples in github.com/istio/istio use `istio-mixer`,
`istio-manager`, and `istio-ingress-controller`. Use the same naming
in the manager e2e integration tests.

Also update `istioctl kube-inject` to accept manager and mixer
addresses as host:port instead of only port number.


ayj:
@kuat, what was the alternative to *discovery*? *manager*? I don't feel too strongly about the name so long as we use it consistently.

`inject.go` can be updated to use default names from whatever central place you add, so long as it's available in manager repo.

kyessenov:
@andraxylia, have any proposal for a better name?

andraxylia:
Indeed, we converged into calling it "istio-manager" or simply manager. It would be good to use ManagerAddr instead of DiscoveryAddr, etc.

PR #387, opened 2017-03-24
Keep ingress around

kyessenov:
Fix a bug that was failing the post-submit. Ingress should be deleted once, not at every iteration.

### DIFF
--- a/test/integration/driver.go
+++ b/test/integration/driver.go
@@ -103,6 +103,10 @@ func main() {
 		check((&reachability{}).run())
 		check(testRouting())
 	}
+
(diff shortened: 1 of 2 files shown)


test/integration/driver.go
```diff
@@ -103,6 +103,10 @@ func main() {
 		check((&reachability{}).run())
 		check(testRouting())
 	}
+
+	glog.Info("Cleaning up ingress secret.")
```
ayj: Perhaps move into teardown() so ingress secret is cleaned up if check() calls fail?

kyessenov: done

PR #388, opened 2017-03-24
istioctl: use --tag/--hub instead --initImage/--runtimeImage

ayj:
Use --tag and --hub instead of --initImage and --runtimeImage flags.
This makes usage a bit more concise because the hub is unlikely to
change and the tag is common between init and runtime image.
Furthermore, it is consistent with the manager integration test flags.


platform/kube/inject/inject.go
```diff
@@ -37,8 +37,8 @@ import (
 // Defaults values for injecting istio proxy into kubernetes
 // resources.
 const (
-	DefaultInitImage            = "docker.io/istio/init:latest"
-	DefaultRuntimeImage         = "docker.io/istio/runtime:latest"
+	DefaultHub                  = "docker.io"
+	DefaultTag                  = "2017-03-22-17.30.06"
```
kyessenov: Maybe use build version here?

ayj: That would work, but maybe we should wait until we have full acceptance test in place which verifies latest version of manager, proxy, and mixer together with a matching version of istioctl?

PR #392, opened 2017-03-24
add script to cross-compile istioctl for osx and windows

ayj:


bin/cross-compile-istioctl
```diff
@@ -0,0 +1,29 @@
...
+while read line; do
+  read SYMBOL VALUE < <(echo $line)
+  LDFLAGS=${LDFLAGS}" -X istio.io/manager/cmd/version.${SYMBOL}=${VALUE}"
+done < <(./bin/get_workspace_status)
+
+GOOS=darwin GOARCH=amd64 go build -ldflags "${LDFLAGS}" istio.io/manager/cmd/istioctl
+mv istioctl istioctl-osx
+
+# install missing spf13/cobra dependency (windows only)
+go get github.com/inconshreveable/mousetrap
```
kyessenov: ugh, that's probably because bazel ignores build pragmas...

ayj: Looks like the opposite case. bazel *is* observing build flags (or more accurately, golang toolchain is setting build flags correctly) and bazel is only building for linux so `windows` build flag isn't defined. Not usually a problem unless you decide to build non-linux golang binaries outside of bazel ...


```golang
$ head vendor/github.com/spf13/cobra/command_win.go
// +build windows

package cobra

import (
        "os"
        "time"

        "github.com/inconshreveable/mousetrap"
)
```

PR #394, opened 2017-03-24
Change service encoding scheme to satisfy Envoy

kyessenov:
Envoy is enforcing schema on cluster names and that means ":" is no longer allowed.

### DIFF
--- a/proxy/envoy/route.go
+++ b/proxy/envoy/route.go
@@ -31,10 +31,10 @@ import (
 
 const (
 	// InboundClusterPrefix is the prefix for service clusters co-hosted on the proxy instance
-	InboundClusterPrefix = "in:"
+	InboundClusterPrefix = "in."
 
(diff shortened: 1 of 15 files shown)


proxy/envoy/route.go
```diff
@@ -31,10 +31,10 @@ import (
 
 const (
 	// InboundClusterPrefix is the prefix for service clusters co-hosted on the proxy instance
-	InboundClusterPrefix = "in:"
+	InboundClusterPrefix = "in."
```
ayj: Why `.` and not `|` here? 

Also, does it make sense to parameterize the separator? e.g.

```golang
const ServiceSeperator = "|"
```

kyessenov: I'll revisit it next week, once Matt decides what the valid cluster names are. We'll need to rename all clusters with synthetic short names anyways.

PR #397, opened 2017-03-24
istioctl: use correct default docker hub name

ayj:


platform/kube/inject/inject_test.go
```diff
@@ -24,12 +24,12 @@ import (
 )
 
 func TestImageName(t *testing.T) {
-	want := "docker.io/init:latest"
-	if got := InitImageName("docker.io", "latest"); got != want {
+	want := "docker.io/istio/init:latest"
```
andraxylia: Re-use defaultHub if you can.

PR #399, opened 2017-03-25
Stop reconciliation after success

kyessenov:


proxy/agent.go
```diff
@@ -175,8 +175,14 @@ func (a *agent) Run(stop <-chan struct{}) {
 			}
 
 		case <-time.After(delay):
-			glog.V(2).Infof("Reconciling after delay %v", delay)
-			a.reconcile()
+			if reflect.DeepEqual(a.desired, a.current) {
+				glog.V(2).Info("Reconciliation not needed")
+				a.retryBudget = a.maxRetries
```
andraxylia: reconcile already handles the DeepEqual, move reset of retryBudget and delay in reconcile, he and in the above case. They should all be in one place.

proxy/agent.go
```diff
@@ -164,9 +164,9 @@ func (a *agent) Run(stop <-chan struct{}) {
 			// schedule a retry for a transient error
 			if status.err != nil && !reflect.DeepEqual(a.desired, a.current) {
 				if a.retryBudget > 0 {
-					delay = a.retryInterval * (2 << uint(a.maxRetries-a.retryBudget))
+					delay = a.retryInterval * (1 << uint(a.maxRetries-a.retryBudget))
```
andraxylia: Initialize retryBudget with maxRetries when agent is initialized, there is no guarantee case config is handled before case status.

andraxylia: Do we have to retry with the same epoch for the case status, if the epoch exited normally?

proxy/agent.go
andraxylia: Managers needs to exit if maxRetries exhausted, so the pod can be re-started. Having envoy not running while manager keeps running is a big issue. This can be addressed in another PR.

kyessenov:
PTAL, I moved delay to be a property of the agent. The permanent failure case should be handled in another PR.

kyessenov:
The tricky case is when to reset the retry budget. I decided to do it only when a new config is sent, rather than whenever reconciliation succeeds, since the proxy might not initialize within the timeframe of the initial retry delay.

proxy/agent.go
```diff
@@ -66,27 +67,58 @@ type Agent interface {
...
 	run func(interface{}, int) error
 
 	// proxy cleanup command
 	cleanup func(int)
 
+	// retry configuration
+	retry *retry
+
 	// desired configuration state
 	desired interface{}
```
andraxylia: The name of "config" field in agent is bad for a channel. Can you rename it to something else, also rename desired --> desiredConf, current-->currentConf.

kyessenov: Done, thanks.

proxy/agent.go
```diff
@@ -129,25 +151,22 @@ func (a *agent) Run(stop <-chan struct{}) {
...
 
 		select {
-		case config := <-a.config:
-			// reset retry budget if and only if the desired config changes
-			if !reflect.DeepEqual(a.desired, config) {
-				a.retryBudget = a.maxRetries
-				delay = defaultDelay
-				a.desired = config
+		case config := <-a.configCh:
+			if !reflect.DeepEqual(a.desiredConfig, config) {
```
andraxylia: This is exercised when a new config must be sent to envoy. Shouldn't it compare config with currentConfig, and set desiredConfig to config?

kyessenov: The client of the agent doesn't know if the new config "must" be sent. The way it is set up, config is generated every time the inputs change, and the config gets sent here. It is up to the agent to decide if it "must" be applied or not, and this check is doing that.

kyessenov: The granularity of watches is not good right now, which means we might be generating a lot of identical configs. We can detect finer deltas with more work in the watcher.

andraxylia: Makes sense.

kyessenov:
#336

proxy/agent.go
```diff
@@ -159,23 +178,22 @@ func (a *agent) Run(stop <-chan struct{}) {
...
 					// TODO: update proxy agent monitoring status about this error
-					delay = defaultDelay
+					a.retry.delay = a.retry.defaultDelay
 				}
 			}
 
-		case <-time.After(delay):
-			glog.V(2).Infof("Reconciling after delay %v", delay)
+		case <-time.After(a.retry.delay):
+			glog.V(2).Infof("Reconciling after delay %v (budget %d)", a.retry.delay, a.retry.budget)
```
andraxylia: Shouldn't retry.budget be decremented here as well?
Best is to move the logic that changes the values in retry inside reconcile(), so they get adjusted after each reconcile().

PR #401, opened 2017-03-26
Use IP address in the listeners

kyessenov:
Change config generation to use three types of listeners:
1. inbound listeners, do not use CDS, RDS, SDS; bind to pod IP, port port pair.
2. outbound TCP listeners, use only SDS, bind to service IP, service port pair
3. outbound HTTP listeners, use CDS, RDS, SDS; bind to 0.0.0.0 and share ports

This PR touches a lot of cases, so please help reviewing this. This is needed to enable auth on inbound HTTP listeners.

### DIFF
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"sort"
 
 	"github.com/golang/glog"
 	multierror "github.com/hashicorp/go-multierror"
@@ -125,7 +124,10 @@ func buildListeners(context *ProxyContext) (Listeners, Clusters) {
 
 	inbound, inClusters := buildInboundListeners(instances, context.MeshConfig)
 	outbound, outClusters := buildOutboundListeners(instances, services, context)
-	listeners := append(inbound, outbound...).normalize()
+
+	listeners := append(inbound, outbound...)
+	listeners.normalize()
+
 	clusters := append(inClusters, outClusters...).normalize()
 
 	// inject Mixer filter with proxy identities
@@ -190,48 +192,24 @@ func buildTCPListener(tcpConfig *TCPRouteConfig, ip string, port int) *Listener 
 	}
 }
 
-// buildOutboundListeners combines HTTP and TCP outbound routes and creates listeners for them.
-// Returns TCP clusters
+// buildOutboundListeners combines HTTP routes and TCP listeners
 func buildOutboundListeners(instances []*model.ServiceInstance, services []*model.Service,
 	context *ProxyContext) (Listeners, Clusters) {
-	listeners := make(Listeners, 0)
-	clusters := make(Clusters, 0)
-	httpOutbound, tcpOutbound := buildOutboundRoutes(instances, services, context)
+	httpOutbound := buildOutboundHTTPRoutes(instances, services, context)
+	listeners, clusters := buildOutboundTCPListeners(services)
 
 	for port, routeConfig := range httpOutbound {
-		// TODO(github.com/istio/manager/issues/237)
-		//
-		// Sharing tcp_proxy and http_connection_manager filters on
-		// the same port for different destination services doesn't work
-		// with Envoy (yet). When the tcp_proxy filter's route matching
-		// fails for the http service the connection is closed without
-		// falling back to the http_connection_manager.
-		//
-		// Temporary workaround is to not share ports between
-		// destination services. If the user does share ports, remove the
-		// TCP service from the envoy config and print a warning.
-		if config, ok := tcpOutbound[port]; ok {
-			glog.Warningf("TCP and HTTP services on same port not supported")
-			glog.Warningf("Omitting tcp service %v on port %v", config, port)
-			delete(tcpOutbound, port)
-		}
 		listeners = append(listeners, buildHTTPListener(routeConfig, WildcardAddress, port, true))
-	}
-
-	for port, tcpConfig := range tcpOutbound {
-		clusters = append(clusters, tcpConfig.clusters()...)
-		listeners = append(listeners, buildTCPListener(tcpConfig, WildcardAddress, port))
 	}
 
 	return listeners, clusters
 }
 
-// buildOutboundRoutes creates route configs indexed by ports for the traffic outbound
-// from the proxy instance
-func buildOutboundRoutes(instances []*model.ServiceInstance, services []*model.Service,
-	context *ProxyContext) (HTTPRouteConfigs, TCPRouteConfigs) {
+// buildOutboundHTTPRoutes creates HTTP route configs indexed by ports for the
+// traffic outbound from the proxy instance
+func buildOutboundHTTPRoutes(instances []*model.ServiceInstance, services []*model.Service,
+	context *ProxyContext) HTTPRouteConfigs {
 	httpConfigs := make(HTTPRouteConfigs)
-	tcpConfigs := make(TCPRouteConfigs)
 
 	// used for shortcut domain names for outbound hostnames
 	suffix := sharedInstanceHost(instances)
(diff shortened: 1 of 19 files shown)


test/integration/driver.go
```diff
@@ -139,11 +139,10 @@ func setup() {
 	pods = make(map[string]string)
 
 	// setup ingress resources
-	_, err = shell(fmt.Sprintf("kubectl -n %s create secret generic ingress "+
+	_, _ = shell(fmt.Sprintf("kubectl -n %s create secret generic ingress "+
```
andraxylia: Add check(err) right after.

proxy/envoy/testdata/envoy-v0.json.golden
```diff
@@ -1,7 +1,53 @@
 {
   "listeners": [
     {
-      "address": "tcp://0.0.0.0:1081",
+      "address": "tcp://0.0.0.0:80",
```
andraxylia: In the golden files, can we add comment lines describes each type of listener and filter? It is hard to understand otherwise what each listener and filter does.

andraxylia:
Can you link this PR to issue or proposal describing the design? It is hard to understand otherwise.

proxy/envoy/testdata/envoy-v0.json.golden
ijsnellf: We'd have to change the way we were checking against our golden files. Right now we do a simple diff between the golden files and the generated files to check for issues.

proxy/envoy/config.go
```diff
@@ -116,54 +116,94 @@ func Generate(context *ProxyContext) *Config {
 	}
 }
 
-// build combines the outbound and inbound routes prioritizing the latter
+// buildListeners produces a list of listeners and referenced clusters
 // build also returns all inbound clusters, and outbound clusters referenced by TCP proxy
```
rshriram: typo: buildListeners also returns

proxy/envoy/config.go
```diff
@@ -116,54 +116,94 @@ func Generate(context *ProxyContext) *Config {
...
+	sort.Slice(listeners, func(i, j int) bool { return listeners[i].Address < listeners[j].Address })
+	clusters := append(inClusters, outClusters...).Normalize()
+
+	return listeners, clusters
+}
+
+// buildHTTPListener constructs a listener for the network interface address and port
+// Use "0.0.0.0" IP address to listen on all interfaces
+// RDS parameter controls whether to use RDS for the route updates.
+func buildHTTPListener(routeConfig *HTTPRouteConfig, ip string, port int, rds bool) *Listener {
```
rshriram: Why do we need to have an optional RDS? why can't it be the only option?

proxy/envoy/testdata/envoy-v0.json.golden
myidpt: +1

proxy/envoy/config_test.go
```diff
@@ -82,12 +82,12 @@ func TestRoutesByPath(t *testing.T) {
 func TestTCPRouteConfigByRoute(t *testing.T) {
 	cases := []struct {
 		name string
-		in   []TCPRoute
-		want []TCPRoute
+		in   []*TCPRoute
```
wattli: What's the point of this change?

proxy/envoy/testdata/envoy-v0.json.golden
ayj: Comments aren't allowed in JSON files so they would need to be stripped before parsing
(see https://plus.google.com/+DouglasCrockfordEsq/posts/RK8qyGVaGSr). Furthermore, the golden files are currently auto-generated so we'd need to annotate them by hand. Alternatively, we could start encoding test data directly in *_test.go files as JSON string or the unmarshaled golang struct equivalent, either of which may make it easier to document via comments, though it would be more tedious to update/maintain the golden data itself.

proxy/envoy/config.go
kyessenov: Inbound http routes are not using RDS. They are just simple routes to localhost.

proxy/envoy/testdata/envoy-v0.json.golden
kyessenov: It's difficult to annotate them. This PR is invasive which should not be usual. I think it's easier to make sense when the change is smaller.

kyessenov:
I can send a follow up documentation and cleanup when I'm back. An additional test that exercises combined HTTP routes on a port would also help.

test/integration/driver.go
kyessenov: Left out deliberately so the test can run twice.

proxy/envoy/config_test.go
kyessenov: Change in the schema to use pointers like the rest of the schema.

proxy/envoy/testdata/envoy-v0.json.golden
andraxylia: The point is to have the info somewhere, it does not matter where.

andraxylia:
Is there an urgency to get this code in before we can review it?

wattli:
@andraxylia, yes, Auth is targeting on code complete by this week. And this is a PR that we are depending on.

andraxylia:
I see Isaiah approved, so it must be good to go.

If this unblocks auth, it's wiser to get it in.

proxy/envoy/config.go
```diff
@@ -116,172 +114,116 @@ func Generate(context *ProxyContext) *Config {
...
+			Config: TCPProxyFilterConfig{
+				StatPrefix:  "tcp",
+				RouteConfig: tcpConfig,
+			},
+		}},
+	}
+}
+
+// buildOutboundListeners combines HTTP routes and TCP listeners
+func buildOutboundListeners(instances []*model.ServiceInstance, services []*model.Service,
```
myidpt: Suggestion about the code structure:
Can you make this function similar to buildInboundListeners, by putting the loop inside the function and have buildOutboundHTTPRoutes and buildHTTPListener work together?

proxy/envoy/testdata/envoy-v0.json.golden
myidpt: Maybe adding a README file in the testdata dir, or have an md file in doc? I think this can be done separately from this PR.

test/integration/driver.go
kyessenov: linters require explicit assignments.

proxy/envoy/testdata/envoy-v0.json.golden
kyessenov: Test artifacts change so often that README file will not be sustainable. It's best to document it upfront in the main documentation "doc" directory.

proxy/envoy/config.go
kyessenov: We need HTTP routes for RDS, not the listeners.

wattli:
Do we have a tracking issue for the README doc mentioned in this thread?

wattli:
That will make the code much more readable, thanks.

andraxylia:
I will open tracking in istio/manager, so we have matching code and docs, although we may need to have it also in istio/istio.

kyessenov:
The translation is changing every two weeks, so it's best to document in the code.

wattli:
What is the translation? And why it changes every two weeks?

Btw, could you help to explain the second and third rules of that PR
<https://github.com/istio/manager/pull/401>? More specifically, why we only
use SDS for outbound TCP listener while use everything for HTTP listener?

- inbound listeners, do not use CDS, RDS, SDS; bind to pod IP, port port
pair.
- outbound TCP listeners, use only SDS, bind to service IP, service port
pair
- outbound HTTP listeners, use CDS, RDS, SDS; bind to 0.0.0.0 and share
ports

kyessenov:
Translation is from istio config to envoy config. Changes happen because envoy keeps changing schema/adding features, and because manager gets new requirements (like auth).

Regarding TCP treatment, that's because there's no RDS for TCP proxy, and CDS doesn't function without RDS.

ayj:
Envoy config churn should eventually slow down and be more predictable (see [config: create config breaking change policy](https://github.com/lyft/envoy/issues/625))

kyessenov:
We will write proper documentation on the config generation once the churn is over. Otherwise, we'll be changing config gen before documentation is ready :)

andraxylia:
I still need some documentation before the churn is over, otherwise I will not be able to review or see if things work as expected. Link to envoy configs are fine as well.

Issue #278, opened 2017-03-08
Ingress controller: testing


rshriram:
Test the ingress controller function, using both unit tests (where applicable) and e2e test.

ijsnellf:
Integration tests were added as part of https://github.com/istio/manager/pull/352

ijsnellf:
Unit tests and e2e tests are in place. We just need to maintain them as we add functionality now.

PR #405, opened 2017-03-28
Enhanced ingress class annotation processing

### LINKED ISSUE #371: Istio controller must process only k8s ingress resources annotated with "istio"
This is to allow co-existence of other ingress controller in a cluster.

zcahana:
This PR adds two flags to the ingress controller (`manager ingress` command):
- `--ingress_class="istio"` - The class of ingress resources to be processed by the ingress controller.
- `--default_ingress_controller=true` - Specifies whether running as the cluster's default ingress controller, thereby processing unclassified ingress resources.

Implementation note: the user-facing `--default_ingress_controller` flag is represented internally as an `IngressSyncMode=Off|Strict|Default` enum parameter which is passed down into the `kube.Controller`. The `Off` value disables ingress watching/processing altogether, and is used by the `manager sidecar` and `manager discovery` commands.

Resolves #371

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -29,10 +29,12 @@ import (
 )
 
 type args struct {
-	proxy         envoy.MeshConfig
-	identity      envoy.ProxyNode
-	sdsPort       int
-	ingressSecret string
+	proxy                    envoy.MeshConfig
+	identity                 envoy.ProxyNode
+	sdsPort                  int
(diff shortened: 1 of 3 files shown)


zcahana:
committed another patch with a unit-test for `kube.Controller` ingress handling, so that coverage goes :arrow_up: and not :arrow_down: with this PR. This also contributes to #278.

PR #408, opened 2017-03-29
Ingress controller SSL unit tests

ijsnellf:
Adds unit tests for the Envoy SSL config generation and does some minor refactoring to make it easier to test.


ijsnellf:
Contributes to https://github.com/istio/manager/issues/278.

cmd/manager/main.go
```diff
@@ -107,13 +107,17 @@ var (
...
 				IngressClass:    flags.ingressClass,
 			}
-			if flags.defaultIngressController {
-				controllerConfig.IngressSyncMode = kube.IngressDefault
-			}
 			controller := kube.NewController(cmd.Client, controllerConfig)
-			w, err := envoy.NewIngressWatcher(controller, controller,
-				&model.IstioRegistry{ConfigRegistry: controller},
-				cmd.Client, &flags.proxy, flags.ingressSecret, cmd.RootFlags.Namespace)
+			context := &envoy.IngressContext{
```
kyessenov: I think `IngressConfig` is more appropriate here - proxy context is used to differentiate  multiple envoy instances, ingress proxies seem to act the same way. We can make a more general proxy context later that includes "ingress role" for the proxy, if you want.

proxy/envoy/ingress.go
```diff
@@ -179,44 +185,26 @@ func generateIngress(namespace, secret string, secrets model.SecretRegistry,
 		Listeners: listeners,
 		Admin: Admin{
 			AccessLogPath: DefaultAccessLog,
-			Address:       fmt.Sprintf("tcp://0.0.0.0:%d", mesh.AdminPort),
+			Address:       fmt.Sprintf("tcp://0.0.0.0:%d", context.Mesh.AdminPort),
```
kyessenov: nite: there is `WildcardAddress` that can be used for `0.0.0.0`

cmd/manager/main.go
ijsnellf: I agree, that is better.

proxy/envoy/ingress.go
ijsnellf: fixed

PR #411, opened 2017-03-29
istioctl: add option to enable core dumps in injected runtime proxy

ayj:
Enable core dumps in the runtime proxy for better always-on debugging
experience. "sysctl -w kernel.core_pattern=/tmp/core.%e.%p.%t"
requires additional privileges so use separate init-container to avoid
leaking elevated privileges into non-debug files/images.

This commit adds the `--coreDump` flag to control whether core dumps
are enabled or not and defaults to true for pre-alpha development.

### DIFF
--- a/cmd/istioctl/inject.go
+++ b/cmd/istioctl/inject.go
@@ -36,6 +36,7 @@ var (
 	sidecarProxyPort int
 	runtimeVerbosity int
 	versionStr       string // override build version
+	enableCoreDump   bool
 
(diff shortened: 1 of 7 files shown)


ayj:
ref https://github.com/istio/istio/pull/71 (deprecated)
ref https://github.com/istio/manager/issues/364 (fixes)

ayj:
Fixes https://github.com/istio/manager/issues/364

PR #413, opened 2017-03-29
Add sslcontext to inbound http listener

wattli:


wattli:
@kyessenov @myidpt @lookuptable

proxy/envoy/route.go
```diff
@@ -36,9 +36,22 @@ const (
 	OutboundClusterPrefix = "out."
 )
 
-// buildSSLContextWithSAN returns an SSLContextWithSAN struct with VerifySubjectAltName when auth is enabled.
-// Otherwise, it returns nil.
-func buildSSLContextWithSAN(hostname string, context *ProxyContext) *SSLContextWithSAN {
+// buildListenerSSLContext returns an SSLContext struct when auth is enabled.
+// Otherwise, it returns nil. This is used in listener SSLContext.
+func buildListenerSSLContext(mesh *MeshConfig) *SSLContext {
```
wattli: @myidpt , I intend to call this function "buildListener(...)". But turns out the name is occupied in route.go. So I call it "buildListenerSSLContext(...)", and to make it consistent, I renamed the existing buildSSLContextWithSAN(...) to buildClusterSSLContext(...). Thanks.

proxy/envoy/config.go
```diff
@@ -167,8 +167,17 @@ func buildHTTPListener(routeConfig *HTTPRouteConfig, ip string, port int, rds bo
 		config.RouteConfig = routeConfig
 	}
 
+	var sslContext *SSLContext
+	// Build the sslContext for http inbound listener (only inbound listener does not use rds)
+	if !rds {
+		sslContext = buildListenerSSLContext(mesh)
```
lookuptable: The logic here is a little confusing to me, particularly since `buildListenerSSLContext()` can potentially returns nil. Can we do something like this:
```
if !rds && mesh.authEnabled() {
 sslContext = buildListenerSSLContext()
}
```

proxy/envoy/config.go
```diff
@@ -167,8 +167,17 @@ func buildHTTPListener(routeConfig *HTTPRouteConfig, ip string, port int, rds bo
...
+	} else {
+		sslContext = nil
```
lookuptable: Guess you don't need this since it's initialized to be nil

proxy/envoy/config.go
kyessenov: +1. Please do not use nils. All functions should return actual values. I know that @myidpt did the opposite, but I let it slip to let that PR in faster.

wattli: Done.

proxy/envoy/config.go
wattli: Done

proxy/envoy/route.go
myidpt: OK. I'm fine with that.

wattli:
I don't think so

Oliver Liu <notifications@github.com>于2017年3月29日 周三下午3:44写道：

proxy/envoy/testdata/envoy-v0-auth.json.golden
```diff
@@ -0,0 +1,416 @@
...
+                }
+              ]
+            }
+          }
+        }
+      ],
+      "bind_to_port": false
+    },
+    {
+      "address": "tcp://10.1.1.0:80",
```
myidpt: @kyessenov (the question is not related to this PR) why do we use 10.1.1.0 which is an invalid IP if it is not a wildcard? Maybe using something like 10.1.1.2?

proxy/envoy/config.go
```diff
@@ -167,8 +167,15 @@ func buildHTTPListener(routeConfig *HTTPRouteConfig, ip string, port int, rds bo
 		config.RouteConfig = routeConfig
 	}
 
+	var sslContext *SSLContext
+	// Build the sslContext for http inbound listener (only inbound listener does not use rds)
+	if !rds && mesh.EnableAuth {
```
kyessenov: Why is RDS affecting SSL? We may want to use auth without RDS.

myidpt: We need a way to differentiate the inbound vs outbound traffic. How do we differentiate them if not by RDS?

wattli: The previous PR set rds to true for outbound http listener and false for inbound listener. I just follow the convention. 

Yes, the current logic uses auth without RDS.  

If rds check is not safe, we have two potential solutions 
- Add another bool to indicate whether it's a inbound listener or outbound
- Check whether the IP is wildcard. But that may not be safe either since an inbound listener may have a wildcard address (right?)

kyessenov: we need another parameters to use auth then. RDS can be used for inbound traffic (weird, but possible).

wattli: Done.

proxy/envoy/config.go
```diff
@@ -220,7 +227,11 @@ func buildOutboundHTTPRoutes(instances []*model.ServiceInstance, services []*mod
 	// outbound connections/requests are directed to service ports; we create a
 	// map for each service port to define filters
 	for _, service := range services {
-		sslContext := buildSSLContextWithSAN(service.Hostname, context)
+		var sslContext *SSLContextWithSAN
+		// Build the sslContext for http inbound listener (only inbound listener does not use rds)
```
lookuptable: I don't think it's related to rds anymore. Please update the comment.

proxy/envoy/route.go
```diff
@@ -36,20 +36,27 @@ const (
 	OutboundClusterPrefix = "out."
 )
 
-// buildSSLContextWithSAN returns an SSLContextWithSAN struct with VerifySubjectAltName when auth is enabled.
-// Otherwise, it returns nil.
-func buildSSLContextWithSAN(hostname string, context *ProxyContext) *SSLContextWithSAN {
+// buildListenerSSLContext returns an SSLContext struct when auth is enabled.
+// Otherwise, it returns nil. This is used in listener SSLContext.
```
lookuptable: Update comment here as well please.

proxy/envoy/route.go
```diff
@@ -36,20 +36,27 @@ const (
...
+func buildListenerSSLContext(mesh *MeshConfig) *SSLContext {
+	return &SSLContext{
+		CertChainFile:  mesh.AuthConfigPath + "/cert-chain.pem",
+		PrivateKeyFile: mesh.AuthConfigPath + "/key.pem",
+		CaCertFile:     mesh.AuthConfigPath + "/root-cert.pem",
+	}
+}
+
+// buildClusterSSLContext returns an SSLContextWithSAN struct with VerifySubjectAltName when auth is enabled.
+// Otherwise, it returns nil. This is used in cluster SSLContext
```
lookuptable: ditto

proxy/envoy/config.go
wattli: Done.

proxy/envoy/route.go
wattli: Done.

proxy/envoy/route.go
wattli: Done.

wattli:
@kyessenov , can you help to merge this PR please? I don't have the permission.

PR #414, opened 2017-03-29
istioctl get form to list rules or policies

esnible:


cmd/istioctl/main.go
```diff
@@ -183,6 +213,10 @@ var (
 		Use:   "list <type>",
```
rshriram: why do we need to support list? Since we don't have a release, lets not worry about deprecation

rshriram:
@esnible Sorry, I am out of sync with this piece of codebase, to do a proper review atm.
@ayj can you please take a look?

cmd/istioctl/main.go
```diff
@@ -299,15 +306,15 @@ func readInputs() ([]inputDoc, error) {
...
 		if !ok {
 			return nil, fmt.Errorf("unknown spec type %s", v.Type)
 		}
-		rr, err := schema.FromJSON(string(byteRule))
+		rr, err := ischema.FromJSON(string(byteRule))
 		if err != nil {
 			return nil, fmt.Errorf("cannot parse proto message: %v", err)
 		}
-		glog.V(2).Info(fmt.Sprintf("Parsed %v %v into %v %v", v.Type, v.Name, schema.MessageName, rr))
+		glog.V(2).Info(fmt.Sprintf("Parsed %v %v into %v %v", v.Type, v.Name, ischema.MessageName, rr))
```
ayj: You can use `glog.V(2).Infof` directly and do away with the extra call to `fmt.Sprintf`. I realize this code was here before, but it would be good to clean-up.

cmd/istioctl/main.go
```diff
@@ -299,15 +306,15 @@ func readInputs() ([]inputDoc, error) {
 			return nil, fmt.Errorf("could not encode Spec: %v", err)
 		}
 
-		schema, ok := model.IstioConfig[v.Type]
+		ischema, ok := model.IstioConfig[v.Type]
```
ayj: The use of global variables (e.g. `schema`) makes this code harder to follow. It would be good to clean this up at some point, possibly in this PR or a follow-up PR.

cmd/istioctl/main.go
```diff
@@ -316,3 +323,36 @@ func readInputs() ([]inputDoc, error) {
...
+	}
+
+	return nil
+}
+
+// Print as YAML
+func printYamlOutput(list map[model.Key]proto.Message) error {
+	for key, item := range list {
+		out, err := schema.ToYAML(item)
+		if err != nil {
```
ayj: Probably better to return the `err` if `schema.ToYAML` failed so error can be propagated to os.Exit. Otherwise, the program would exit successfully printing non-valid output.

cmd/istioctl/istioctl_test.go
```diff
@@ -99,3 +99,70 @@ func TestBogusExplicitKubeConfig(t *testing.T) {
...
+
+func TestNewGet(t *testing.T) {
+	rootSetup(t)
+
+	file = "testdata/dest-policy.yaml"
+	if err := postCmd.RunE(postCmd, []string{}); err != nil {
+		t.Fatalf("Could not create destination policy: %v", err)
+	}
+
+	file = ""
```
ayj: Some of the `getCmd` code could be converted to table-driven tests (see https://github.com/golang/go/wiki/TableDrivenTests) to avoid duplication.

cmd/istioctl/main.go
esnible: Agreed.  Let's do this in a follow-up PR.

cmd/istioctl/istioctl_test.go
esnible: I will look into that for future tests.

cmd/istioctl/main.go
```diff
@@ -316,3 +323,39 @@ func readInputs() ([]inputDoc, error) {
...
+}
+
+// Print as YAML
+func printYamlOutput(list map[model.Key]proto.Message) error {
+	var retVal error
+
+	for key, item := range list {
+		out, err := schema.ToYAML(item)
+		if err != nil {
+			retVal = err // Save the error, but keep going, so we can print the valid rules that follow
```
ayj: This will only report the most recent error. Perhaps you could use https://github.com/hashicorp/go-multierror if you want to collect all of the errors?

Also, I wouldn't print the error value here, but instead let the caller handle it accordingly, e.g. after normal output is printed and not interleaved with successful output.

PR #424, opened 2017-03-30
update config to match latest mixer config

ZackButcher:
This shouldn't be merged until the manager codebase starts using a mixer image from 3/30 or later.

### DIFF
--- a/test/integration/mixer.yaml.tmpl
+++ b/test/integration/mixer.yaml.tmpl
@@ -8,90 +8,12 @@ data:
     subject: "namespace:ns"
     revision: "2022"
     adapters:
-      - name: default
-        kind: quotas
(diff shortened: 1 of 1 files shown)


ZackButcher:
Meh, I don't have rights on this repo to assign reviewers or merge it later.

ZackButcher:
Yeah, if you don't want anything out of the mixer just delete all the config. A no-op config should still print out the return log line with the status and request ID.

ZackButcher:
I'll update this PR to change config so mixer is a no-op (but will still produce the log message you need).

kyessenov:
Yeah, that would require less work to keep it up to date.

ZackButcher:
PTAL

Issue #410, opened 2017-03-29
Split runtime image into "proxy" and "manager"


kyessenov:
Currently, the proxy agent and the discovery service are packaged together into a "runtime" image. I suggest we split it into two parts:
- proxy image that includes envoy, ingress controller, proxy agent and watcher
- manager image that includes the discovery service only.

This should clear some confusion regarding the docker image names.

I also recommend not publishing envoy proxy container image by itself and only ship it with the proxy agent inside it.

@rshriram @andraxylia @qiwzhang

PR #430, opened 2017-03-31
Split "runtime" image into "proxy" and "manager"

kyessenov:
Addresses #410


kyessenov:
We can split the binary if that makes sense in another PR. I can see there are two distinct commands:
* proxy (ingress, sidecar, egress roles)
* discovery service
They are tightly coupled though since the two work together.

PR #431, opened 2017-03-31
Update Manager testing workflows

sebastienvas:
- Create stable branch presubmit and postsubmit
- Move bazel.rc to tools dir
- update bin/cross-compile-istioctl to upload binaries


kyessenov:
Is there a particular reason to move bazelrc files to `tools`?

sebastienvas:
Yes this is a follow up from Lizan change in Proxy. https://github.com/istio/istio-testing/blob/master/src/org/istio/testutils/Bazel.groovy#L54 is expecting those file to be in tools. This allow user to still use their own .bazelrc

kyessenov:
Most of our devs don't even know what `.bazelrc` is. Let's not create a directory just to hold a single file.

sebastienvas:
mixer and proxy are already using this. Because of the way this is setup, all module needs to follow the same pattern.

Jenkinsfile
```diff
@@ -37,10 +41,8 @@ def presubmit(gitUtils, bazel, utils) {
       bazel.fetch('-k //...')
       bazel.build('//...')
     }
-    stage('Go Build') {
-      sh('bin/init.sh')
-    }
     stage('Code Check') {
+      sh('bin/init.sh')
```
kyessenov: let's keep these as separate stages so we can track the duration of each.

bin/cross-compile-istioctl
```diff
@@ -21,9 +42,14 @@ while read line; do
...
+mv istioctl "${BIN_DIR}/istioctl-osx"
 
 # install missing spf13/cobra dependency (windows only)
 go get github.com/inconshreveable/mousetrap
 GOOS=windows GOARCH=amd64 go build -ldflags "${LDFLAGS}" istio.io/manager/cmd/istioctl
-mv istioctl.exe istioctl-win.exe
+mv istioctl.exe "${BIN_DIR}/istioctl-win.exe"
+
+if [[ -n "${BUCKET_PATH}" ]]; then
+  gsutil -m cp -r "${BIN_DIR}"/istioctl-* "${BUCKET_PATH}" \
```
kyessenov: This script should be runnable by people not having a GCP account.

Jenkinsfile
```diff
@@ -52,12 +60,45 @@ def presubmit(gitUtils, bazel, utils) {
...
+}
+
+def stablePresubmit(gitUtils, bazel, utils) {
+  goBuildNode(gitUtils, 'istio.io/manager') {
+    bazel.updateBazelRc()
+    utils.initTestingCluster()
+    sh('ln -s ~/.kube/config platform/kube/')
+    stage('Integration Tests') {
+      timeout(30) {
+        sh("bin/e2e.sh -count 10 -debug -tag debug${gitUtils.GIT_SHA} -v 2")
```
kyessenov: This shouldn't be a pre-submit. This test is too flaky for that.

bin/cross-compile-istioctl
sebastienvas: they can run it if they don't provide the -p flag.

Jenkinsfile
sebastienvas: Done

Jenkinsfile
sebastienvas: I have to qualify on something. If the test fail then the next commit will be qualified. We don t have to qualify every commit.

PR #432, opened 2017-03-31
Use mock config for istioctl testing

kyessenov:
The goal of this PR is to avoid using real k8s cluster for testing `istioctl` as it is causing problems with concurrent tests operating on the same namespace. I split the root command into proxy/discovery and a separate istioctl version since the two seem to be diverging.


cmd/istioctl/main.go
```diff
@@ -55,6 +61,38 @@ var (
...
+			model.IstioConfig.Kinds()),
+		PersistentPreRunE: func(*cobra.Command, []string) (err error) {
+			if kubeconfig == "" {
+				if os.Getenv("KUBECONFIG") != "" {
+					glog.V(2).Infof("Setting configuration from KUBECONFIG environment variable")
+					kubeconfig = os.Getenv("KUBECONFIG")
+				}
+			}
+
+			var client *kube.Client
```
ayj: I'm curious, does the following give a linter warning about variable shadowing? It would seem cleaner and is allowed by spec:

```
client, err := kube.NewClient(...)
```

test/mock/config.go
```diff
@@ -73,8 +73,11 @@ func (cr *ConfigRegistry) Get(key model.Key) (proto.Message, bool) {
 
 // Delete implements config registry method
 func (cr *ConfigRegistry) Delete(key model.Key) error {
-	delete(cr.data, key)
-	return nil
+	if _, ok := cr.data[key]; ok {
+		delete(cr.data, key)
+		return nil
+	}
+	return fmt.Errorf("Item is missing")
```
kyessenov: done

cmd/cmd.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+)
+
+// AddFlags carries over glog flags with new defaults
+func AddFlags(rootCmd *cobra.Command) {
+	flag.CommandLine.VisitAll(func(gf *flag.Flag) {
+		switch gf.Name {
+		case "logtostderr":
+			err := gf.Value.Set("true")
```
kyessenov: done

cmd/istioctl/main.go
```diff
@@ -55,6 +61,38 @@ var (
...
 	schema model.ProtoSchema
 
+	rootCmd = &cobra.Command{
+		Use:   "istioctl",
+		Short: "Istio control interface",
+		Long: fmt.Sprintf("Istio configuration command line utility. Available configuration types: %v",
+			model.IstioConfig.Kinds()),
+		PersistentPreRunE: func(*cobra.Command, []string) (err error) {
+			if kubeconfig == "" {
+				if os.Getenv("KUBECONFIG") != "" {
```

cmd/istioctl/main.go

kyessenov:
Thanks, PTAL.

zcahana:
One comment:

The use of `/cmd/version/` package violates the Go project layout convention where each subdirectory of `/cmd/` is a `main` package which compiles into a binary.  I'd suggest we move `/cmd/version/version.go` to `/cmd/version.go`.

Otherwise LGTM.

ayj:
@zcahana, `version` is a utility package and isn't its own `main` package and therefore still violates  the go project layout convention you described =/ Maybe we need to adopt istio/mixer layout convention of having a `pkg` directory under which all non-main packages reside, e.g. `https://github.com/istio/mixer/tree/master/pkg/version`? But regardless, any change of that nature is outside the scope of this PR since `cmd/version/version` existed prior.

kyessenov:
I'm open to moving code to `pkg/` style of organization. We should do that with minimal disruption to folks working on PRs.

PR #438, opened 2017-03-31
enable pprof in manager discovery service

ayj:
Enable pprof in the manager discovery service and add /debug/pprof/*
http handlers to the service discovery port (i.e. 8080). Example
usage:

   kubectl port-forward <istio-manager-pod> 8080:8080
   go tool pprof http://localhost:8080/debug/pprof/profile

See https://golang.org/pkg/net/http/pprof/ for more details.


kyessenov (approved):
Nice. Should we have a flag to turn off the pprof endpoint? Seems rather dangerous to expose it in production.

kyessenov (approved):

PR #439, opened 2017-03-31
Parameterize envoy's RefreshDelayMs

ayj:
Add the --discovery_refresh_delay flag to configure the refresh delay
for SDS/CDS/RDS.


proxy/envoy/config.go
```diff
@@ -106,11 +106,11 @@ func Generate(context *ProxyContext) *Config {
 			Clusters: clusters,
 			SDS: &SDS{
 				Cluster:        buildDiscoveryCluster(mesh.DiscoveryAddress, "sds"),
-				RefreshDelayMs: 1000,
+				RefreshDelayMs: (int)(mesh.DiscoveryRefreshDelay.Nanoseconds() / 1e6),
```
kyessenov: nit: `int64(duration / time.Millisecond)` might be easier to understand. Duration is just int64 in nanoseconds (https://golang.org/pkg/time/#Duration)

ayj: I based this on Russ Cox's recommendation [here](https://github.com/golang/go/issues/5491#issuecomment-66079585). 

That results in mismatched types because `time.Millisecond` is of type `time.Duration` and golang doesn't do implicit type conversion (see https://play.golang.org/p/Qq1hSq6AVm)

ayj:
Oops, looks like my change to config.go wasn't included in original review due to bad merge. Fixed now. PTAL.

proxy/envoy/config.go
kyessenov: But this works: https://play.golang.org/p/RqfoBqeJwj

ayj: Ah, I misinterpreted your original comment. Updated based on your suggestion - PTAL.

Issue #428, opened 2017-03-30
CDS is unresponsive when `enable_auth` is true


lookuptable:
When setting `--enable_auth` for the discovery service, CDS becomes unresponsive. Initial investigation reveals that the call to kube-apiserver is somehow blocked (https://github.com/istio/manager/blob/master/platform/kube/controller.go#L648).

kyessenov:
@lookuptable feel free to re-open if you think the issue is still present.

PR #440, opened 2017-03-31
Read pods from local cache instead of from apiserver directly

lookuptable:


platform/kube/controller.go
```diff
@@ -630,34 +630,37 @@ const (
...
-	}
-	lo := v1.ListOptions{
-		LabelSelector: labels.Set(svc.Spec.Selector).String(),
-	}
-	// TODO: This is fragile, improve it.
-	pods, err := c.client.client.CoreV1().Pods(svc.Namespace).List(lo)
-	if err != nil {
-		glog.Warningf("Failed to get pods for service %s.", hostname)
-		return saArray, err
+		return nil, errors.New(err)
```
ayj: You can return error directly here instead of making a new copy.

kyessenov:
Basically, it seems you are re-computing the pods that belong to a service. This is done by k8s using `Endpoints` resources. The main difference is that pods get admitted to the service at a slightly different stage than acquiring a service account.

platform/kube/controller.go
lookuptable: The `err` defined on Lin 639 is a string instead of an error so we need to use `errors.New()`. I've renamed the var from `err` to `msg` to make it less confusing.

ayj: Actually, it would be better to let the caller handle/print the error here, otherwise we could end up with a double-printed error message in the logs in the future, e.g.

```golang
if !exists {
    return nil, fmt.Errorf("failed to get service for hostname %v", hostname)
}

// Make buildSSLContextWithSAN handle the error

serviceAccounts, err := context.Discovery. GetIstioServiceAccounts(hostname)
if err == nil  {
    return &SSLContextWithSAN{...}
} 
glog.Warningf("Error: %v", err)
```

lookuptable:
@kyessenov In your alternative approach, where do we do pod-to-service membership check?

platform/kube/controller.go
```diff
@@ -628,37 +627,34 @@ const (
...
-	}
+func (c *Controller) GetIstioServiceAccounts(hostname string, ports []string) []string {
 	saSet := make(map[string]bool)
-	for _, p := range pods.Items {
-		sa := makeIstioServiceAccount(p.Spec.ServiceAccountName, namespace, LocalDomain)
-		if _, exists := saSet[sa]; !exists {
-			saSet[sa] = true
-			saArray = append(saArray, sa)
+	for _, si := range c.Instances(hostname, ports, model.TagsList{}) {
+		ip := si.Endpoint.Address
```
lookuptable: call it addr instead of ip

platform/kube/controller.go
```diff
@@ -707,6 +703,7 @@ func newPodCache(ch cacheHandler) *PodCache {
 	}
 
 	ch.handler.append(func(obj interface{}, ev model.Event) error {
+		fmt.Println("event handler is called")
```
lookuptable: This should've been removed

lookuptable:
This fixes #428

lookuptable:
@kyessenov PTAL

platform/kube/controller_test.go
```diff
@@ -466,67 +467,97 @@ func makeService(n, ns string, cl kubernetes.Interface, t *testing.T) {
...
-	if err != nil {
-		t.Error("Error returned: ", err)
-	} else if len(sa) != 0 {
+	sa = controller.GetIstioServiceAccounts(hostname, []string{})
+	if len(sa) != 0 {
 		t.Error("Failure: Expected to resolve 0 service accounts, but got: ", sa)
 	}
+}
+
+func createEndpoints(controller *Controller, name, namespace string, portNames, ips []string, t *testing.T) {
```
kyessenov: nit: I think we should switch to using "real" kubernetes interface instead of the fake one. Creation of the endpoints is really the job of k8s controllers.

PR #443, opened 2017-03-31
Auto-generate proto and mocks with Bazel

kyessenov:


proxy/envoy/policy.go
```diff
@@ -21,8 +21,9 @@ import (
 	"sort"
 	"strings"
 
+	proxyconfig "istio.io/api/proxy/v1/config"
+
```
ayj: Why the extra newline between proxyconfig and the other `istio.io` imports? They are both from the same base import path which is traditionally grouped together.

doc/build.md
```diff
@@ -16,12 +12,17 @@ to update all `BUILD` files:
 
     gazelle -go_prefix "istio.io/manager" -mode fix -repo_root .
 
-[Gazelle tool](https://github.com/bazelbuild/rules_go/tree/master/go/tools/gazelle) to automatically generate `BUILD` files ships with Bazel and is located in Bazel workspace cache:
+Install [Gazelle tool](https://github.com/bazelbuild/rules_go/tree/master/go/tools/gazelle) that automatically generates Bazel `BUILD` files as follows:
+
+    go install github.com/bazelbuild/rules_go/go/tools/gazelle/gazelle
```
ayj: Should this install go into `bin/install-prereqs.sh`?

doc/build.md
```diff
@@ -16,12 +12,17 @@ to update all `BUILD` files:
...
+
+_Note_: Gazelle incorrectly rewrites Istio API dependency. You need to manually replace:
```
ayj: This note doesn't say _where_ this needs to be done. Perhaps we put line 13 and this fix-up into a script to abstract these details away, e.g. `bin/gazelle.sh` ?

proxy/envoy/policy.go
kyessenov: done

doc/build.md
kyessenov: ok, adding a script

doc/build.md
kyessenov: the script bin/gazelle will install it if necessary.

zcahana:
This seems to have broken the non-Bazel builds (e.g., in-IDE compilation):
```sh
zvic@zvic-vbox:~/Workspace/go/src/istio.io/manager (master)$ go build ./cmd/manager/
model/config.go:24:2: no buildable Go source files in /home/zvic/Workspace/go/src/istio.io/api/proxy/v1/config
```

Can we extend `bin/bazel_to_go.py` to link the proto-generated go files into `istio.io/api/proxy/v1/config`?

kyessenov:
I have updated `bin/init.sh` script to link the generated files. Try to run that again.

zcahana:
Thanks, `bin/init.sh` did the trick.

PR #445, opened 2017-04-01
[wip] Initial manager API functionality

liamawhite:

### DIFF
--- a/cmd/manager/api/api.go
+++ b/cmd/manager/api/api.go
@@ -1,0 +1,28 @@
+package api
+
+import (
+	"errors"
+	"net/http"
+
+	"github.com/gorilla/mux"
+)
+
+type API struct {
+	router  *mux.Router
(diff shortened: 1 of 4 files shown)


cmd/manager/api/api.go
```diff
@@ -0,0 +1,28 @@
+package api
+
+import (
+	"errors"
+	"net/http"
+
+	"github.com/gorilla/mux"
```
rshriram: You might want to add this to bazel workspace file in cmd/manager/api/.

liamawhite: That was the plan, just wanted to put some code up in order to get to quick feedback.

ijsnellf:
@liamawhite: There is some existing API code that you may be able to use for reference: https://github.com/istio/manager/blob/master/proxy/envoy/discovery.go

Issue #418, opened 2017-03-30
Full validation needed for route-rules and destination-policies


esnible:
Note that this is related to #406.  That Issue is about rules that have incorrect or no meaning given the current environment.  This issue is about rules that are just plain invalid.

Here are some things I have noticed
Non-ASCII hostnames can be created by istioctl

```
type: route-rule
name: xyz-rule
spec:
  destination: couchdb.€.svc.cluster.local
  http_fault:
    delay:
      percent: 100
      fixed_delay_seconds: 5
```

FQDNs with length > 255 octets can be created by istioctl

(same as above, with destination couchdb.123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890.svc.cluster.local )

HTTP fault / delay with percent > 100 is allowed.

HTTP fault / delay with fixed_delay_seconds < 0 is allowed

ijsnellf:
Yep, there is no validation in place right now.

esnible:
There is a K8s issue for validation, https://github.com/kubernetes/kubernetes/issues/38117 , but it is not moving forward.

A comment from Kuat in the Slack channel

```
There are several stages of validation - some can be done at API level, some need to be done in manager
there's no validation for TPRs in k8s, and AFAIK not coming any time soon
I spoke to one of their engineers from API machinery```

rshriram:
We need to develop this validation as a standalone reusable package under manager. Until the REST API layer lands, we can use this in istioctl. Once we have the REST API layer, we can move the validation to server side.

PR #446, opened 2017-04-03
Initial model validation

esnible:
This item implements fixes for Issue #418 by adding simple static model validation for route-rule and destination-policy for simple things like ranges and validity of FQDN.

In addition, the CLI calls validation before _create_ and _replace_ as an aid to the user.  (The server currently doesn't validate; validation will likely be refactored to the server soon)


cmd/istioctl/main.go
```diff
@@ -99,6 +99,7 @@ var (
 			if len(args) != 0 {
 				return fmt.Errorf("create takes no arguments")
 			}
+			c.SilenceUsage = true
```
kyessenov: Can you explain what's the purpose of this line?

model/validation.go
```diff
@@ -204,16 +204,353 @@ func (t Tags) Validate() error {
 	return errs
 }
 
+func validateFQDN(fqdn string) error {
```
kyessenov: It might be useful to list all errors rather than the first one. You can use `multierror.Append` to aggregate all errors, see the usage in other validation methods.

model/validation.go
```diff
@@ -204,16 +204,353 @@ func (t Tags) Validate() error {
...
+	if len(fqdn) > 255 {
+		return fmt.Errorf("domain name %q too long (max 255)", fqdn)
+	}
+
+	for _, label := range strings.Split(fqdn, ".") {
+		if !IsDNS1123Label(label) {
+			return fmt.Errorf("domain name %q invalid", fqdn)
```
kyessenov: Specific word `label` could help here.

model/validation.go
```diff
@@ -204,16 +204,353 @@ func (t Tags) Validate() error {
...
+		}
+	}
+
+	return nil
```
kyessenov: Is an empty hostname a valid FQDN?

model/validation.go
```diff
@@ -204,16 +204,353 @@ func (t Tags) Validate() error {
...
+func ValidateMatchCondition(mc *proxyconfig.MatchCondition) error {
+	var retVal error
+
+	if mc.Source != "" {
+		if err := validateFQDN(mc.Source); err != nil {
+			retVal = multierror.Append(retVal, err)
+		}
+	}
+
+	// We do not validate source_tags because they have no explicit rules
```
kyessenov: You could try to validate tags actually, there is a function to match tags against regexes that are allowed.

model/validation.go
```diff
@@ -204,16 +204,353 @@ func (t Tags) Validate() error {
...
+	// TODO verify subnet is "IPv4 or IPv6 ip address with optional subnet. E.g., a.b.c.d/xx form or just a.b.c.d
+
+	return retVal
+}
+
+func validateDelay(delay *proxyconfig.HTTPFaultInjection_Delay) error {
+	var retVal error
+
+	if delay.Percent > 100 {
+		retVal = multierror.Append(retVal, fmt.Errorf("delay percent must not exceed 100"))
```
kyessenov: Make a function to validate percent values?

model/validation.go
```diff
@@ -204,16 +204,353 @@ func (t Tags) Validate() error {
...
+
 	value, ok := msg.(*proxyconfig.RouteRule)
 	if !ok {
-		return fmt.Errorf("Cannot cast to routing rule")
+		return fmt.Errorf("cannot cast to routing rule")
 	}
+	var retVal error
 	if value.Destination == "" {
-		return fmt.Errorf("RouteRule must have a destination service")
+		retVal = multierror.Append(retVal, fmt.Errorf("routeRule must have a destination service"))
```
kyessenov: `routeRule` should not be lower cased. Also try avoid the name `RouteRule` since we may change that name at some point, and would be hard to hunt down all usages in strings.

sebastienvas:
jenkins please rebuid istio/manager-pr-master

model/validation.go
```diff
@@ -228,8 +565,31 @@ func ValidateDestinationPolicy(msg proto.Message) error {
 	if !ok {
 		return fmt.Errorf("Cannot cast to destination policy")
 	}
+
+	var retVal error
+
 	if value.Destination == "" {
-		return fmt.Errorf("DestinationPolicy should have a valid service name in its destination field")
+		retVal = multierror.Append(retVal,
+			fmt.Errorf("destinationPolicy should have a valid service name in its destination field"))
```
kyessenov: similar here, call it `destination policy`.

kyessenov:
We need to add unit tests for all error conditions to keep our code coverage.

cmd/istioctl/main.go
esnible: In Cobra, the default is to have Cobra print a command's syntax if it returns an error.

_istioctl_ returns an error if there is a network problem talking to Kubernetes or if the .yaml file is invalid.  By unsetting SilenceUsage Cobra will print the error,  but not the command syntax.

model/validation.go
esnible: I used multierror to make sure the user sees all fields that are invalid.  This particular method validates a DNS name, and it felt like too much data to receive multiple complaints about the same DNS name.

cmd/istioctl/main.go
kyessenov: I wonder if there is a way to achieve the same without code repetition.

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
 	return errs
 }
 
+// Validate ensures tags are well-formed, but does not insist that tags exist
+func Validate(t map[string]string) error {
```
kyessenov: There is a function `(t Tags) Validate()` that does the exact same. What's the purpose of code repetition? You may need to convert types with a function in `conversion.go` but the conversion does not require the map to be valid.

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+
+	return retVal
+}
+
+func validateDelay(delay *proxyconfig.HTTPFaultInjection_Delay) error {
+	var retVal error
+
+	retVal = validateFloatPercent(retVal, delay.Percent, "delay")
+
+	if delay.GetFixedDelaySeconds() < 0 {
```
kyessenov: nit: @rshriram we should switch to duration proto. Then you can just say "1s" in JSON.

model/validation.go
esnible: I originally had the validation call `(t Tags) Validate()` but that validation fails if there are 0 tags.  And 0 tags are allowed!  I wanted the validation of the tags, but not the insistence on >0 tags.  Please advise how you would prefer it.

cmd/istioctl/main.go
esnible: I have seen a Cobra user saying the best way to do this is to set `SilenceUsage: true` when the root Cmd is defined, and explicitly print the usage messages myself.  I didn't want to introduce a lot of changes to code that I did not write.  Setting this explicitly after validation happens seems clean enough.

model/validation.go
kyessenov: Is there any reason why not treat empty tags and 0 tags as same? Tags are only used as lower-bounds, in which case seems to be the same thing. I'm mostly concerned with duplicating code, so feel free to change the original tags.validate() to fit your needs.

cmd/istioctl/main.go
kyessenov: I think this file is now mostly self-contained, so you can change the root command. We can fix later (maybe if we do not use `RunE` it is not a problem).

model/config_test.go
```diff
@@ -846,6 +846,144 @@ func TestValidateRouteAndIngressRule(t *testing.T) {
...
+		{in: &proxyconfig.RouteRule{
+			Destination: "host.default.svc.cluster.local",
+			Match:       &proxyconfig.MatchCondition{Source: "somehost!.default.svc.cluster.local"},
+		},
+			valid: false},
+		{in: &proxyconfig.RouteRule{
+			Destination: "host.default.svc.cluster.local",
+			Match:       &proxyconfig.MatchCondition{Source: "somehost.default.svc.cluster.local"},
+			Route: []*proxyconfig.DestinationWeight{
+				{Destination: "123456789012345678901234567890123456789012345678901234567890." +
```
kyessenov: use `strings.Repeat`

rshriram:
Can you please add a description to the PR? Its nice to have some description on why/what on the PR for various triaging purposes

rshriram:
The PR to master is failing. I think the istioctl list command does not exist anymore

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+		}
+	}
+
+	if mc.GetUdp() != nil {
+		if err := ValidateL4MatchAttributes(mc.GetUdp()); err != nil {
+			retVal = multierror.Append(retVal, err)
+		}
+	}
+
+	// We do not (yet) validate http_headers.
```
rshriram: We can have either TCP or UDP or HTTP , but not all three. This is another instance of semantics. I think the idea behind the validation code was to validate the semantics of proto and not just the syntax. Http is our main target compared to TCP. Its relatively simple to just check for path/prefix and header being an array.

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+func ValidateDestinationWeight(dw *proxyconfig.DestinationWeight) error {
+	var retVal error
+
+	if dw.Destination != "" {
+		if err := validateFQDN(dw.Destination); err != nil {
+			retVal = multierror.Append(retVal, err)
+		}
+	}
+
+	// We do not validate tags because they have no explicit rules
```
rshriram: What does this comment mean with respect to the line below? It is still validating the tags.

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+	if err := Validate(dw.Tags); err != nil {
+		retVal = multierror.Append(retVal, err)
+	}
+
+	retVal = validatePercent(retVal, dw.Weight, "weight")
+
```
rshriram: Don't we have the most important validation in the weight business? sum of all weights should be equal to 100 ?

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+			retVal = multierror.Append(retVal, err)
+		}
+	}
+
+	return retVal
+}
+
+// ValidateL4Fault validates L4 Fault
+func ValidateL4Fault(fault *proxyconfig.L4FaultInjection) error {
+	var retVal error
```
rshriram: This should probably throw a Not implemented Error

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+		}
+	}
+
+	return retVal
+}
+
+func validateSubnet(subnet string) error {
+	var retVal error
+
+	// TODO verify subnet is "IPv4 or IPv6 ip address with optional subnet. E.g., a.b.c.d/xx form or just a.b.c.d
```
rshriram: For the near future, we will only have ipv4 subnet. Lets start with that and get that functionality in place.

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+	var retVal error
+
+	retVal = validateFloatPercent(retVal, delay.Percent, "delay")
+
+	if delay.GetFixedDelaySeconds() < 0 {
+		retVal = multierror.Append(retVal, fmt.Errorf("delay fixed_seconds invalid"))
+	}
+
+	if delay.GetExponentialDelaySeconds() < 0 {
+		retVal = multierror.Append(retVal, fmt.Errorf("delay exponential_seconds invalid"))
```
rshriram: Not implemented

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+
+	return retVal
+}
+
+func validateAbort(abort *proxyconfig.HTTPFaultInjection_Abort) error {
+	var retVal error
+
+	retVal = validateFloatPercent(retVal, abort.Percent, "abort")
+
+	// No validation yet for grpc_status / http2_error / http_status
```
rshriram: There is certainly validation for HTTP_STATUS. >=0 and <600

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+	// No validation yet for override_header_name
+
+	return retVal
+}
+
+func validateTerminate(terminate *proxyconfig.L4FaultInjection_Terminate) error {
+	var retVal error
+
+	retVal = validateFloatPercent(retVal, terminate.Percent, "terminate")
```
rshriram: Not implemented error

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+	}
+
+	if value.HttpFault != nil {
+		if err := ValidateHTTPFault(value.HttpFault); err != nil {
+			retVal = multierror.Append(retVal, err)
+		}
+	}
+
+	if value.L4Fault != nil {
+		if err := ValidateL4Fault(value.L4Fault); err != nil {
```
rshriram: should throw Not implemented error

cmd/istioctl/istioctl_test.go
```diff
@@ -155,3 +155,19 @@ func TestNewGet(t *testing.T) {
...
+	file = "testdata/invalid-route-rule.yaml"
+	if err := postCmd.RunE(postCmd, []string{}); err == nil {
+		t.Fatalf("Did not fail when presented with invalid route rule")
+	}
+}
+
+func TestInvalidDestinationPolicy(t *testing.T) {
+	rootSetup(t)
+	file = "testdata/invalid-destination-policy2.yaml"
+	if err := postCmd.RunE(postCmd, []string{}); err == nil {
```
esnible: I'd like to keep them.  Although I wrote them hoping that a call to validation from the CLI would improve my code coverage the tests validate that the CLI is, in fact, doing validation.  They have already found a bug in a test.  I request that they stay.

cmd/istioctl/main.go
esnible: SilenceUsage is not an _istioctl_ flag.  It is an internal option to Cobra that effects how Cobra prints error messages.  However at Kuat's request I have changed the logic and the new logic is much more clear.

model/validation.go
esnible: Two things.  Most importantly, this code currently only has access to the current POST or PUT and does not have the ability to globally sum all of the weights.  There might be another weight already there that brings this into compliance.

Second I have asked you in personal and via Slack why the sum of weights needs to be equal to 100 and what it means if it isn't and you have not yet answered.  Suppose I have three services and want to divide traffic equally.  Why must I have weights 33, 33, and 34 instead of 33, 33, and 33?  What happens if I delete a rule and then post a new rule and for a brief time the weight is only 66?  Please explain explain how these cases are intended to work.

cmd/istioctl/istioctl_test.go
rshriram: Then please rename the other test accordingly or the test data file accordingly.

model/validation.go
rshriram: I don't think we have patch semantics. Post creates a rule and put update it. And in both cases, we expect that the user provide weights for all versions, not just a particular version. The question about 33,33,33 - that's the idea behind the validation you are doing. To prevent such a scenario. And you cannot have a rule where you update weight of a particular version. It's update all or none.

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+	}
+
+	return retVal
+}
+
+func validatePercent(err error, val int32, label string) error {
+	if val > 100 {
+		err = multierror.Append(err, fmt.Errorf("%v must not exceed 100", label))
+	}
+	if val < 0 {
```
ijsnellf: Simplify the validatePercent logic to a single check?
```
if val < 0 || val > 100 {
    return fmt.Errorf("%v must be in range 0..100", label)
}
```

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+			return fmt.Errorf("domain name %q invalid (label %q invalid)", fqdn, label)
+		}
+	}
+
+	return nil
+}
+
+// ValidateMatchCondition validates a Match Condition
+func ValidateMatchCondition(mc *proxyconfig.MatchCondition) error {
+	var retVal error
```
ijsnellf: Use named return variables?: https://tour.golang.org/basics/7

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+		err = multierror.Append(err, fmt.Errorf("%v must not exceed 100", label))
+	}
+	if val < 0 {
+		err = multierror.Append(err, fmt.Errorf("%v must be in range 0..100", label))
+	}
+	return err
+}
+
+func validateFloatPercent(err error, val float32, label string) error {
+	if val > 100.0 {
```
ijsnellf: Same comment as above

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+	return retVal
+}
+
+// ValidateHTTPTimeout validates HTTP Timeout
+func ValidateHTTPTimeout(timeout *proxyconfig.HTTPTimeout) error {
+	var retVal error
+
+	if simple := timeout.GetSimpleTimeout(); simple != nil {
+		if simple.TimeoutSeconds < 0 {
+			retVal = multierror.Append(retVal, fmt.Errorf("timeout_seconds must be in range [0..]"))
```
ijsnellf: nit: just `return fmt.Errorf("timeout_seconds must be in range [0..]")` since there is only one error here?

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+		if err := validateFQDN(mc.Source); err != nil {
+			retVal = multierror.Append(retVal, err)
+		}
+	}
+
+	if err := Validate(mc.SourceTags); err != nil {
+		retVal = multierror.Append(retVal, err)
+	}
+
+	if mc.GetTcp() != nil {
```
ijsnellf: Maybe we should put the `!= nil` checks inside the validate functions? 
```
func validateX(x *X) error {
    if x == nil {
        return nil
    }
    ...
}
```

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+	// We do not (yet) validate http_headers.
+
+	return retVal
+}
+
+// ValidateL4MatchAttributes validates L4 Match Attributes
+func ValidateL4MatchAttributes(ma *proxyconfig.L4MatchAttributes) error {
+	var retVal error
+
+	if ma.SourceSubnet != nil {
```
ijsnellf: nit: `if ma.SourceSubset != nil` isn't necessary because `range` considers a nil array to be empty.

model/validation.go
```diff
@@ -204,16 +204,367 @@ func (t Tags) Validate() error {
...
+	return retVal
+}
+
+// ValidateCircuitBreaker validates Circuit Breaker
+func ValidateCircuitBreaker(cb *proxyconfig.CircuitBreaker) error {
+	var retVal error
+
+	if simple := cb.GetSimpleCb(); simple != nil {
+		if simple.MaxConnections < 0 {
+			retVal = multierror.Append(retVal, fmt.Errorf("circuit_breaker max_connections must be in range [0..]"))
```
ijsnellf: `circuit_breaker max_connections must be non-negative`?

model/validation.go
```diff
@@ -228,8 +579,33 @@ func ValidateDestinationPolicy(msg proto.Message) error {
 	if !ok {
 		return fmt.Errorf("Cannot cast to destination policy")
 	}
+
+	var retVal error
```
ijsnellf: nit: `retVal` is non-descriptive. I think the convention for `multierror` is generally `errs`.

model/config_test.go
```diff
@@ -840,19 +841,173 @@ func TestTagsEquals(t *testing.T) {
...
+			HttpFault: &proxyconfig.HTTPFaultInjection{
+				Abort: &proxyconfig.HTTPFaultInjection_Abort{
+					Percent:   100,
+					ErrorType: &proxyconfig.HTTPFaultInjection_Abort_HttpStatus{HttpStatus: -1},
+				},
+			},
+		},
+			valid: false},
+		{name: "route rule bad delay exp seconds", in: &proxyconfig.RouteRule{
+			Destination: "host.default.svc.cluster.local",
```
rshriram: We do not support these in our protocol. These should return an error as I noted earlier.

model/validation.go
```diff
@@ -204,16 +202,453 @@ func (t Tags) Validate() error {
...
+			retVal = multierror.Append(retVal, err)
+		}
+	}
+
+	if err := Tags(mc.SourceTags).Validate(); err != nil {
+		retVal = multierror.Append(retVal, err)
+	}
+
+	if mc.GetTcp() != nil {
+		if err := ValidateL4MatchAttributes(mc.GetTcp()); err != nil {
```
rshriram: NOT IMPLEMENTED error

model/validation.go
```diff
@@ -204,16 +202,453 @@ func (t Tags) Validate() error {
...
+}
+
+func validateAbort(abort *proxyconfig.HTTPFaultInjection_Abort) error {
+	var retVal error
+
+	retVal = validateFloatPercent(retVal, abort.Percent, "abort")
+
+	switch abort.ErrorType.(type) {
+	case *proxyconfig.HTTPFaultInjection_Abort_GrpcStatus:
+		// No validation yet for grpc_status / http2_error / http_status
```
rshriram: NOT IMPLEMENTED error

model/validation.go
```diff
@@ -204,16 +202,453 @@ func (t Tags) Validate() error {
...
+	var retVal error
+
+	retVal = validateFloatPercent(retVal, terminate.Percent, "terminate")
+
+	return retVal
+}
+
+func validateThrottle(throttle *proxyconfig.L4FaultInjection_Throttle) error {
+	var retVal error
+
```
rshriram: NOT IMPLEMENTED error

model/validation.go
```diff
@@ -204,16 +202,453 @@ func (t Tags) Validate() error {
...
+			retVal = multierror.Append(retVal,
+				fmt.Errorf("circuit_breaker http_max_requests_per_connection must be in range [0..]"))
+		}
+		retVal = validatePercent(retVal, simple.HttpMaxEjectionPercent, "circuit_breaker http_max_ejection_percent")
+	}
+
+	return retVal
+}
+
+func validateWeighs(routes []*proxyconfig.DestinationWeight, defaultDestination string) error {
```
rshriram: typo. validateWeighs -> validateWeights (missing t)

model/validation.go
ijsnellf: Route rules are updated atomically. @rshriram is referring to ensuring that the destination weights inside the route rule sum to 100. See: https://github.com/istio/api/blob/master/proxy/v1/config/cfg.proto#L170

cmd/istioctl/istioctl_test.go
esnible: I do not understand this comment.  We have a file _istio/manager/cmd/istioctl/testdata/invalid-dest-policy.yaml_ and a file _istio/manager/cmd/istioctl/testdata/invalid-dest-policy2.yaml_.  If I rename the "2" file, it will collide with the "1" file.

model/validation.go
esnible: I had misunderstood the spec, because I believed it was possible for two route-rules to refer to the same Destination hostname.  I have revised the code; see https://github.com/istio/manager/pull/446/commits/26008741935d8c73b848003a985130fd764c90bc#diff-3296bce329ea4b3db021a65e93b42ada

model/validation.go
esnible: I felt the L4Fault itself was valid.  The flaw is that it was put into a route-rule, and it is not implemented there.  The caller throws it:
```
	if value.L4Fault != nil {
		if err := ValidateL4Fault(value.L4Fault); err != nil {
			retVal = multierror.Append(retVal, err)
		}
		retVal = multierror.Append(retVal, fmt.Errorf("L4 faults are not implemented"))
	}
```
In this way it would be possible to write tests for the validity of naked L4 faults but still get the desired behavior for the CLI and the overall model validation.

model/validation.go
esnible: Same comment as before, I put validation for unimplemented features in in the container, not in the thing itself.

cmd/istioctl/istioctl_test.go
rshriram: Please sort out some sane naming for the files. And please keep it consistent as well. Whether it's *policy1.yaml, *policy2.yaml, etc or something else. As long as the names follow a consistent pattern, it's fine

model/validation.go
rshriram: We don't even provide l4 fault support. So telling the user that she had provided an incorrect specification of l4 fault followed by a message stating that it's not implemented seems counter intuitive. We should simply throw a not implemented error for clarity.

model/config_test.go
esnible: You will see the `valid: false` a few lines down.  The test validates that the rule is invalid, as you expected.  Actually two errors are returned via multierror -- one error for the bad form, and an additional one for the fact that the feature is unimplemented.

model/validation.go
esnible: Sam comment as before.  The Throttle is correct, but we don't support it, so I have put the check on the caller.

model/config_test.go
```diff
@@ -840,19 +841,191 @@ func TestTagsEquals(t *testing.T) {
...
+		name  string
 		in    proto.Message
 		valid bool
 	}{
-		{in: &proxyconfig.DestinationPolicy{}, valid: false},
-		{in: &proxyconfig.RouteRule{}, valid: false},
-		{in: &proxyconfig.RouteRule{Destination: "foobar"}, valid: true},
+		{name: "empty destination policy", in: &proxyconfig.DestinationPolicy{}, valid: false},
+		{name: "empty route rule", in: &proxyconfig.RouteRule{}, valid: false},
+		{name: "route rule w destination", in: &proxyconfig.RouteRule{Destination: "foobar"}, valid: true},
```
ijsnellf: typo?

model/config_test.go
```diff
@@ -840,19 +841,191 @@ func TestTagsEquals(t *testing.T) {
...
+		{name: "route rule bad destination", in: &proxyconfig.RouteRule{
+			Destination: "badhost@.default.svc.cluster.local",
+		},
+			valid: false},
+		{name: "route rule bad match source", in: &proxyconfig.RouteRule{
+			Destination: "host.default.svc.cluster.local",
+			Match:       &proxyconfig.MatchCondition{Source: "somehost!.default.svc.cluster.local"},
+		},
+			valid: false},
+		{name: "route rule bad weight dest", in: &proxyconfig.RouteRule{
```
ijsnellf: nit: destination?

model/config_test.go
```diff
@@ -840,19 +841,191 @@ func TestTagsEquals(t *testing.T) {
...
+			Destination: "host.default.svc.cluster.local",
+			HttpFault: &proxyconfig.HTTPFaultInjection{
+				Abort: &proxyconfig.HTTPFaultInjection_Abort{
+					Percent:   100,
+					ErrorType: &proxyconfig.HTTPFaultInjection_Abort_HttpStatus{HttpStatus: -1},
+				},
+			},
+		},
+			valid: false},
+		{name: "route rule bad delay exp seconds", in: &proxyconfig.RouteRule{
```
ijsnellf: nit: expiration

model/validation.go
```diff
@@ -204,16 +202,433 @@ func (t Tags) Validate() error {
...
+	errs = validateFloatPercent(errs, abort.Percent, "abort")
+
+	switch abort.ErrorType.(type) {
+	case *proxyconfig.HTTPFaultInjection_Abort_GrpcStatus:
+		// TODO No validation yet for grpc_status / http2_error / http_status
+		errs = multierror.Append(errs, fmt.Errorf("Istio does not support gRPC fault injection yet"))
+	case *proxyconfig.HTTPFaultInjection_Abort_Http2Error:
+		// TODO No validation yet for grpc_status / http2_error / http_status
+	case *proxyconfig.HTTPFaultInjection_Abort_HttpStatus:
+		if err := validateAbortHTTPStatus(abort.ErrorType.(*proxyconfig.HTTPFaultInjection_Abort_HttpStatus)); err != nil {
```
ijsnellf: nit: long line

model/validation.go
```diff
@@ -204,16 +202,433 @@ func (t Tags) Validate() error {
...
+			errs = multierror.Append(errs, err)
+		}
+	}
+
+	// No validation yet for override_header_name
```
ijsnellf: `// TODO: validate override_header_name`?

model/validation.go
```diff
@@ -204,16 +202,433 @@ func (t Tags) Validate() error {
...
+
+	return
+}
+
+func validateTerminate(terminate *proxyconfig.L4FaultInjection_Terminate) (errs error) {
+
+	errs = validateFloatPercent(errs, terminate.Percent, "terminate")
```
ijsnellf: nit: could be a single line: `return validateFloatPercent(errs, terminate.Percent, "terminate")`

cmd/istioctl/main.go
```diff
@@ -77,7 +77,7 @@ var (
 				}
 			}
 
-			client, err := kube.NewClient(kubeconfig, model.IstioConfig)
+			client, err = kube.NewClient(kubeconfig, model.IstioConfig)
```
kyessenov: thanks, surprised linters did not catch this obvious mistake.

cmd/istioctl/main.go
```diff
@@ -202,13 +213,20 @@ var (
 		},
 	}
 
+	listCmd = &cobra.Command{
```
kyessenov: Let's remove this, the demo will be updated after merging this.

model/validation.go
```diff
@@ -204,16 +202,433 @@ func (t Tags) Validate() error {
...
+		destination := destWeight.Destination
+		if destination == "" {
+			destination = defaultDestination
+		}
+
+		oldWeight := weights[destination]
+		weights[destination] = oldWeight + destWeight.Weight
+	}
+
+	// From cfg.proto "If there is only [one] destination in a rule, the weight value is assumed to be 100.
```
kyessenov: Why do we sum by destination? Isn't the total always 100 no matter where destination goes? Also, please check the case a single weight that is not 100.

PR #449, opened 2017-04-04
Utilize global mesh config ConfigMap

kyessenov:
Reorganizing configuration to use a shared config file instead of providing flags to each container.
Included in the change:
1) disable mixer by default
2) change drain and parent shutdown to 2s and 3s.
3) fetch newer API proto definitions
4) remove stale `go generate` calls (all files are generated by bazel prior to go tooling)
5) wire `ConnectTimeout` through config generation
6) apply SSL contexts in a separate pass rather than passing it around

Some remaining issues:
1) kube-inject should query values from the config map
2) init container should read port from the config map
3) istioctl should provide validation for the config map
4) config map is fetched from the namespace parameter; that means "all" namespaces will break it, and we might need another flag for the config namespace


proxy/envoy/config.go
```diff
@@ -278,6 +282,21 @@ func buildOutboundHTTPRoutes(instances []*model.ServiceInstance, services []*mod
...
 			}
 		}
+
+		clusters.setTimeout(context.MeshConfig.ConnectTimeout)
+
+		// apply SSL context to outbound clusters for authentication policy
+		switch context.MeshConfig.AuthPolicy {
+		case proxyconfig.ProxyMeshConfig_NONE:
+		case proxyconfig.ProxyMeshConfig_MUTUAL_TLS:
+			serviceAccounts, _ := context.Discovery.GetIstioServiceAccounts(service.Hostname)
```
wattli: Do we need to throw some exception or add log something when it fails to get istio service ?

kyessenov: Dunno, I just moved the code around.

kyessenov: @myidpt

kyessenov:
#251

proxy/envoy/config.go
lookuptable: Currently `GetIstioServiceAccount()` logs the error and returns it, though I'm not in favor of this pattern. I have an open PR #440 that changes the behavior to only logging the error.

proxy/envoy/route.go
```diff
@@ -37,22 +39,20 @@ const (
 )
 
 // buildListenerSSLContext returns an SSLContext struct.
-func buildListenerSSLContext(mesh *MeshConfig) *SSLContext {
+func buildListenerSSLContext(certs string) *SSLContext {
```
lookuptable: Is `certDir` a better name?

cmd/manager/main.go
```diff
@@ -257,5 +233,31 @@ func setFlagsFromEnv() {
...
+	}
+
+	// values in the data are strings, while proto might use a different data type.
+	// therefore, we have to get a value by a key
+	yaml, exists := config.Data[ConfigMapKey]
+	if !exists {
+		return nil, fmt.Errorf("missing configuration map key %q", ConfigMapKey)
+	}
+
+	mesh := *envoy.DefaultMeshConfig
```
lookuptable: Do you need de-referencing here?

proxy/envoy/config.go
kyessenov: Agree, hiding the error is not a good idea. Logging seems fine.

proxy/envoy/route.go
kyessenov: Sure, can rename it.

cmd/manager/main.go
kyessenov: Yes, this is important, otherwise, we'll be overriding default values. De-referencing to a stack variable does a  shallow copy of the pointer.

proxy/envoy/resources.go
```diff
@@ -18,96 +18,67 @@ import (
...
+	// HeaderURI is URI HTTP header
 	HeaderURI = "uri"
 
 	// WildcardAddress binds to all IP addresses
 	WildcardAddress = "0.0.0.0"
 )
 
+var (
+	// DefaultMeshConfig configuration
+	DefaultMeshConfig = &proxyconfig.ProxyMeshConfig{
```
ayj: Any reason this can't be a value instead of a pointer? 

Also, it's unfortunate this can't be a `const` to avoid allowing users to accidentally modify it, but I'm guessing the `ptypes.DurationProto` below prevents that? Alternatively, you could replace this with a function that returns a copy of the default config, e.g.
```golang
func DefaultMeshConfig() proxyconfig.proxyMeshConfig{
    return proxyconfig.proxyMeshConfig{ 
        // ... 
    }
}
```

proxy/envoy/resources.go
```diff
@@ -18,96 +18,67 @@ import (
...
+		AuthPolicy:    proxyconfig.ProxyMeshConfig_NONE,
+		AuthCertsPath: "/etc/certs",
+	}
+)
+
+// convertDuration converts to golang duration and logs errors
+func convertDuration(d *duration.Duration) time.Duration {
+	dur, err := ptypes.Duration(d)
+	if err != nil {
+		glog.Warningf("error converting duration %#v: %v", d, err)
```
ayj: How does envoy behave if we specific zero value for timeouts/delays? Since ptypes.Duration() isn't considered an error here, should we fall back to default timeout values? That might require requiring the caller invoke ptypes.Duration() *or* adding a default value parameter to convertDuration() which is used if the duration cannot be converted.

cmd/istioctl/main.go
```diff
@@ -251,7 +250,7 @@ var (
 func init() {
 	rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "c", "",
 		"Use a Kubernetes configuration file instead of in-cluster configuration")
-	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", "",
+	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", "default",
```

model/conversion.go
```diff
@@ -24,6 +24,16 @@ import (
 	"github.com/golang/protobuf/proto"
 )
 
+// Make creates a new instance of the proto message
+func (ps *ProtoSchema) Make() (proto.Message, error) {
+	pbt := proto.MessageType(ps.MessageName)
+	if pbt == nil {
+		return nil, fmt.Errorf("unknown type %q", ps.MessageName)
+	}
+	pb := reflect.New(pbt.Elem()).Interface().(proto.Message)
```
kyessenov: done

proxy/envoy/resources.go
kyessenov: Yes, it is unfortunate that we cannot make a struct immutable. I changed to a value since it makes more sense for a const that can never be `nil`.

proxy/envoy/resources.go
kyessenov: I suggest we validate all proto durations to be valid golang durations prior to consuming the value. Hardcoding a default duration might cause some surprising behaviors I think. We should just outright reject durations >10000 years :) In that case, this warning should never happen.

proxy/envoy/resources.go
```diff
@@ -491,6 +466,12 @@ func (s Clusters) normalize() Clusters {
 	return out
 }
 
+func (clusters Clusters) setTimeout(timeout *duration.Duration) {
+	for _, cluster := range clusters {
+		cluster.ConnectTimeoutMs = int(convertDuration(timeout) / time.Millisecond)
```
kyessenov: thanks, done

proxy/envoy/resources.go
ayj: Early and completion validation sounds good. In that case, we should probably treat the warning here as an error/fatal/panic since it's clearly a programmer error at that point and otherwise envoy might be incorrectly configured?

kyessenov: Ok, I will add a fallback duration argument after all. Hard error is not a good idea in a sidecar before we have health checks.

test/util/diff.go
```diff
@@ -0,0 +1,104 @@
...
+	"strings"
+	"testing"
+
+	"github.com/pmezard/go-difflib/difflib"
+)
+
+// Refresh controls whether to update the golden artifacts instead.
+// It is set using the environment variable REFRESH_GOLDEN.
+func Refresh() bool {
+	v, exists := os.LookupEnv("REFRESH_GOLDEN")
```
kyessenov: PTAL, I changed the golden refresh procedure, since this PR affected many golden artifacts.

ayj: Does this work well with bazel or do we need to use `go test` to regenerate artifacts?

kyessenov: It's possible but I don't it's worth explaining how to do it with bazel.  On the other hand `env REFRESH_GOLDEN=true go test ./...` just works.

PR #450, opened 2017-04-04
Inject Istio key/certs using istioctl

lookuptable:


lookuptable:
@rshriram It seems all test YAML files in `platform/kube/inject/testdata` are using gcr.io currently (These are for testing kube-inject and are NOT actually deployed). I think we should switch them all away from gcr.io but I'd prefer doing it in a separate PR.

Meanwhile @myidpt is working on implementing integration tests.

kyessenov:
Would it better to make proxy agent fetch files directly instead of mounting them in the pod spec?

platform/kube/inject/inject.go
```diff
@@ -50,6 +50,7 @@ const (
 	istioSidecarAnnotationSidecarKey   = "alpha.istio.io/sidecar"
 	istioSidecarAnnotationSidecarValue = "injected"
 	istioSidecarAnnotationVersionKey   = "alpha.istio.io/version"
+	istioCertVolumeName                = "istio-cert"
```
kyessenov: This needs to be a flag.

platform/kube/inject/inject.go
```diff
@@ -137,46 +140,71 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
...
+		args = append(args, "--enable_auth", "--auth_config_path", p.AuthConfigPath)
+		volumeMounts = append(volumeMounts, v1.VolumeMount{
+			Name:      istioCertVolumeName,
+			ReadOnly:  true,
+			MountPath: p.AuthConfigPath,
+		})
+
+		sa := t.Spec.ServiceAccountName
+		if sa == "" {
+			sa = "default"
```
kyessenov: is this a convention to call it default or configurable by a cluster admin?

kyessenov: I'm somewhat concerned that you are making an assumption about the name of the service account. This script can be run before AC injects the service account. What happens with deployments? Do pods share the same service account?

platform/kube/inject/inject.go
```diff
@@ -137,46 +140,71 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
...
+		}
+		t.Spec.Volumes = append(t.Spec.Volumes, v1.Volume{
+			Name: istioCertVolumeName,
+			VolumeSource: v1.VolumeSource{
+				Secret: &v1.SecretVolumeSource{
+					SecretName: "istio." + sa,
```
kyessenov: Oh I see, the name of the secret is "istio.default". I think this needs to be configurable and agreed with Istio CA.

platform/kube/inject/inject.go
lookuptable: I feel this is internal and doesn't need to be configurable. As long as the name of Volume and that of VolumeMount match we should be fine.

platform/kube/inject/inject.go
lookuptable: I believe K8S uses the default service account unless it's otherwise specified in the pod spec.

platform/kube/inject/inject.go
lookuptable: Istio CA uses a fixed naming scheme for the secret. It's `istio.${serviceAccountName}`

lookuptable:
@kyessenov mounting secrets as a volume is what we have designed in the design doc. Changing it to directly pull certs from api means we need another design to handle certificate rotations.

platform/kube/inject/inject.go
kyessenov: I thought each pod has a unique service account.

platform/kube/inject/inject.go
kyessenov: This needs to pulled out to a constant. Magic strings are hard to find.

kyessenov:
Is there work on rotating certs within Envoy? Use a volume just shifts the problem to Envoy.

platform/kube/inject/inject.go
lookuptable: Multiple pods can share the same service account.

platform/kube/inject/inject.go
lookuptable: Done.

PR #451, opened 2017-04-04
Read TLS secrets for ingress from ingress resource.

ijsnellf:
Currently, only wildcard secret is supported.


proxy/envoy/ingress.go
```diff
@@ -161,17 +159,19 @@ func generateIngress(conf *IngressConfig) *Config {
 		},
 	}
 
-	// configure for HTTPS if provided with a secret name
-	if conf.Secret != "" {
-		// configure Envoy
+	// configure for HTTPS if there is a secret
+	if secret, err := conf.Secrets.GetSecret("*"); err != nil {
```
kyessenov: We still need to pass the name of the secret `*` somehow...

rshriram: Basically the secret name is the hostname. So, isn't * just a wildcard domain ?

ijsnellf: The name of the secret is hidden by the Secret interface. Instead of addressing secrets by secret name, they are addressed by host.

proxy/envoy/ingress.go
```diff
@@ -161,17 +159,19 @@ func generateIngress(conf *IngressConfig) *Config {
...
+		glog.Warningf("Error retrieving ingress secret: %v", err)
+	} else if secret != nil {
+		// configure Envoy for HTTPS
 		listener.Address = fmt.Sprintf("tcp://%s:443", WildcardAddress)
```
zcahana: The current approach uses a single listener, either on port 80 **OR** port 443. In practice, there could be several ingress resources where one specifies TLS config, whereas the other doesn't (even for the same vhost). We should be able to distinguish between the two cases, and assign each derived envoy route to either port 80 or 443.

Currently there's no structured way to capture this in the model, but maybe we can add a `RouteRule.Match.Protocol` field for this.

platform/kube/controller.go
```diff
@@ -251,6 +252,21 @@ func (c *Controller) appendIngressConfigHandler(k string, f func(model.Key, prot
...
+
+			if len(tls.Hosts) == 0 {
+				c.config.Secrets.put("*", tls.SecretName)
+			} else {
+				for _, host := range tls.Hosts {
+					c.config.Secrets.put(host, tls.SecretName)
+				}
+			}
+		}
+
```
ijsnellf: Yes. I'll document it.

proxy/envoy/ingress.go
ijsnellf: I was intending to extend it that way in a follow-on PR, but I can add it to this PR instead.

platform/kube/controller.go
```diff
@@ -251,6 +252,21 @@ func (c *Controller) appendIngressConfigHandler(k string, f func(model.Key, prot
 			return nil
 		}
 
+		// TODO: priority?
+		for _, tls := range ingress.Spec.TLS {
```
ijsnellf: Done.

platform/kube/secret.go
```diff
@@ -0,0 +1,43 @@
...
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	s.secrets[k] = v
+}
+
+func (s *Secret) GetSecret(host string) (map[string][]byte, error) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	name := s.secrets[host]
+	if name == "" {
```
ijsnellf: Done.

proxy/envoy/ingress.go
rshriram: We already have it. RouteRule.Match. L4MatchAttributes

platform/kube/secret.go
```diff
@@ -0,0 +1,44 @@
+package kube
+
+import (
+	"sync"
+
+	"k8s.io/client-go/kubernetes"
+	metav1 "k8s.io/client-go/pkg/apis/meta/v1"
+)
+
+func NewSecret(namespace string, client kubernetes.Interface) *Secret {
```
kyessenov: `NewSecretStorage`?

platform/kube/controller.go
kyessenov: Yes, there are two event notifications that need to be handled: 1) updating the secret storage map (user changing the ingress resource may change the host secret 2) restarting the proxy if the content of the secret changes

proxy/envoy/ingress.go
zcahana: I'm looking for a way to distinguish between HTTP vs HTTPS route rules. `L4MatchAttributes` doesn't seem to provide that:
```protobuf
/// L4 connection match attributes. Note that L4 connection matching
/// support is incomplete.
message L4MatchAttributes {
  /// IPv4 or IPv6 ip address with optional subnet. E.g., a.b.c.d/xx form or
  /// just a.b.c.d
  repeated string source_subnet = 1;

  /// IPv4 or IPv6 ip address of destination with optional subnet.
  /// E.g., a.b.c.d/xx form or just a.b.c.d. This is only valid when the destination
  /// service has several IPs and the application explicitly specifies a particular IP.
  repeated string destination_subnet = 2;
}
```

proxy/envoy/ingress.go
```diff
@@ -111,8 +130,38 @@ func generateIngress(conf *IngressConfig) *Config {
...
+	tlsHosts := make(map[string]bool)
+	for host := range rulesByHost {
+		if t, err := conf.Secrets.GetSecret(host); err != nil {
+			glog.Warningf("Error retrieving TLS context for %q: %v", host, err)
+			tlsHosts[host] = true
+		} else if t != nil {
+			if tls == nil {
+				tls = t
+			} else if string(tls.PrivateKey) != string(t.PrivateKey) || string(tls.Certificate) != string(t.Certificate) {
+				glog.Warningf("Invalid ingress configuration for %q: multiple TLS configs", host)
```
zcahana: nit: this is not an invalid configuration, but rather an unsupported one due to lack of SNI in Envoy.

proxy/envoy/ingress.go
```diff
@@ -111,8 +130,38 @@ func generateIngress(conf *IngressConfig) *Config {
...
+		glog.Warningf("Error retrieving TLS context for wildcard: %v", err)
+		tlsValid = false
+	}
+	wildcardTLS := tls != nil
+
+	tlsHosts := make(map[string]bool)
+	for host := range rulesByHost {
+		if t, err := conf.Secrets.GetSecret(host); err != nil {
+			glog.Warningf("Error retrieving TLS context for %q: %v", host, err)
+			tlsHosts[host] = true
```
zcahana: `tlsHosts[host] = true` is probably by mistake?

model/secret.go
```diff
@@ -20,5 +20,10 @@ package model
 // TODO: add controller hooks for notification about changes to the secret
 type SecretRegistry interface {
 	// GetSecret retrieves the secret data by a platform specific URI
-	GetSecret(uri string) (map[string][]byte, error)
+	GetSecret(uri string) (*TLSContext, error)
```
zcahana: It would also be good to rename this method to `GetCertificate(host string)`. This clearly captures what this method does and on what parameters it operates on, compared to `GetSecret(uri)`.

platform/kube/secretstore.go
```diff
@@ -0,0 +1,71 @@
...
+	name := s.secrets[host]
+	if name == "" {
+		name = s.secrets["*"] // try falling back to wildcard
+	}
+	if name == "" {
+		return nil, nil // no secret name for this host
+	}
+
+	// retrieve the secret
+	secret, err := s.client.Core().Secrets(s.namespace).Get(name, metav1.GetOptions{})
```
zcahana: The namespace here is still incorrect - it should be the namespace in which the referring ingress resource has been defined.

I would change `put(host, secretName string)` to `putTLSSecret(host, namespace, secretName string)`, store that in a _host --> namespace+secretName_ map, and use the correct namespace in `s.client.Core().Secrets(namespace).Get(name, metav1.GetOptions{})`.

platform/kube/controller.go
zcahana: Raised #486 to capture this.

proxy/envoy/ingress.go
zcahana: Raised #485 to capture this.

proxy/envoy/ingress.go
zcahana: Raised #484 for implementing SNI support (tracking Envoy SNI support).

proxy/envoy/ingress.go
ijsnellf: Fixed.

ijsnellf:
@zcahana, @kyessenov: I'll be adding more code coverage, but otherwise I believe I have addressed the outstanding code review comments.

platform/kube/secretstore.go
ijsnellf: Done

model/secret.go
ijsnellf: Done

proxy/envoy/ingress.go
ijsnellf: I'm missing `tlsValid = false`. If we can't determine whether something is TLS, we omit it from the vhosts that we generate.

platform/kube/secret.go
```diff
@@ -5,40 +5,58 @@ import (
...
+	}
+
+	// retrieve the secret
 	secret, err := s.client.Core().Secrets(s.namespace).Get(name, metav1.GetOptions{})
 	if err != nil {
 		return nil, err
 	}
-	return secret.Data, nil
+
+	return &model.TLSContext{
```
ijsnellf: Done

platform/kube/controller.go
```diff
@@ -248,6 +250,8 @@ func (c *Controller) appendIngressConfigHandler(k string, f func(model.Key, prot
 			return nil
 		}
 
+		c.mapIngressSecrets(ingress.Spec.TLS)
```
kyessenov: We should be deleting secrets on delete event for garbage collection.

model/secret.go
```diff
@@ -19,6 +19,12 @@ package model
 // the data immediately to the client of this interface.
 // TODO: add controller hooks for notification about changes to the secret
 type SecretRegistry interface {
-	// GetSecret retrieves the secret data by a platform specific URI
-	GetSecret(uri string) (map[string][]byte, error)
+	// GetTLSSecret retrieves a TLS secret by namespace and host
+	GetTLSSecret(namespace, host string) (*TLSSecret, error)
```
kyessenov: Why do we need namespace here?

ijsnellf: As opposed to encoding it into the URI instead?

platform/kube/controller.go
ijsnellf: Fixed.

model/secret.go
rshriram: I agree with @kyessenov . Namespace is a very k8s specific thing. A generic URI provides more flexibility in how platform specific nuances can be encoded in the URI.

proxy/envoy/testdata/ingress-route-2.yaml.golden
```diff
@@ -0,0 +1,13 @@
+destination: foo.default.svc.cluster.local
+match:
+  http_headers:
```
kyessenov: what's the difference between the two? can you rename from x-1 and x-2 to something more meaningful?

platform/kube/secretstore.go
```diff
@@ -0,0 +1,110 @@
...
+
+	var namespace string
+	if uri == "*" {
+		namespace = s.wildcardNamespace
+	} else {
+		parts := strings.Split(uri, ".")
+		if len(parts) < 2 {
+			return nil, fmt.Errorf("%q is not in the expected URI format: name.namespace.svc.cluster.local", uri)
+		}
+		namespace = parts[1]
```
kyessenov: This is not right. Please read an actual example of an ingress here: https://github.com/kubernetes/contrib/blob/master/ingress/controllers/nginx/examples/multi-tls/multi-tls.yaml#L73. The host in TLS spec is not internal service hostname. I'd expect your test to test actual ingress rather than our internal representation.

proxy/envoy/testdata/ingress-route-2.yaml.golden
rshriram: I think this PR has already had 50+ reviews with most of them being super pointless discussions, while PRs with bugs (or those that don't pass tests) have been happily force merged. So let's shelve these cosmetic nits and try to get the feature working first.

platform/kube/secretstore.go
rshriram: @ijsnellf Kuat is right here. Please comment out all the stuff related to host-based TLS in the function. Just match on * and return the secret. We can deal with single host tls post alpha.

proxy/envoy/testdata/ingress-route-2.yaml.golden
kyessenov: this is a nit. feel free to disregard.

ijsnellf: They are identical except what they point to. I can change them into `ingress-route-world.yaml` and `ingress-route-foo.yaml`, I guess.

Issue #375, opened 2017-03-22
upgrade client-go for k8s 1.6


kyessenov:
1.6 is getting released soon and we need to upgrade WORKSPACE and client-go for bug fixes.

PR #456, opened 2017-04-05
Upgrade client-go dependencies

kyessenov:
Use latest client library (compatible with 1.6). 
No functional changes except for:
- increase resync period to 1s since that is the minimum allowed.

#375


kyessenov:
@zcahana, yes that was intentional. that repo is mostly maintained by a bot (via sync to main kube repo and some code generation I think.) Since they recommend v2, it's same risk to use v3beta or HEAD. The big change there was the split into apimachinery repo that is tightly coupled with client-go, so from now on it should be easier to sync up once v3 is officially released.

ayj:
Should we be using HEAD of the v3 release branch so eventually we can pin it to non-beta 3.0 tag when things stabilize?

PR #460, opened 2017-04-05
Default to POD_NAMESPACE for -n flag

kyessenov:


cmd/manager/main.go
```diff
@@ -187,20 +202,17 @@ func init() {
 	discoveryCmd.PersistentFlags().BoolVar(&flags.enableProfiling, "profile", true,
 		"Enable profiling via web interface host:port/debug/pprof")
 
-	proxyCmd.PersistentFlags().StringVar(&flags.identity.IP, "nodeIP", "",
-		"Proxy node IP address. If not provided uses ${POD_IP} environment variable.")
-	proxyCmd.PersistentFlags().StringVar(&flags.identity.Name, "nodeName", "",
-		"Proxy node name. If not provided uses ${POD_NAME}.${POD_NAMESPACE}")
+	proxyCmd.PersistentFlags().StringVar(&flags.ipAddress, "ipAddress", "",
```
ayj: What is the use case for these flags vs. the env vars?

kyessenov: Just leaving it open. Don't see any use case right now.

PR #461, opened 2017-04-05
Added api server and tests

liamawhite:
This is the work for issue #227.

Still a bit rough around the edges but all the functionality is there.


apiserver/BUILD
```diff
@@ -0,0 +1,29 @@
...
+    srcs = [
+        "apiserver.go",
+        "config.go",
+        "handlers.go",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//model:go_default_library",
+        "@com_github_emicklei_go_restful//:go_default_library",
+        "@com_github_gogo_protobuf//proto:go_default_library",
```
kyessenov: Please use golang protobuf, not gogo protobuf. Auto-import tools tend to choose gogo for some reason.

model/errors.go
```diff
@@ -0,0 +1,17 @@
+package model
```
kyessenov: we try to name source files and directories in a singular form (e.g. `error.go`).

model/errors.go
```diff
@@ -0,0 +1,17 @@
...
+
+type ItemAlreadyExistsError struct {
```
kyessenov: config `Key` should probably be a field in the struct.

test/mock/config.go
```diff
@@ -78,7 +77,7 @@ func (cr *ConfigRegistry) Delete(key model.Key) error {
 		delete(cr.data, key)
 		return nil
 	}
-	return errors.New("item is missing")
+	return &model.ItemNotFoundError{"item not found"}
```
kyessenov: you can drop the text, since it's stutter.

cmd/manager/main.go
```diff
@@ -90,10 +90,18 @@ var (
 				Port:            flags.sdsPort,
 				EnableProfiling: flags.enableProfiling,
 			}
+			apiserver := api.NewAPI(api.APIServiceOptions{
```
kyessenov: `s/api/apiserver`?

cmd/manager/main.go
```diff
@@ -90,10 +90,18 @@ var (
...
+				Version: "v1alpha1",
+				Port:    8081,
```
kyessenov: make this a flag `apiserverPort`

apiserver/config.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/glog"
+
+	"istio.io/manager/model"
+)
+
+// Config is the complete configuration including a parsed spec
+type Config struct {
```
kyessenov: can you please remove `inputDoc` from `istioctl` and use this `Config` instead there?

apiserver/config.go
```diff
@@ -0,0 +1,41 @@
...
+	// Type SHOULD be one of the kinds in model.IstioConfig; a route-rule, ingress-rule, or destination-policy
+	Type string      `json:"type,omitempty"`
+	Name string      `json:"name,omitempty"`
+	Spec interface{} `json:"spec,omitempty"`
+	// ParsedSpec will be one of the messages in model.IstioConfig: for example an
+	// istio.proxy.v1alpha.config.RouteRule or DestinationPolicy
+	ParsedSpec proto.Message `json:"-"`
+}
+
+func (c *Config) parseSpec() error {
```
kyessenov: Make it public and use in `istioctl`. We should parse proto on the client-side, and skip sending it to the API server.

apiserver/apiserver.go
```diff
@@ -0,0 +1,96 @@
...
+
+	ws.Route(ws.
+		POST(fmt.Sprintf("/config/{%s}/{%s}/{%s}", Kind, Namespace, Name)).
+		To(api.AddConfig).
+		Doc("Add a config").
+		Reads(Config{}))
+
+	ws.Route(ws.
+		PUT(fmt.Sprintf("/config/{%s}/{%s}/{%s}", Kind, Namespace, Name)).
+		To(api.UpdateConfig).
```
kyessenov: are these implemented?

apiserver/apiserver_test.go
```diff
@@ -0,0 +1,332 @@
...
+	_, ok := mockReg.Get(key)
+	if !ok && present {
+		t.Errorf("Expected config wasn't present in the registry for key: %+v", key)
+	} else if ok && !present {
+		t.Errorf("Unexpected config was present in the registry for key: %+v", key)
+	}
+	// To Do: compare protos
+}
+
+func compareJSON(jsonFile string, t *testing.T) {
```
kyessenov: please use `test/util/diff.go`. I added a flag to regenerate testdata files that should work across all usages of that diff library.

apiserver/apiserver_test.go
```diff
@@ -0,0 +1,332 @@
...
+var (
+	routeRuleKey = model.Key{Name: "name", Namespace: "namespace", Kind: "route-rule"}
+
+	validRouteRuleJSON              = []byte(`{"type":"route-rule","name":"name","spec":{"destination":"service.namespace.svc.cluster.local","precedence":1,"route":[{"tags":{"version":"v1"},"weight":25}]}}`)
+	validUpdatedRouteRuleJSON       = []byte(`{"type":"route-rule","name":"name","spec":{"destination":"service.namespace.svc.cluster.local","precedence":1,"route":[{"tags":{"version":"v2"},"weight":25}]}}`)
+	validDiffNamespaceRouteRuleJSON = []byte(`{"type":"route-rule","name":"name","spec":{"destination":"service.differentnamespace.svc.cluster.local","precedence":1,"route":[{"tags":{"version":"v3"},"weight":25}]}}`)
+
+	itemExistsResponse  = "item already exists"
+	invalidTypeResponse = "unknown configuration type not-a-route-rule; use one of [destination-policy ingress-rule route-rule]"
+	notFoundResponse    = "item not found"
```
kyessenov: These are expected error strings? I think you should instantiate the error and compare it's `Error()` instead of hardcoding the strings.

apiserver/apiserver.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+const (
+	Kind      = "kind"
+	Name      = "name"
+	Namespace = "namespace"
+)
+
+var (
+	schema model.ProtoSchema
```
kyessenov: unused variable? we have a linter script `bin/check.sh` that can help with finding dead code.

kyessenov:
Overall, looks good to me. I think some files/code is missing though.

apiserver/apiserver.go
liamawhite: yes, they are all in the handlers.go which I will now rename to handler.go.

liamawhite:
@kyessenov I should have covered all of your requested changes. I somehow managed to miss `handler.go` in my PR so I've added that as well. Planning to comment it all this evening.

rshriram:
Please update your branch.

cmd/manager/main.go
```diff
@@ -182,8 +192,10 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&flags.config, "meshConfig", cmd.DefaultConfigMapName,
 		fmt.Sprintf("ConfigMap name for Istio mesh configuration, key should be %q", cmd.ConfigMapKey))
 
-	discoveryCmd.PersistentFlags().IntVarP(&flags.sdsPort, "port", "p", 8080,
+	discoveryCmd.PersistentFlags().IntVarP(&flags.sdsPort, "sdsPort", "p", 8080,
```
ayj: This should probably be `discoveryPort` since it serves CDS/RDS as well. Also, this may cause integration test failures unless everything has been switched over to use the ConfigMap - @kyessenov, this should be a safe change from that perspective now, correct?

cmd/istioctl/main.go
```diff
@@ -325,38 +314,23 @@ func readInputs() ([]inputDoc, error) {
...
 		if err == io.EOF {
 			break
 		}
 		if err != nil {
 			return nil, fmt.Errorf("cannot parse proto message: %v", err)
 		}
-
-		// Do a second decode pass, to get the data into structured format
-		byteRule, err := json.Marshal(v.Spec)
+		err = v.ParseSpec()
```
ayj: (nit) Combine err assignment and check on the same line, e.g.

```golang
if err := v.ParseSpec(); err != nil {
    return nil, err
}
```

apiserver/config.go
```diff
@@ -0,0 +1,43 @@
...
+	schema, ok := model.IstioConfig[c.Type]
+	if !ok {
+		return fmt.Errorf("unknown spec type %s", c.Type)
+	}
+	message, err := schema.FromJSON(string(byteSpec))
+	if err != nil {
+		return fmt.Errorf("cannot parse proto message: %v", err)
+	}
+	c.ParsedSpec = message
+	glog.V(2).Info(fmt.Sprintf("Parsed %v %v into %v %v", c.Type, c.Name, schema.MessageName, message))
```
ayj: You can use `glog.V(2).Infof` here instead of Info+Sprintf.

apiserver/handler.go
```diff
@@ -0,0 +1,230 @@
...
+		return
+	}
+
+	retrieved, err := schema.ToJSON(proto)
+	if err != nil {
+		api.writeError(http.StatusInternalServerError, err.Error(), response)
+		return
+	}
+	var retJSON interface{}
+	err = json.Unmarshal([]byte(retrieved), &retJSON)
```
ayj: (nit) - combine err assignment and check on the same line here and elsewhere.

apiserver/handler.go
```diff
@@ -0,0 +1,230 @@
...
+		return
+	}
+
+	err = api.registry.Post(key, config.ParsedSpec)
+	if err != nil {
+		response.AddHeader("Content-Type", "text/plain")
+		switch err.(type) {
+		case *model.ItemAlreadyExistsError:
+			api.writeError(http.StatusConflict, err.Error(), response)
+			return
```
ayj: (nit) - The multiple `returns` can be moved outside of the switch statement since this error path returns regardless of the error type.

apiserver/handler.go
```diff
@@ -0,0 +1,230 @@
...
+	"fmt"
+	"net/http"
+
+	"istio.io/manager/model"
+
+	restful "github.com/emicklei/go-restful"
+	"github.com/golang/glog"
+)
+
+var schema model.ProtoSchema
```
ayj: Can this be moved to line 32? It would be best to avoid use of global variables as someone will inevitably use `schema = ` instead of `schema :=` somewhere and the compiler won't complain.

apiserver/apiserver_test.go
```diff
@@ -0,0 +1,327 @@
+package apiserver
```
ayj: Many of the tests in this file would be good candidates for table driven tests (see https://github.com/golang/go/wiki/TableDrivenTests). For example most (all?) of the GET/UPDATE/LIST tests can be combined into a single test with a table defining the input/output parameters. Adding a new test only requires adding a new entry to the table (see snippet below). This also avoids the need for the extra helper functions at the bottom of the file.


```golang
func TestConfig(t *testing.T) {
    cases := []struct {
        name       string
        url        string
        method     string
        data       []byte
        wantStatus int
        wantErr    error
    }{
        {
            name:       "TestNotFoundGetConfig",
            url:        "/test/config/route-rule/namespace/name",
            method:     "GET",
            wantStatus: http.StatusNotFound,
            wantBody:   errInvalidType.Error(),
        },
        {
            name:       "TestInvalidConfigTypeGetConfig",
            url:        "/test/config/not-a-route-rule/namespace/name",
            method:     "GET",
            wantStatus: http.StatusBadRequest,
            wantBody:   errInvalidType.Error(),
        },
    }

    for _, c := range cases {
        api := makeAPIServer(mock.MakeRegistry())
        gotStatus, gotBody, err := makeAPIRequest(api, c.method, c.url, c.data)
        if err != nil {
            t.Errorf("%s: makeAPIRequest failed: %v", c.name, err)
        }
        if gotStatus != c.wantStatus {
            t.Errorf("%s: got status code %v, want %v", c.name, gotStatus, c.wantStatus)
        }
        if gotBody != c.wantBody {
            t.Errorf("%s: got body %q, want %q", c.name, gotBody, c.wantBody)
        }
    }
}
```

apiserver/apiserver.go
```diff
@@ -0,0 +1,101 @@
...
+// NewAPI creates a new instance of the API using the options passed to it
+// It returns a pointer to the newly created API
+func NewAPI(o APIServiceOptions) *API {
+	out := &API{
+		version:  o.Version,
+		registry: o.Registry,
+	}
+	container := restful.NewContainer()
+	out.Register(container)
+	if o.Port != 0 {
```
ayj: What behavior do you want when port is zero?

apiserver/handler.go
ayj: (see https://github.com/istio/manager/pull/446/files#diff-d6311766e3a7c1d9998905f0760eccdbR79 where we had this exact problem)

cmd/manager/main.go
kyessenov: I can take care of that later. I like the idea of parsing the configmap to get the port (no need to deal with pod port != service port)

apiserver/apiserver.go
kyessenov: Let's remove this check. Port 0 means an arbitrary port which is not necessarily wrong.

apiserver/handler.go
kyessenov: Yes, please remove globals, they are error-prone

apiserver/apiserver_test.go
liamawhite: Agreed, I've moved all the error cases to tables but left the golden path because each test case isn't isolated so it should be explicit that they aren't. Also when I wrote them in tables I ended up doubling the size of the tests.

apiserver/handler.go
```diff
@@ -0,0 +1,226 @@
...
+
+	params := request.PathParameters()
+	key, err := setup(params)
+	if err != nil {
+		api.writeError(http.StatusBadRequest, err.Error(), response)
+		return
+	}
+
+	config := &Config{}
+	err = request.ReadEntity(config)
```
ayj: Combine `err` assignment and check on single line here and lines 72, 78, 107, 113, 119, and 145.

apiserver/handler.go
```diff
@@ -0,0 +1,226 @@
...
+
+// ListConfigs lists the configuration objects in the the configuration registry
+// If kind and namespace are passed then it retrieves all rules of a kind in a namespace
+// If kind is passed and namespace is an empty string it retrieves all rules of a kind across all namespaces
+func (api *API) ListConfigs(request *restful.Request, response *restful.Response) {
+
+	params := request.PathParameters()
+	namespace, kind := params["namespace"], params["kind"]
+
+	_, ok := model.IstioConfig[kind]
```
ayj: `ok` assignment and check can be combined to minimize scope of `ok` variable. Also applies to line 216.

sebastienvas:
Was this force merged ? I don t see any results from manager/presubmit and https://testing.istio.io/view/All%20Jobs/job/istio/job/manager-pr-master/96/ is failing

liamawhite:
@sebastienvas I think pr-master was failing beforehand the presubmit passed just after you posted ^^

sebastienvas:
We fixed manager-pr-master yesterday. It has been working so far if I look at the history

liamawhite:
Ah, I see. There was some kind of disconnect between this PR and the manager-pr-master because it had called it but it never actually reported back.

sebastienvas:
Yeah, that lasy jenkins. But still, the checks being required I d rather hold on merging. (Which is not your fault)

liamawhite:
I'll see if I can work out what the issue is locally

kyessenov:
I wasn't sure if jenkins is all working and got used to seeing all red. Sorry about that.

Issue #67, opened 2017-01-30
Egress proxy for external services


kyessenov:
For services that don't have pods backing them, there's just an external IP and no deployment with a proxy. We need to deploy a set of proxies to handle cluster egress traffic, and configure them to capture and route traffic.

@vaikas-google

rshriram:
Do we need to create a set of such dummy services adding another unnecessary hop? If we augment the routing proto to include static IPs or fqdns, what then? It might complicate code a bit but it won't require the end user to declare what services are external or be surprised to see a pod for each such service or service instance. (E.g. lb1.foo.maps.com, lb2.foo.maps.com).

kyessenov:
This issue is specifically for the `ExternalName` type services. Here is an example:
```
apiVersion: v1
kind: Service
metadata:
  name: my-app
spec:
  ports:
  - port: 12345
  type: ExternalName
  externalName: myapp.whatever.google.com
```

This service does not get a cluster IP and does not have pods in the cluster. 
Imagine we want to monitor the traffic through this service and enforce control policies like quota.
We need to route traffic through Istio Proxy, so we create a deployment with proxies. We then rewrite the external name in the service "my-app" to point to Istio proxies, and make sure Istio proxies route the incoming traffic to "myapp.whatever.google.com". With this change, Kube DNS sends proxy address to Istio and non-Istio in-cluster clients. For Istio client-side proxy-enabled clients, we can skip Kube DNS and use Envoy discovery service.

kyessenov:
Initial design for this feature.

I. Redirect traffic to external services to Istio proxy.
Manager rewrites `externalName` field to point to a dedicated k8s service that selects proxy-egress pods running Istio proxy. The original `externalName` is stored as an annotation on the service definition.
Manager also _requires port definitions_ on the external services. Ports are optional for `externalName` services starting from k8s 1.6, but Envoy needs explicit listener ports in its configuration to capture traffic and choose the right protocol filter. Manager rejects external services with no port declarations.

II. Deploy egress proxies.
There are two choices for deploying dedicated egress proxies:

a) Deploy proxies per each external services. 
Manager defines a new service for each external service to point to these proxies, which has the same port declarations and serves as the target DNS name.

b) Share egress proxies between external services.
There is one shared deployment and one service for cluster-egress proxies.
The service is managed to list all ports across external services  This shared service is the target DNS name in the external service `externalName`.
Proxies identify traffic using the external service hostname or VIP address.

[...]

PR #463, opened 2017-04-06
Implement egress proxy for external services

GregHanson:
Assumptions:
- external services can have only one port, and that port will be either 80 or 443
- Support for external services of type https will be addressed in a later pull request

See issue #67 for discussion


proxy/envoy/config.go
```diff
@@ -267,6 +268,21 @@ func buildOutboundHTTPRoutes(instances []*model.ServiceInstance, services []*mod
 					routes = append(routes, buildDefaultRoute(cluster))
 				}
 
+				if service.External {
+					for _ , route := range routes {
+						for _ , cluster := range route.clusters {
+							cluster.ServiceName = ""
```
rshriram: Why empty the service name in the cluster?

GregHanson: ServiceName is only for clusters of type `sds`

kyessenov: I wonder if we should use SDS for egress proxies. Probably not in this PR.

proxy/envoy/resources.go
```diff
@@ -73,6 +76,7 @@ type MeshConfig struct {
 var (
 	// DefaultMeshConfig configuration
 	DefaultMeshConfig = &MeshConfig{
+		EgressAddress:             "istio-egress:80",
```
kyessenov: `EgressProxyAddress` (we'll converge with ingress eventually)

proxy/envoy/resources.go
```diff
@@ -210,6 +214,8 @@ type HTTPRoute struct {
 	// faults contains the set of referenced faults in the route; the field is special
 	// and used only to aggregate fault filter information after composing routes
 	faults []*HTTPFilter
+
+	AutoHostRewrite bool `json:"auto_host_rewrite,omitempty"`
```
kyessenov: can you move it up above private fields?

platform/kube/conversion.go
```diff
@@ -83,11 +83,16 @@ func convertPort(port v1.ServicePort) *model.Port {
 
 func convertService(svc v1.Service) *model.Service {
 	ports := make([]*model.Port, 0)
-	addr := ""
+	addr, external := "", false
 	if svc.Spec.ClusterIP != "" && svc.Spec.ClusterIP != v1.ClusterIPNone {
 		addr = svc.Spec.ClusterIP
 	}
 
+	if svc.Spec.ExternalName != "" {
```
kyessenov: You should compare `svc.Spec` service type, like in https://github.com/istio/manager/blob/master/platform/kube/controller.go#L520. There is some dead code there that needs to be updated as well.

proxy/envoy/config.go
GregHanson: Yes I was debating about whether to use a default URL from MeshConfig or pointing it to SDS.  I chose the meshconfig option because I figured otherwise I would have to either store the egress model.Service or cache it somehow so I could reliably generate a `ServiceName` for the cluster

rshriram: We cannot use SDS for egress proxies. Because SDS cannot serve DNS names. Only IP/ports are allowed. If we convince Matt that SDS could allow logical_dns and strict_dns, then we might use SDS for egress as well.

GregHanson: @rshriram this is for the sidecar proxy config gen.  SDS in this case would return the IP of the egress service.  Egress proxy config gen would NOT use SDS because it will have to statically define the cluster with hosts (i.e. google.com, etc)

proxy/envoy/egress.go
```diff
@@ -0,0 +1,197 @@
...
+			vhosts = append(vhosts, buildEgressHTTPRoute(service))
+		}
+	}
+
+	sort.Slice(vhosts, func(i, j int) bool { return vhosts[i].Name < vhosts[j].Name })
+
+	rConfig := &HTTPRouteConfig{VirtualHosts: vhosts}
+
+	listener := &Listener{
+		Address:    fmt.Sprintf("tcp://%s:80", WildcardAddress),
```
rshriram: Hard coding port here? Isn't istio egress service port configurable ?

proxy/envoy/egress.go
```diff
@@ -0,0 +1,197 @@
...
+							Name:   "router",
+							Config: FilterRouterConfig{},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	//TODO
```
rshriram: Doesn't apply to egress right? Please remove comment. But add a comment related to tcp service

proxy/envoy/egress.go
kyessenov: This is egress proxy port, not the service port. @GregHanson please add a comment saying that service must have exactly one port or they get conflated.

proxy/envoy/egress.go
```diff
@@ -0,0 +1,185 @@
...
+			route.clusters = append(route.clusters, cluster)
+
+			host = &VirtualHost{
+				Name:    buildEgressClusterName(svc.Address, servicePort.Port),
+				Domains: []string{svc.Hostname},
+				Routes:  []*HTTPRoute{route},
+			}
+
+		case model.ProtocolTCP, model.ProtocolHTTPS:
+		// TODO
```
kyessenov: drop this clause, so that default case will print a warning.

proxy/envoy/egress.go
```diff
@@ -0,0 +1,185 @@
...
+// buildEgressRoute translates an egress rule to an Envoy route
+func buildEgressHTTPRoute(svc *model.Service) *VirtualHost {
+	var host *VirtualHost
+
+	// TODO Cluster names are not unique if external service definition contains more than one port
+	for _, servicePort := range svc.Ports {
+		protocol := servicePort.Protocol
+		switch protocol {
+		case model.ProtocolHTTP, model.ProtocolHTTP2, model.ProtocolGRPC:
+
```
kyessenov: extra white space?

proxy/envoy/egress.go
```diff
@@ -0,0 +1,185 @@
...
+			cluster.Hosts = []Host{
+				{
+					URL: fmt.Sprintf("tcp://%s:%d", svc.Address, servicePort.Port),
+				},
+			}
+
+			route.clusters = append(route.clusters, cluster)
+
+			host = &VirtualHost{
+				Name:    buildEgressClusterName(svc.Address, servicePort.Port),
```
kyessenov: reuse `cluster.Name`

proxy/envoy/egress.go
```diff
@@ -0,0 +1,185 @@
...
+	// TODO Cluster names are not unique if external service definition contains more than one port
+	for _, servicePort := range svc.Ports {
+		protocol := servicePort.Protocol
+		switch protocol {
+		case model.ProtocolHTTP, model.ProtocolHTTP2, model.ProtocolGRPC:
+
+			route := &HTTPRoute{
+				Prefix:          "/",
+				Cluster:         buildEgressClusterName(svc.Address, servicePort.Port),
+				AutoHostRewrite: true,
```
kyessenov: Should we just put `HostRewrite` to the external name? It's known ahead of time based on the cluster definition.

test/mock/service.go
```diff
@@ -23,12 +23,20 @@ import (
...
 // Mock values
 var (
-	HelloService                        = MakeService("hello.default.svc.cluster.local", "10.1.0.0")
-	WorldService                        = MakeService("world.default.svc.cluster.local", "10.2.0.0")
-	Discovery    model.ServiceDiscovery = &ServiceDiscovery{
+	HelloService   = MakeService("hello.default.svc.cluster.local", "10.1.0.0")
+	WorldService   = MakeService("world.default.svc.cluster.local", "10.2.0.0")
+	ExtHTTPService = MakeExternalHTTPService("httpbin.default.svc.cluster.local",
+		"httpbin.org", "10.3.0.0")
+	ExtHTTPSService = MakeExternalHTTPSService("httpsbin.default.svc.cluster.local",
```
kyessenov: please remove this until it is used.

platform/kube/conversion.go
```diff
@@ -82,11 +82,16 @@ func convertPort(port v1.ServicePort) *model.Port {
...
 func convertService(svc v1.Service) *model.Service {
 	ports := make([]*model.Port, 0)
-	addr := ""
+	addr, external := "", false
 	if svc.Spec.ClusterIP != "" && svc.Spec.ClusterIP != v1.ClusterIPNone {
 		addr = svc.Spec.ClusterIP
 	}
 
+	if svc.Spec.Type == v1.ServiceTypeExternalName && svc.Spec.ExternalName != "" {
+		addr = svc.Spec.ExternalName
```
kyessenov: I think you should add a field to `Service` called `ExternalName` to hold this value. Address is needed for TCP/HTTPS routing for the in-cluster IP address.

test/mock/service.go
```diff
@@ -79,6 +115,9 @@ func MakeInstance(service *model.Service, port *model.Port, version int) *model.
 // MakeIP creates a fake IP address for a service and instance version
 func MakeIP(service *model.Service, version int) string {
 	ip := net.ParseIP(service.Address).To4()
+	if ip == nil {
```
kyessenov: This isn't needed if you add a field for the external name.

platform/kube/conversion.go
rshriram: Do you mean adding it to model.Service ? If so, then it simply conflates the model. Services already have a hostname in the model and there is nothing in the model that says a service is in cluster or not. So adding an ExternalName makes little sense from a generic model perspective. 

If anything, I would add a type field to the model.Service, that can qualify a service as either "internal/external", deployment mode as a boolean "user-facing or not". This would let ingress/gateway Envoy pick up only user-facing services of internal Type, non-k8s environments), egress Envoy pick external services while other envoys can pickup internal services and external service (with a host rewrite to egress envoy).

test/mock/service.go
GregHanson: I left it in there originally because until external https services are supported - it should not appear in RDS/CDS/egress outputs.  And until they are supported, if someone pushes a change to the config generation logic - I wanted there to be something in place to catch it

proxy/envoy/egress.go
GregHanson: In CloudFoundry case: `catalog-1.mystore.com`, `catalog-2.mystore.com`, `catalog-3.mystore.com` are all under under `catalog.mystore.com`
Although the current egress implementation does not support this yet, I tried to model the envoy config with this in mind

test/mock/service.go
GregHanson: as a general istio user wishing to use an external service - are they required to know the hostname (google.com) and the IP it resolves to?  Or from a coding standpoint then, if an external service is specified how do we populate the IP field, or would we just leave IP/Address as an empty string in this case? And if left empty, we would still need this `ip == nil` check in the code or else the next few lines trigger an index out of bounds exception

platform/kube/conversion.go
GregHanson: I can also use service.Hostname to store the ExternalName value and just not touch service.Address in the external case .  I originally chose to overwrite service.Address with ExternalName because I wanted to preserve the `service.default.local.svc` value because the logic for generating it was not exposed.

test/mock/service.go
rshriram: I think this will be the case until we start using Endpoints for external services

proxy/envoy/egress.go
```diff
@@ -0,0 +1,185 @@
...
+	// to avoid racing with controller events lurking around the corner.
+	// This can be improved once we switch to a mechanism where reloads
+	// are linearized (e.g., by a single goroutine reloader).
+	w.reload()
+	w.ctl.Run(stop)
+}
+
+// EgressConfig defines information for engress
+type EgressConfig struct {
+	Namespace string
```
GregHanson: This struct is still used as the input struct for for `generateEgress()`

proxy/envoy/config.go
```diff
@@ -269,6 +270,22 @@ func buildOutboundHTTPRoutes(instances []*model.ServiceInstance, services []*mod
 					routes = append(routes, buildDefaultRoute(cluster))
 				}
 
+				if service.External {
+					for _, route := range routes {
+						route.HostRewrite = service.Hostname
```
GregHanson: it was a design choice.  For example the proxy in the pod matches on this list of domains:
```
     "httpbin:80",
     "httpbin",
     "httpbin.default:80",
     "httpbin.default",
     "httpbin.default.svc:80",
     "httpbin.default.svc",
     "httpbin.default.svc.cluster:80",
     "httpbin.default.svc.cluster",
     "httpbin.default.svc.cluster.local:80",
     "httpbin.default.svc.cluster.local",
     "httpbin.org:80",
     "httpbin.org"
```
By setting route.HostRewrite to a known value, the egress proxy explicitly only listens/matches for one of those domains:
```
    "httpbin.org"
```

kyessenov: OK, I see. I don't think "httpbin.org" should be in that list. This doesn't sound right @rshriram

test/mock/service.go
kyessenov: OK, so it's a negative test. That's fine then.

kyessenov:
The way I see it, `ExternalName` service is still a regular Istio service with a hostname (cluster-local), an IP address, but no actual pods. Instead, the service is redirected to out-of-cluster DNS name. Logically, this means external name service is a subtype of a service, so we need to add fields to express this in golang.

proxy/envoy/config.go
rshriram: what do you mean? The application addresses the external service directly by the name and not via the kubernetes version of the service (httpbin.default.svc.cluster.local).

rshriram:
ExternalName service is a platform specific implementation detail - with respect to kubernetes. The way I see it, external service is just an extension of the Istio mesh, allowing services outside the mesh to be represented in Istio. Doing the external name exposes the internal implementation detail of having an explicit istio-egress service. If we move towards a model where the sidecars directly route to external services (after talking to mixer), then this whole thing becomes redundant.

PR #465, opened 2017-04-07
Update Jenkinsfile

sebastienvas:


sebastienvas:
build istio/manager-pr-master

Jenkinsfile
```diff
@@ -78,7 +78,7 @@ def stablePresubmit(gitUtils, bazel, utils) {
     sh('ln -s ~/.kube/config platform/kube/')
     stage('Integration Tests') {
       timeout(30) {
-        sh("bin/e2e.sh -count 10 -debug -tag debug${gitUtils.GIT_SHA} -v 2")
```
kyessenov: these are different images, don't conflate them under one SHA.

sebastienvas: it won t be the same SHA as one SHA is from a PR to master, while the second will be the SHA after merge

kyessenov: Which ones are being pushed to docker.io though?

sebastienvas: it is not great, but otherwise I need to pass in another build parameter to istio/istio Jenfkinsfile.

sebastienvas: we are rebuilding them at https://github.com/istio/manager/blob/master/Jenkinsfile#L95

kyessenov: Ok, that's ok then. I still wonder what's the difference between post-submit and staging from master...

Issue #372, opened 2017-03-22
Istio config resources can be created for non-existent namespaces


kyessenov:
Apparently, k8s allows you to create TPRs for non-existent namespaces:
`istioctl create -f ... -n idontexist`
You can create namespace later and the resource is there. Resource is correctly cleaned up if you delete namespace though.

kyessenov:
@sebastienvas discovered this

rshriram:
Should we investigate using something like redis (like what mixer is doing) as our datastore backend? Or force mixer to go the other way. It seems we are heading for a data management nightmare with multiple data stores for the control plane (one platform dependent and one platform independent).

kyessenov:
This is a bug specifically for kube binding. The design of the data store for Mixer is still under consideration. I agree that we should converge but I don't want to be blocked by lack of proper Mixer config. @mandarjog

PR #467, opened 2017-04-07
Check namespace on istioctl create

esnible:
Fix for Issue #372 .  Istioctl does not check namespace on create


platform/kube/client_test.go
```diff
@@ -121,3 +123,16 @@ func deleteNamespace(cl kubernetes.Interface, ns string) {
 		glog.Infof("Deleted namespace %s", ns)
 	}
 }
+
+func TestPostFailsOnInvalidNamespace(t *testing.T) {
```
kyessenov: I think we can leave client_test and mock config unmodified. The fact that k8s allows access to non-existent namespaces is a bug, and we don't need to hardcode that. Basically, if namespace doesn't exist, the underlying kube client should be throwing an error without any special handling by us.

Issue #469, opened 2017-04-07
istioctl CLI should use hub and tag from istio.VERSION if defined


esnible:
istio.VERSION currently defines

```
export MANAGER_HUB="gcr.io/istio-testing"
export MANAGER_TAG="alpha9a73dd7feb916a7af889b94558b579ccee261a26"
```

If these are exported, _istioctl_ should default --hub and --tag to these values.

PR #473, opened 2017-04-07
Let MANAGER_HUB and MANAGER_TAG env vars override defaults

esnible:
Let MANAGER_HUB and MANAGER_TAG env vars override defaults in inject.go if present

These variables are defined in _istio/istio/istio.VERSION_.

See issue #469

### DIFF
--- a/cmd/istioctl/inject.go
+++ b/cmd/istioctl/inject.go
@@ -27,6 +27,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	DefaultHubEnvVar       = "MANAGER_HUB"
+	DefaultTagEnvVar       = "MANAGER_TAG"
+)
+
 var (
 	hub             string
 	tag             string
@@ -59,6 +64,12 @@ Example usage:
 		RunE: func(_ *cobra.Command, _ []string) (err error) {
 			if inFilename == "" {
 				return errors.New("filename not specified (see --filename or -f)")
+			}
+			if hub == "" {
+				return fmt.Errorf("specify --hub or define %v", DefaultHubEnvVar)
+			}
+			if tag == "" {
+				return fmt.Errorf("specify --tag or define %v", DefaultTagEnvVar)
 			}
 			var reader io.Reader
 			if inFilename == "-" {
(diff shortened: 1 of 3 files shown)


platform/kube/inject/inject.go
```diff
@@ -40,7 +40,9 @@ import (
 // Defaults values for injecting istio proxy into kubernetes
 // resources.
 const (
+	DefaultHubEnvVar       = "MANAGER_HUB"
```
ayj: `platform/kube/inject/inject.go` doesn't use `Default{Tag,Hub}EnvVar` at all so I'm not sure it makes sense to define them in this file. Probably better to keep that in `cmd/istioctl/inject.go`? In hindsight the DefaultHub and DefaultTag should also probably be under cmd/ instead of platform/

cmd/istioctl/inject.go
```diff
@@ -107,10 +107,20 @@ Example usage:
 )
 
 func init() {
+	defaultHub := os.Getenv(inject.DefaultHubEnvVar)
```
ayj: Any concern about this being an implicit change based on an environment variable that could be different/non-existent based on which terminal sessions a user runs `istioctl`? For example, I frequently have multiple windows open (w/tmux) and run interactive commands in one window and monitor output in another. Sometimes I accidently switch the window I'm using for interactive commands. In this case the envvar might not be set but inject would proceed without printing an error and it might take me some time to figure out I'm using the wrong injected version. I'm wondering if it might be better to remove the default values altogether and explicitly require the user to specify hub/tag with flags or envvars. Thoughts?

esnible: Long term having these variables is a mistake, because in the enterprise there will be some admins with the ability to create the manager and some mid-level ops who only have the authority to deploy their microservice.  Those folks won't know or care about the manager version.

Better would be for the manager on startup to write it's hub and tag to the istio configmap in its namespace.  By default if --hub and --tag aren't used is to have _istioctl kube-inject_ should find the manager version by asking the configmap.

platform/kube/inject/inject.go
esnible: I put the new constants with the old constants.

I can't reason about the best structure for these constants because we plan to support multiple platforms soon.  The other platforms will have different default constants and probably entirely different flags in some cases.  Keeping the Kubernetes specific constants in platform/kube reinforces that.

ayj: FWIW, the current cmd/istioctl/inject.go is _very_ kubernetes specific (i.e. `kube-inject` and not `inject`).

cmd/istioctl/inject.go
ayj: Long term the plan is kubernetes admission controller which should do all of this injection server-side. 

Shorter term ConfigMap approach could work. There are corner cases to deal with there as well, e.g. ensure user runs `kube-inject` sufficient time after ConfigMap is created/updates so that the most recent proxy version is picked up.

PR #481, opened 2017-04-08
Adds a faster regression suite for manager deps.

sebastienvas:
- No need to run linter and other as code does not change
- Triggering fast forward as last step of POSTSUBMIT stage.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 #!groovy
 
-@Library('testutils@stable-2d6eb00')
+@Library('testutils@stable-518be06')
 
 import org.istio.testutils.Utilities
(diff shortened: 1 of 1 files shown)


kyessenov:
presubmit failed with a true bug (possibly envoy). feel free to retry.

sebastienvas:
Thanks. I also noticed that we starting having issue when we upgraded the e2e cluster to 1.6.

kyessenov:
Oh, we're on 1.6 now? Good. My guess is that k8s got faster in sending watcher events, and that is triggering race conditions.

sebastienvas:
Yeah both clusters are on 1.6

PR #483, opened 2017-04-09
Touching platform/kube/config file before build

sebastienvas:


sebastienvas:
The issue is that I am not running any test here. I am just building. Why is it preventing me from building ?

sebastienvas:
Just doing it in Jenkins now.

PR #495, opened 2017-04-11
Let discovery list all services from /v1/registration/

esnible:
This is an implementation of the SDS part of #403 .  (The CDS and RDS parts can arrive in a subsequent PRs immediately following this one).  I separated the function out to keep the PR small and get early feedback.

The idea is to let tools see a global view of the services rather than just their dependencies.  It may help with debugging.

The JSON returned by "/v1/registration/" doesn't exactly match the suggestion in the Issue because I wanted to follow the convention of being able to serialize into a well formed structure and the example could only have gone into a map[string]hosts.


proxy/envoy/discovery.go
```diff
@@ -302,6 +323,42 @@ func (ds *DiscoveryService) clearCache() {
...
+			servicesArray = append(servicesArray, &service{
+				Key:   svc.Key(port, nil),
+				Hosts: hostArray,
+			})
+		}
+	}
+
+	// Sort servicesArray.  This is not strictly necessary, but discovery_test.go will
+	// be comparing against a golden example using test/util/diff.go which does a textual comparison
+	sort.Sort(serviceSorter(servicesArray))
```
ayj: You can use [sort.Slice](https://golang.org/pkg/sort/#Slice) here instead of implementing sort.Interface

proxy/envoy/discovery.go
```diff
@@ -302,6 +323,42 @@ func (ds *DiscoveryService) clearCache() {
...
+
+	var err error
+	var out []byte
+	if out, err = json.MarshalIndent(services{Services: servicesArray}, " ", " "); err != nil {
```
ayj: You can use `response.WriteEntity(services{Services: servicesArray})` here since we don't need to cache the marshalled JSON from subsequent requests.

proxy/envoy/discovery.go
```diff
@@ -302,6 +323,42 @@ func (ds *DiscoveryService) clearCache() {
...
 
+// ListAllEndpoints responds with all Services and is restricted to a single service key
+func (ds *DiscoveryService) ListAllEndpoints(request *restful.Request, response *restful.Response) {
+
+	servicesArray := make([]*service, 0)
+	for _, svc := range ds.services.Services() {
+		for _, port := range svc.Ports {
+
+			hostArray := make([]*host, 0)
+			for _, ep := range ds.services.Instances(svc.Hostname, []string{port.Name}, []model.Tags{}) {
```
ayj: (nit): you can use `nil` instead of an empty slice here for the tags.

proxy/envoy/discovery.go
```diff
@@ -302,6 +321,37 @@ func (ds *DiscoveryService) clearCache() {
 	ds.rdsCache.clear()
 }
 
+// ListAllEndpoints responds with all Services and is restricted to a single service key
```
frankbu: What do you mean by "and is restricted to a single service key"? Did you mean "is not restricted ..."?

proxy/envoy/discovery.go
```diff
@@ -225,6 +235,15 @@ func (ds *DiscoveryService) Register(container *restful.Container) {
 	ws := &restful.WebService{}
 	ws.Produces(restful.MIME_JSON)
 
+	// List all known services (informational, not invoked by Envoy)
+	ws.Route(ws.
+		GET(fmt.Sprintf("/v1/registration/")).
```
ijsnellf: Remove trailing slash and unused `fmt.Sprintf`:
`GET("/v1/registration").`

proxy/envoy/discovery.go
```diff
@@ -302,6 +321,37 @@ func (ds *DiscoveryService) clearCache() {
...
 
+// ListAllEndpoints responds with all Services and is restricted to a single service key
+func (ds *DiscoveryService) ListAllEndpoints(request *restful.Request, response *restful.Response) {
+
+	var servicesArray []*service
+	for _, svc := range ds.services.Services() {
+		for _, port := range svc.Ports {
+
+			var hostArray []*host
+			for _, ep := range ds.services.Instances(svc.Hostname, []string{port.Name}, []model.Tags{}) {
```
ijsnellf: Since we know the size out the output array, you can pre-allocate the space that you'll need.
```
instances := ds.services.Instances(...)
hosts := make([]*host, 0, len(hosts))
for _, ep := range instances {
   ...
}
```

ayj: Preallocating the slice is not strictly necessary since this isn't in the fast path. That said, the original declaration is not 100% correct since hostArray is eventually encoded as a JSON object (see https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices). 

Any of the following variants should work, though I have a preference for the first since it's the least amount of code.

```golang
hosts := []*host
hosts := make([]*host)
hosts := make([]*host, 0, len(instances))
```

esnible:
I have modified the expected output.  It is now an array of "service-key" and "hosts", with one entry for each unique service key.  After working on the all-cluster implementation decided the JSON tag for key should match the parameter for the regular list, and that there was no need to wrap the returned array in another set of braces.

proxy/envoy/discovery.go
```diff
@@ -321,33 +317,33 @@ func (ds *DiscoveryService) clearCache() {
...
 }
 
-// ListAllEndpoints responds with all Services and is restricted to a single service key
+// ListAllEndpoints responds with all Services and is not restricted to a single service-key
 func (ds *DiscoveryService) ListAllEndpoints(request *restful.Request, response *restful.Response) {
 
-	var servicesArray []*service
-	for _, svc := range ds.services.Services() {
-		for _, port := range svc.Ports {
+	var services []*keyAndService
```
ayj: Extra care should be taken when declaring `services` and `hosts` since they will be encoded as JSON objects. See https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices.

Any one of the following variants should work:
```golang
hosts := []*host
hosts := make([]*host)
hosts := make([]*host, 0, len(instances))
```

esnible:
I had planned to deliver the listing of all services separate to keep the PRs small and quick but decided to combine listing all routes and all clusters in this PR due to the difficulties getting Jenkins to build cleanly.

proxy/envoy/discovery.go
```diff
@@ -412,3 +572,26 @@ func writeResponse(r *restful.Response, data []byte) {
 		glog.Warning(err)
 	}
 }
+
+// Get a map where the keys are the service nodes (typically IPv4 addresses) and the values are all true
+func allServiceNodes(ds *DiscoveryService) []string {
```
ayj: `allServiceNodes` can be made a method of `DiscoveryService`, e.g. 
```golang
func (ds *DiscoverySerivce) allServiceNodes() []string { ... }
```

proxy/envoy/discovery.go
```diff
@@ -356,6 +479,43 @@ func (ds *DiscoveryService) ListClusters(request *restful.Request, response *res
...
+		services := ds.services.Services()
+		httpRouteConfigs := buildOutboundHTTPRoutes(instances, services, &ProxyContext{
+			Discovery:  ds.services,
+			Config:     ds.config,
+			MeshConfig: ds.mesh,
+			IPAddress:  ip,
+		})
+
+		for port, httpRouteConfig := range httpRouteConfigs {
+
```
ayj: (nit): extra newline here and elsewhere at the beginning of function and conditional blocks is not necessary.

PR #500, opened 2017-04-12
Fix get_workspace_status for Jenkins

sebastienvas:
The builds done with Jenkins were marked as dirty. The reason was that we update the tools/bazel.rc with the content of tools/bazel.rc.jenkins. Version stable-70699be of fixes that. Update pipeline to build istioctl first.

### DIFF
--- a/bin/get_workspace_status
+++ b/bin/get_workspace_status
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 BUILD_GIT_REVISION=$(git rev-parse --short HEAD 2> /dev/null)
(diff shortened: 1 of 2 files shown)


bin/get_workspace_status
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 BUILD_GIT_REVISION=$(git rev-parse --short HEAD 2> /dev/null)
+GIT_SHA="${BUILD_GIT_REVISION}"
```
kyessenov: why add another variable?

sebastienvas: changed it a little bit

PR #504, opened 2017-04-12
API client initial functionality

liamawhite:
Go client for the manager apiserver. This is the first half of the work required for making istioctl talk to the manager rather than directly to kube.

Comments will be coming in a bit.


liamawhite:
@rshriram @kyessenov @ayj

rshriram:
Rebase please.

rshriram:
And a bit more description on the PR for tracking purposes:).

liamawhite:
@rshriram done and done

client/config.go
```diff
@@ -0,0 +1,114 @@
...
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	config := []apiserver.Config{}
+	if err = json.Unmarshal(body, &config); err != nil {
+		return nil, err
+	}
+	return config, nil
```
rshriram: This function can be removed and we could reuse getconfig if we make namespace and name optional in the doConfigCrud

client/config_test.go
```diff
@@ -0,0 +1,274 @@
...
+		switch c.function {
+		case "get":
+			config, err = client.GetConfig(c.key)
+		case "add":
+			err = client.AddConfig(c.key, *c.config)
+		case "update":
+			err = client.UpdateConfig(c.key, *c.config)
+		case "delete":
+			err = client.DeleteConfig(c.key)
+		case "list":
```
rshriram: We no longer have a list in istioctl. We do get route-rules, destination-policies, etc. similar to kubectl. But if you think this is RESTful, then please ignore this comment and the previous one about deduplicating ListConfig

liamawhite: Ah, well I suppose it doesn't matter if the client has the code there to do it? Would make adding the istioctl functionality at a later date easier...

client/client.go
```diff
@@ -0,0 +1,64 @@
...
+	"istio.io/manager/model"
+)
+
+type ManagerClient struct {
+	base             *url.URL
+	versionedAPIPath string
+	client           *http.Client
+}
+
+func NewManagerClient(base *url.URL, apiVersion string, client *http.Client) *ManagerClient {
```
ayj: This function is modifying the original `url.URL` pass despite the use of `trimmedBase`. You can pass by value instead of pointer if that is not intended.

Also, this could be rewritten to use strings.TrimSuffix and strings.TrimPrefix, e.g.

```golang
base.Path = strings.TrimSuffix(base.Path, "/")
apiVersion = strings.TrimPrefix(strings.TrimSuffix(apiVersion, "/"), "/")
```

client/client.go
```diff
@@ -0,0 +1,64 @@
...
+		trimmedBase.Path = trimmedBase.Path[:len(trimmedBase.Path)-1]
+	}
+	trimmedVersion := apiVersion
+	if strings.HasSuffix(trimmedVersion, "/") {
+		trimmedVersion = trimmedVersion[:len(trimmedVersion)-1]
+	}
+	if strings.HasPrefix(trimmedVersion, "/") {
+		trimmedVersion = trimmedVersion[1:]
+	}
+	if client == nil {
```
ayj: Is this really necessary? Seems like it would complicate the API usage and it isn't much trouble the require the user to always pass in a `http.Client`. Also, without this extra check this function and the TrimSuffix/TrimPrefix changes this function can be written as:

```golang
return &ManagerClient{
    base:             strings.TrimSuffix(base.Path, "/"),
    versionedAPIPath: strings.TrimPrefix(strings.TrimSuffix(apiVersion, "/"), "/"),
}
```

client/config.go
```diff
@@ -0,0 +1,111 @@
...
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"istio.io/manager/apiserver"
+	"istio.io/manager/model"
+)
+
+func (m *ManagerClient) GetConfig(key model.Key) (*apiserver.Config, error) {
+
```
ijsnellf: super nit: I think the most common style is no leading whitespace at the start of functions?

client/client.go
```diff
@@ -0,0 +1,64 @@
+package client
```
ayj: It seems like `client/client.go` and `client/config.go` should be combined since both files are fairly short and tightly coupled.

Also, we will be adding the equivalent client support for mixer configuration soon to the manager repo. It might be good to rename this package `proxy` and then we'd end up with the following directory/package layout, e.g.

```
client/proxy.go         // package proxy
client/proxy_test.go  
client/mixer.go         // package mixer
client/mixer_test.go  
```

client/config_test.go
```diff
@@ -0,0 +1,274 @@
...
+func TestGetAddUpdateDeleteConfig(t *testing.T) {
+
+	cases := []struct {
+		name           string
+		function       string
+		key            model.Key
+		kind           string
+		namespace      string
+		config         *apiserver.Config
+		expConfig      *apiserver.Config
```
ayj: s/exp/want

client/config_test.go
```diff
@@ -0,0 +1,274 @@
...
+func (f *FakeHandler) HandlerFunc(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(f.expStatusCode)
+	for k, v := range f.expHeaders {
+		w.Header().Set(k, v[0])
+	}
+	w.Write([]byte(f.expResponse))
+}
+
+var (
+	errInvalidSpec = errors.New("cannot parse proto message: json: " +
```
ayj: Testing for specific error strings seems fragile and it looks like the HTTP status codes, which are canonical, should be sufficient.

client/config_test.go
```diff
@@ -0,0 +1,274 @@
...
+		case "delete":
+			err = client.DeleteConfig(c.key)
+		case "list":
+			configSlice, err = client.ListConfig(c.kind, c.namespace)
+		default:
+			t.Fatal("didn't supply function to test case, don't know which client function to call")
+		}
+
+		// Verify
+		if c.expError == nil && err != nil {
```
ayj: It would be useful to include the test name in each of these failures so developers don't need to inspect the source (or re-run the test) to understand the specific test failure.

client/client.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+type ManagerClient struct {
+	base             *url.URL
+	versionedAPIPath string
+	client           *http.Client
+}
+
+func NewManagerClient(base *url.URL, apiVersion string, client *http.Client) *ManagerClient {
+	trimmedBase := base
```
liamawhite: Its ensuring we are consistent with what we use as the url and apiversion regardless of whether people pass us stuff with trailing and leading slashes. So we don't end up making requests like: `http://url//v1/config...` or `http://url/v1//config...`.

client/client.go
```diff
@@ -0,0 +1,64 @@
...
+
+	fullURL, err := url.Parse(fmt.Sprintf("%s/%s/%s",
+		m.base.String(), m.versionedAPIPath, request.URL.String()))
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse URL: %v", err)
+	}
+	request.URL = fullURL
+	response, err := m.client.Do(request)
+	if response.StatusCode < 200 || response.StatusCode >= 300 {
+		defer response.Body.Close()
```
liamawhite: Its the responsibility of the function that opens it to close it. The body is only opened in this if statement or (if there's no error) its opened by the downstream function that called client .

client/client.go
liamawhite: Sounds sensible although maybe the following is more idiomatic?
```
client/proxy/proxy.go         // package proxy
client/proxy/proxy_test.go  
client/mixer/mixer.go         // package mixer
client/mixer/mixer_test.go  
```
Single package name per directory?

client/client.go
liamawhite: Good catch!

client/client.go
```diff
@@ -0,0 +1,64 @@
...
+		if err == nil && len(body) > 0 {
+			if response.StatusCode == 404 {
+				return nil, &model.ItemNotFoundError{Msg: string(body)}
+			}
+			return nil, errors.New(string(body))
+		} else {
+			return nil, fmt.Errorf("received non-success status code %v", response.StatusCode)
+		}
+	}
+	return response, err
```
liamawhite: Not that I can think of? But I can see the issue with blanket status code. The idea was that the body gives more information but maybe we could just return an error based on status code?

client/config.go
liamawhite: We can but it gets pretty messy. If it was more than just one list doing it then I would but I don't think we gain much by doing this.

liamawhite:
@rshriram @ayj @ijsnellf I have addressed all of your comments either with updates or more comments.

client/config_test.go
rshriram: we have get route-rules instead of list route-rule.

rshriram:
Presubmit still fails

client/config_test.go
liamawhite: get route-rules in istioctl would just call list route-rule, no?

client/client.go
ayj: Whoops, yes single package per directory would be better.

client/client.go
ijsnellf: It is "opened" when you make the request. You always have to close the `resp.Body`.

client/client.go
ayj: It looks like the callers of `do` always ignore the response if `err != nil`. As such, I think this function would read easier if it returned immediately if `m.client.Do` returns error. Then the return flow would be as follows:

1) error on network connectivity problem (err != nil)
2) error on non-2XX status (err != nil)
3) success (err == nil)

client/proxy/proxy.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// GetConfig retrieves the configuration resource for the passed key
+func (m *ManagerClient) GetConfig(key model.Key) (config *apiserver.Config, err error) {
+	response, err := m.doConfigCRUD(key, http.MethodGet, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() { err = response.Body.Close() }()
```
ijsnellf: I don't think this will have desirable behavior. If line 62 is true, then the defer will run. If `response.Body.Close()` returns `nil`, we return `nil, nil`, which is incorrect. If it returns an err, then we will lose the error being returned in line 63.

client/proxy/proxy.go
```diff
@@ -0,0 +1,147 @@
...
+	request, err := http.NewRequest(http.MethodGet, reqURL, nil)
+	if err != nil {
+		return nil, err
+	}
+	response, err := m.do(request)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() { err = response.Body.Close() }()
```
ijsnellf: See above comment.

client/proxy/proxy.go
ayj: Good catch. In this case we probably don't care whether the response body was closed successfully or not so we could simply drop the error? e.g.

```golang
defer func() ( _ = response.Body.Close() }()
```

ijsnellf: Is there a reason we care if `response.Body.Close()` returns an error? We've already read it without an issue.

ayj: Agreed. I think dropping the error here and elsewhere should be fine.

ijsnellf: @ayj: I could just be `defer response.Body.Close()`

liamawhite: It can't because it fails the linting (for both suggestions)
`warning: Errors unhandled.,LOW,HIGH (gas) -> defer func() ( _ = response.Body.Close() }()`
`warning: error return value not checked (defer response.Body.Close()) (errcheck) -> defer response.Body.Close()`

ayj: Maybe just disable the linter warning altogether? (see https://github.com/GoASTScanner/gas#annotating-code)

```golang
defer func() {
    /* #nosec */
    response.Body.Close()
}()
```

Issue #474, opened 2017-04-07
Pod health probes are not passing through the proxy


kyessenov:
If a pod declares a pod health probe using HTTP GET method, then the proxy drops the health check. The problem is that the pod health probe can use any port on the pod, not necessarily defined by the associated service or by the pod container ports. The ideal solution would be to simply not trap health probes, but it is not clear how to identify health check packets. 

The probes are defined per pod, meaning the services model does not have to be extended with the information about each service instance health ports - a service instance will never need to know about another instance's health port.

Therefore, I suggest we simply add a flag to the proxy agent to open specific ports and populate this flag in kube-inject script. For example, if the pod defines livenessProbe on port 3333, we will need to pass "3333" to the proxy, and the proxy can open this port if it is not already opened.

kyessenov:
Thoughts? @ayj @andraxylia @rshriram

kyessenov:
It seems we cannot identify the health probes easily. These are sent by node agent (kubelet) here: https://github.com/kubernetes/kubernetes/blob/master/pkg/probe/http/http.go#L61

ayj:
Looks like we would need to pass  `path` as well as `port and do so for both liveliness and readiness probes, right? 

I'm assuming "open this port" means updating envoy configuration to pass through HTTP health checks? If so, how do we reconcile `path/port` combinations with existing routes from RDS that are coming from remote discovery service?

kyessenov:
We don't need path in the probe. In fact, there are TCP liveness checks as well. If there is already a service port used for health checking, then it should work as-is since the HTTP route adds POD_IP:POD_PORT as one of the possible hosts. Am I missing something?

ayj:
I was thinking health checks needed to be HTTP (not HTTPS), but maybe I'm confusing that with L4 health checks for GCLB. But assuming HTTPS, how does that work with mTLS and Istio CA? Also, do we need to exclude health checks from things like quota/access checks?

kyessenov:
Good point. I'll file an issue regarding auth + health probes problem.

Without auth, if the port is not used by the service, then we will use tcp_proxy on that pod_ip, port combination. If the port is already used, then we do not do anything, and trust that the port is handled correctly (e.g. if the port is declared to be HTTP port, then the liveness probe will not attempt to use non-HTTP probes). 

The app has to handle the path in the probe correctly, we just need to let the packets through the proxy.

[...]

PR #505, opened 2017-04-12
Add support to passthrough platform health probes

kyessenov:
Reference #474.
`kube-inject` extract both liveness and readiness ports and passes it along to the sidecar proxy. 

Other non-functional refactoring changes:
- move proxy content to top-level and use it whenever appropriate
- clean up main function to reuse existing structs in the flags
- introduce `ServiceAccounts` interface in addition to `ServiceDiscovery` to keep auth functionality isolated
- rename `ControllerConfig` to `ControllerOptions` to be consistent with other code


proxy/context.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// Config interface for listing routing rules
+	Config *model.IstioRegistry
+
+	// MeshConfig defines global configuration settings
+	MeshConfig *proxyconfig.ProxyMeshConfig
+
+	// IPAddress is the IP address of the proxy used to identify it and its
+	// co-located service instances
+	IPAddress string
```
rshriram: What does this mean? Specifically, what do you mean by colocated service instances? Concrete examples would help.

proxy/context.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// PassthroughPorts is a list of ports on the proxy IP address that must be
+	// allowed through irrespective of the services model. In case, there is an
+	// overlap, that is the port is utilized by a service for a service instance,
+	// the passthrough port can be assumed to be handled. These ports are
+	// utilized by the underlying platform infrastructure for health checking.
```
rshriram: Clarify the "can be assumed to be handled" part. The subsequent sentence about using the ports for health checking seems specific to k8s.

proxy/envoy/route.go
```diff
@@ -66,13 +66,13 @@ func buildDefaultRoute(cluster *Cluster) *HTTPRoute {
...
+func buildInboundCluster(port int, protocol model.Protocol, timeout *duration.Duration) *Cluster {
 	cluster := &Cluster{
-		Name:     fmt.Sprintf("%s%d", InboundClusterPrefix, port),
-		Type:     "static",
-		LbType:   DefaultLbType,
-		Hosts:    []Host{{URL: fmt.Sprintf("tcp://%s:%d", "127.0.0.1", port)}},
-		hostname: hostname,
+		Name:             fmt.Sprintf("%s%d", InboundClusterPrefix, port),
+		Type:             "static",
+		ConnectTimeoutMs: int(convertDuration(timeout) / time.Millisecond),
```
rshriram: Sorry for not noticing this earlier. I don't think we can have a single universal connect timeout for the entire app, can we? Because these are ultimately service specific timeouts (request level, TCP level, etc.).

proxy/context.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// MeshConfig defines global configuration settings
+	MeshConfig *proxyconfig.ProxyMeshConfig
+
+	// IPAddress is the IP address of the proxy used to identify it and its
+	// co-located service instances
+	IPAddress string
+
+	// PassthroughPorts is a list of ports on the proxy IP address that must be
+	// allowed through irrespective of the services model. In case, there is an
```
ayj: Extra comma after "in case"?

proxy/context.go
```diff
@@ -0,0 +1,46 @@
...
+// Context defines local proxy context information about the global service mesh
+type Context struct {
+	// Discovery interface for listing services and instances
+	Discovery model.ServiceDiscovery
+
+	// Accounts interface for listing service accounts
+	Accounts model.ServiceAccounts
+
+	// Config interface for listing routing rules
+	Config *model.IstioRegistry
```
kyessenov: Yes, but I had to stop refactoring somewhere. Already got carried away cleaning up the top-level code.

proxy/context.go
kyessenov: Thanks.

proxy/context.go
kyessenov: Service Instances that share the IP address? Meaning for a sidecar proxy, its IP address determines the service containers in the same pod.

proxy/context.go
kyessenov: Ok, will expand the comment.

proxy/envoy/route.go
kyessenov: It's inbound cluster to localhost. I think your comment applies more to outbound clusters.

kyessenov:
Ok, I can drop this PR and split it into two if that makes it easier.

PR #510, opened 2017-04-12
Support passthrough ports on the sidecar proxy

kyessenov:
Reference #474

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -41,6 +41,7 @@ type args struct {
 	ipAddress     string
 	podName       string
 	ingressSecret string
+	passthrough   []int
 	apiserverPort int
 
 	controllerOptions kube.ControllerOptions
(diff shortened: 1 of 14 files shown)


ayj (approved):

markvincze:
@kyessenov,
Do I see it correctly that this has only been released for `itio/proxy` and not `istio/proxy_debug`? If I try to use the `--passthrough` flag with `proxy_debug`, I get the error `unknown flag: --passthrough`.  
Isn't `proxy` and `proxy_debug` the same, just with different build config? (Sorry if this is described somewhere in the docs, I couldn't find any info about this with google.)

markvincze:
@kyessenov Sorry, my bad. The difference is not between `proxy` and `proxy_debug`, but the version, because I was trying with `0.4.0` and `latest`.
So this flag is not present yet in `0.4.0`, and it'll come with the next release?

rshriram:
We dont have this anymore, since we parse health check ports correctly,

markvincze:
@rshriram,
You mean that if we have a setup like
```
kind: Deployment
...
spec:
...
  template:
    spec:
      containers:
      - name: my-service
        livenessProbe:
          httpGet:
            path: /status
            port: 5000
            scheme: HTTP
        ...
      - name: istio-proxy
        ...
```
Then the proxy should automatically pass through the port `5000`?

emedina:
@rshriram This is still an issue in 0.4.0 without mTLS

@markvincze Have you make it work?

Issue #506, opened 2017-04-12
Separate config server from discovery server


kyessenov:
We should add another command `manager config` just for apiconfig server and move that from `manager discovery`. Logically, they are distinct components.

PR #512, opened 2017-04-12
Separate discovery and apiserver

kyessenov:
Reference #506

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -118,7 +118,7 @@ var (
 	apiserverCmd = &cobra.Command{
 		Use:   "apiserver",
 		Short: "Start Istio Manager config API service",
-		RunE: func(*cobra.Command, []string) (err error) {
+		Run: func(*cobra.Command, []string) {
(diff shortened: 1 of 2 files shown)


cmd/manager/main.go
```diff
@@ -95,30 +96,36 @@ var (
...
 			go discovery.Run()
+			cmd.WaitSignal(stop)
+			return
+		},
+	}
+
+	apiserverCmd = &cobra.Command{
+		Use:   "apiserver",
+		Short: "Start Istio Manager config API service",
+		RunE: func(*cobra.Command, []string) (err error) {
```
ayj: s/RunE/Run since this function doesn't actually return an error.

ayj:
I believe istioctl is still talking directly to k8s API server so this is a non-breaking change. However, as part of or prior to @liamawhite API server changes we'll need to add another k8s deployment to istio installation for the API server.

kyessenov:
Thanks. `apiserver` has not been deployed yet. They easy way would be to bundle apiserver and discovery as two containers in the same pod.

Issue #489, opened 2017-04-10
Remove `source.service` from proxy static config


ZackButcher:
Right now the manager provides the proxy with static config populating the `source.service` attribute; it needs to be removed as `source.service` does not exist in our attribute vocabulary.

PR #513, opened 2017-04-12
Update mixer filter config

kyessenov:
- remove `source.service`
- add `source.uid` and `target.uid` to be kubernetes pod name, namespace custom URI
- add `source.ip` and `target.ip` to be proxy IPs

Reference #489

### DIFF
--- a/proxy/envoy/config_test.go
+++ b/proxy/envoy/config_test.go
@@ -236,6 +236,7 @@ func testConfig(r *model.IstioRegistry, mesh *proxyconfig.ProxyMeshConfig, insta
 		Config:     r,
 		MeshConfig: mesh,
 		IPAddress:  instance,
+		UID:        fmt.Sprintf("uid://%s", instance),
 		// 1090 is deliberately already used by the instances, 3333 requires a new listener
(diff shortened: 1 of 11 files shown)


ZackButcher:
Attribute changes LGTM, shouldn't break the istio/istio e2e stuff. I'm about to send a PR to update configs there, I'll add `{source, target}.ip` to the set of attributes it knows about.

proxy/envoy/config_test.go
```diff
@@ -236,6 +236,7 @@ func testConfig(r *model.IstioRegistry, mesh *proxyconfig.ProxyMeshConfig, insta
 		Config:     r,
 		MeshConfig: mesh,
 		IPAddress:  instance,
+		UID:        fmt.Sprintf("uid://%s", instance),
```
ayj: Wouldn't hurt to add a fake namespace (i.e. "default") here incase anyone mistakenly looks at the golden reference JSON files and infers that IPAddress == UID.

kyessenov: ah, good point. The mock stuff has no names but we it would clear the confusion to have something at least.

PR #527, opened 2017-04-13
pass quota_name property to mixer

frankbu:
By passing the `quota_name` property to mixer, I was able to see rate limiting kick in when running the bookinfo demo. Unfortunately it didn't work quite as expected. It seemed that not only the service that was configured to rate limit (ratings), but also other services (for example, when I refreshed the UI, it said that "reviews are not available" which implies the review service also was rate limiting). After another refresh, there appeared some strange message on the screen about quota (unfortunately I didn't capture it), and then after that it seemed totally broken. The productpage service was returning an internal error:
```
INTERNAL:unable to resolve config: 1 error occurred:

* unresolved attribute target.service
```
Here is the short script to run the demo to test rate limiting:
```
# setup
cd istio
source ./istio.VERSION
kubectl apply -f ./kubernetes/istio-install
cd demos/apps/bookinfo
kubectl create -f <(istioctl kube-inject -f bookinfo.yaml)
export GATEWAY_URL=$(kubectl get po -l infra=istio-ingress-controller -o jsonpath={.items[0].status.hostIP}):$(kubectl get svc istio-ingress-controller -o jsonpath={.spec.ports[0].nodePort})
istioctl create -f route-rule-all-v1.yaml
istioctl replace -f route-rule-reviews-v3.yaml

# test ratelimit
kubectl apply -f ../../mixer-config-quota-bookinfo.yaml
while true; do curl -s -o /dev/null http://$GATEWAY_URL/productpage; done

# point browser to $GATEWAY_URL/productpage and instead of ratings (stars) you should see message "Ratings not available"
```

### DIFF
--- a/proxy/envoy/policy.go
+++ b/proxy/envoy/policy.go
@@ -61,6 +61,7 @@ func insertMixerFilter(listeners []*Listener, instances []*model.ServiceInstance
 							"source.ip":  context.IPAddress,
 							"source.uid": context.UID,
 						},
+						QuotaName: "RequestCount",
 					},
 				}}, http.Filters...)
 			}
--- a/proxy/envoy/resources.go
+++ b/proxy/envoy/resources.go
@@ -131,6 +131,8 @@ type FilterMixerConfig struct {
 	// ForwardAttributes specifies the list of attribute keys and values that
 	// are forwarded as an HTTP header to the server side proxy
 	ForwardAttributes map[string]string `json:"forward_attributes,omitempty"`
+
+	QuotaName string `json:"quota_name,omitempty"`
 }
 
 // FilterFaultConfig definition


ZackButcher:
Spinning this up so I can dig into mixer logs, I'll report back my findings.

proxy/envoy/policy.go
```diff
@@ -61,6 +61,7 @@ func insertMixerFilter(listeners []*Listener, instances []*model.ServiceInstance
 							"source.ip":  context.IPAddress,
 							"source.uid": context.UID,
 						},
+						QuotaName: "RequestCount",
```
douglas-reid: won't this hard code the enforced quota for all requests?

ZackButcher: Yeah, I'm not sure that fixing the quota name statically in the envoy is actually what we want long term. But, it's what the mixerclient understands today. We should file an issue to figure out the correct way to handle this and update mixerclient  separately.

douglas-reid: @geeknoid thoughts on how this should be configured / work ?

frankbu: I was wondering about that. I thought that this causes the mixer to get called for all requests, which is a performance issue, but I'm assuming mixer would only rate limit the calls to the ratings service, since the mixer config has a `target.service` selector. Am I wrong?
```
    - selector: target.service == "ratings.*"
      aspects:
      - kind: quotas
        params:
          quotas:
          - descriptor_name: RequestCount
```

douglas-reid: To my understanding, `RequestCount` is a specifc configured quota. This looks like the proxy will be telling mixer to always use that limit (in our case, 5 qps).

I would expect that there would be some scheme for mixerclient to discover the set of quotas and then optionally decide which quota name to request in its Quota call (and/or have that done entirely on the mixer side).

mandarjog: The quota name should be part of the proxy config and it is a property of the route.
It also needs to be a `[]string`, because multiple quotas can apply for an incoming request. 

In absence of selector evaluation on the proxy, the proxy *will* call mixer with a quota request for every call. That part is similar to a check request, it is always called.
This argues for selectively enabling mixer calls from the proxy. For example if we know that mixer config does not have any "check" aspects configured, proxy should skip the call.

ZackButcher: Internally we have a static map of `method -> quota bucket` so that the proxy knows what to check against. I don't see why we couldn't use that model.

Perhaps more naturally in mixer-land, we could allow the user to write a set of expressions we evaluate to determine the quota bucket. Then we remove the quota bucket name from the `Quota` API's request object. I like this model more, personally.

douglas-reid: yeah, i feel like it is a bit strange to require config updates on both proxy side and mixer side to get a quota to be effective.

even if the proxy always calls Quota(), it seems like it shouldn't have to specify the name of the quota that applies (and that mixer should be able to determine it at request time).

Otherwise, is the plan to have mixer config updates propagate to proxy on every quota override?

mandarjog: Updating quota values happens completely thru mixer. 
Changing the set of quotas that apply will require a proxy config change.

This is a regression in function when we moved away from check based quota.
Now mixer api server will need to call proxy config api server to co ordinate this change.  

A co-ordination of this kind was always in the plan for versioned config rollout, we will need to move it up now.

Is it possible attach the quota aspect back to the `check` api method?

douglas-reid: I feel like I would vote for moving back to `Check`-based quota enforcement until we have this configuration story figured out. It seems really really odd to go through adding the quota descriptor bits into a scope/subject rule and then not have that work because I need to do the same thing somewhere else (even within the context of alpha). The value of having in that scope/subject config seems completely lost.

For that matter, I'm still confused as to why `RequestCount` is set to 5 QPS globally (and is not configured to 5 QPS inside a selector config), but perhaps that is another matter.

ZackButcher: So due to code changes in mixer it's no longer trivial to move back to `Check` based quota; we could write some shims to do it, but it's not as easy as adding the quota aspect manager into the check manager set unfortunately.

geeknoid: Sorry to come in late to the discussion...

The proxy must know which quota to request from the mixer by name, and how much to request. This is the only way that prefetching of quota can possibly work. In a fully-realized solution, the proxy will need to do a bit of processing to establish, for example, which method is being called in order to query the right quota bucket.

I really tried to make it so the proxy doesn't need to say which quota to get and how much, but it quickly becomes extremely complicated to handle caching/prefetching. So I abandoned that path.

In terms of calling for Quota on every incoming request, we are crafting a pretty nice prefetching algorithm which will generally hide any quota-related latency. Upon first request from a client, the proxy will issue a request for N tokens into the future and will use up those tokens. Before the tokens drop to 0, the proxy will make another request so that it hopefully always has a supply of tokens on hand. But in general, there will never be blocking on quota requests.

douglas-reid: Here's my alpha proposal:

- hardcode in a name here (maybe change it from `RequestCount` to `RateLimit`).
- add a field on the mixer quota aspect config called `limit` which allows specification of a limit that should be applied within that scope/subject
- remove `max_amount` from the descriptor of `RequestCount`/`RateLimit` (no sense in setting an upper bound on rate limits at this point)
- update the quota manager code to use `limit` instead of `max_amount`
- set `limit: 5` in the aspect config for the selector of "ratings*"

That should allow mixerclient to call `Quota()` on all requests, but have mixer just return `status.OK` unless the quota aspect is configured and linked to the `RateLimit` descriptor for a scope/subject.  And it should allow *different* limits based on scope/subject.

Thoughts?

ZackButcher:
This change is correct, it's most of what we need to get rate limiting working. After this change rate limiting will still not work, but because of a bug in the mixer (the mixer filters the manager sets in the proxy do not always set `target.service`; when it's missing we fail to resolve our config and return an internal error).

proxy/envoy/resources.go
```diff
@@ -131,6 +131,8 @@ type FilterMixerConfig struct {
 	// ForwardAttributes specifies the list of attribute keys and values that
 	// are forwarded as an HTTP header to the server side proxy
 	ForwardAttributes map[string]string `json:"forward_attributes,omitempty"`
+
+	QuotaName string `json:"quota_name,omitempty"`
```
douglas-reid: @mandarjog @geeknoid should this be []string, based on the current approach for quota?

ZackButcher: That requires changes to mixerclient first; lets get this change in to unblock rate limiting and we can circle back to fix up how quotas are actually checked.

proxy/envoy/policy.go
douglas-reid: filed: istio/mixer#574

that should allow moving forward from here.

kyessenov:
closing in favor of #555

PR #530, opened 2017-04-14
Add envoy config test run

kyessenov:
Bazel test to spawn envoy with golden config, wait a second, and kill it.

### DIFF
--- a/proxy/envoy/config_test.sh
+++ b/proxy/envoy/config_test.sh
@@ -5,4 +5,4 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-proxy/envoy/envoy -l trace -c proxy/envoy/testdata/envoy-v0.json.golden --service-cluster istio-proxy --service-node 10.0.0.1
+#proxy/envoy/envoy -l trace -c proxy/envoy/testdata/envoy-v0.json.golden --service-cluster istio-proxy --service-node 10.0.0.1
--- a/proxy/envoy/config_test.sh
+++ b/proxy/envoy/config_test.sh
@@ -11,9 +11,6 @@ CONFIG_FILES="envoy-v0 envoy-fault ingress-envoy"
 
 # TODO the test skips over SSL configuration files since referenced cert files are not in the right places
 
-ls /dev/shm/
-touch /dev/shm/test
-
 for f in ${CONFIG_FILES}; do
   echo "##############################"
   echo "# Testing $f..."
--- a/proxy/envoy/config_test.sh
+++ b/proxy/envoy/config_test.sh
@@ -1,8 +1,25 @@
 #!/bin/bash
 # Check that envoy can parse golden configuration artifacts
+# In the future, we can run functional tests, but all configs require iptables rules to be effective.
 
 set -o errexit
 set -o nounset
 set -o pipefail
 
-#proxy/envoy/envoy -l trace -c proxy/envoy/testdata/envoy-v0.json.golden --service-cluster istio-proxy --service-node 10.0.0.1
+ENVOY="proxy/envoy/envoy -l trace --service-cluster istio-proxy --service-node 10.0.0.1"
+CONFIG_FILES="envoy-v0 envoy-fault ingress-envoy"
+
+# TODO the test skips over SSL configuration files since referenced cert files are not in the right places
+
+ls /dev/shm/
+touch /dev/shm/test
+
+for f in ${CONFIG_FILES}; do
+  echo "##############################"
+  echo "# Testing $f..."
+  echo "##############################"
+  ${ENVOY} -c proxy/envoy/testdata/$f.json.golden &
+  PID=$!
+  sleep 1
+  kill -KILL $PID
+done
--- a/proxy/envoy/config_test.sh
+++ b/proxy/envoy/config_test.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+# Check that envoy can parse golden configuration artifacts
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+proxy/envoy/envoy -l trace -c proxy/envoy/testdata/envoy-v0.json.golden --service-cluster istio-proxy --service-node 10.0.0.1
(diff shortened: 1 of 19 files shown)


proxy/envoy/config_test.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# TODO the test skips over SSL configuration files since referenced cert files are not in the right places
+
+for f in ${CONFIG_FILES}; do
+  echo "##############################"
+  echo "# Testing $f..."
+  echo "##############################"
+  ${ENVOY} -c proxy/envoy/testdata/$f.json.golden &
+  PID=$!
+  sleep 1
```
ayj: This sleep is non-deterministic. Can this be replaced with a pseudo-health check, perhaps on one of the admin endpoints?

proxy/envoy/config_test.sh
```diff
@@ -0,0 +1,22 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ENVOY="proxy/envoy/envoy -l trace --service-cluster istio-proxy --service-node 10.0.0.1"
+CONFIG_FILES="envoy-v0 envoy-fault ingress-envoy egress-envoy"
+
+# TODO the test skips over SSL configuration files since referenced cert files are not in the right places
+
+for f in ${CONFIG_FILES}; do
```
ayj: Are bazel tests sandboxed re: ports? If so, can this be restructured as separate `sh_test` targets for each configuration? It should be functionally equivalent and bazel would then be able to handle the logging natively for test success/failure.

kyessenov: Yes, but `sh_test` does not take a parameter, so we would be writing in skylark, which is harder to debug. The longest test takes 12 seconds, this one is exactly 4.

proxy/envoy/config_test.sh
kyessenov: Envoy is asynchronous and I deliberately want to wait till Envoy completes a round of *DS fetches. Admin port might pop up before the round starts.

kyessenov: E.g. this uncovered a hard segfault in c-ares if I leave "manager:8080" as the discovery address.

proxy/envoy/config_test.sh
kyessenov: tried to split but sh_test is not network namespaced. the whole net namespace isolation is a mystery since in some versions of bazel it works, in others it doesn't...

ayj:
I'm getting some libstdc++ errors when I try to run this on my workstation (see https://gist.github.com/ayj/d011e3b91b6851e281b758e013ad1bff). Any ideas?

kyessenov:
Are you on ubuntu? Try `sudo apt-get install libstdc++6`?

ayj:
Goobuntu 

```bash
$ sudo apt-get install libstdc++6
Reading package lists... Done
Building dependency tree
Reading state information... Done
libstdc++6 is already the newest version (4.8.4-2ubuntu1~14.04.3).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

kyessenov:
Hmm, goobuntu is sometimes acting weird. @lizan or @qiwzhang why doesn't envoy binary run on goobuntu?

proxy/envoy/config_test.sh
ayj: Thanks for checking re: namespaces. It looks only user namespaces are supported for now. The `exclusive` tag would solve the same problem and avoid running these tests with any other bazel test via serialization (see https://bazel.build/versions/master/docs/test-encyclopedia.html#tag-conventions).

sh_test supports `args` through the attributes common to all test rules (
https://bazel.build/versions/master/docs/be/common-definitions.html#common-attributes-tests).

With the above two changes the tests would look something like the following:

```bash
sh_test(
    name = "config_envoy_fault",
    size = "small",
    srcs = ["config_test.sh"],
    data = ["testdata/envoy-fault.json.golden"],
    args = ["testdata/envoy-fault.json.golden"],
    tags = ["exclusive"],
)
```

And then test output would look like this:

```bash
$ bazel test //proxy/envoy:config_envoy_v0 //proxy/envoy:config_envoy_fault
INFO: Found 2 test targets...
INFO: Elapsed time: 0.224s, Critical Path: 0.03s
//proxy/envoy:config_envoy_fault                                PASSED in 1.0s
//proxy/envoy:config_envoy_v0                                   PASSED in 1.0s

Executed 0 out of 2 tests: 2 tests pass.
```

proxy/envoy/config_test.sh
ayj: > ... wait till Envoy completes a round of *DS fetches

In that case, what about curl'ing the `localhost:<admin-port>/stats` and grepping for specific SDS/CDS/RDS stats to verify envoy has populated them correctly after polling discovery service? Envoy's discovery interval can be 2 * refresh_delay_ms, e.g. 2 seconds, in which case this test wouldn't sufficiently verify that one round of SDS/CDS/RDS completed successfully.

proxy/envoy/config_test.sh
kyessenov: Yeah, exclusive would work. But is the increased bazel complexity worth it? The only gain is shorted logs, and the negative is necessity to explain what all that bazel magic does :)

proxy/envoy/config_test.sh
kyessenov: Good point. I'll see what can be done about this. I wish envoy would just have a flag to verify config schema.

proxy/envoy/config_test.sh
ayj: I think there is some value in doing things the bazel way here. There are many other useful bazel test features we can start incorporating into our tests, but it's hard to do that if we push all the test configuration into test scripts when equivalent features are available in bazel. 

For example, the following options are available at the per-test level with bazel:

- Add the `flaky` attribute and bazel will re-run the test up to three times.
- Add the `manual` tag to exclude tests from wildcard patterns. This is useful if we need to temporarily exclude a test from presubmit without modifying the test code itself. Bonus is the test is still available via `bazel test`.
- The `timeout` attribute can used in this test to avoid polling for SDS/CDC/RDS stats indefinitely and is overridable via `--test_timeout` flag.

lizan:
I guess it is just because the libstdc++6 in goobuntu is too old, is the binary built on same machine?

PR #534, opened 2017-04-14
Split envoy config test

kyessenov:


proxy/envoy/config_test.sh
```diff
@@ -2,21 +2,20 @@
...
-for f in ${CONFIG_FILES}; do
-  echo "##############################"
-  echo "# Testing $f..."
-  echo "##############################"
-  ${ENVOY} -c proxy/envoy/testdata/$f.json.golden &
-  PID=$!
-  sleep 1
-  kill -KILL $PID
-done
+echo "##############################"
```
ayj: I don't think these three echo lines are needed anymore?

proxy/envoy/config_test.go
```diff
@@ -312,6 +314,7 @@ func makeMeshConfig() proxyconfig.ProxyMeshConfig {
 	mesh := DefaultMeshConfig
 	mesh.MixerAddress = "localhost:9091"
 	mesh.DiscoveryAddress = "localhost:8080"
+	mesh.DiscoveryRefreshDelay = &duration.Duration{Nanos: int32(10 * time.Millisecond / time.Nanosecond)}
```
ayj: You can use the ptypes package to help here, e.g.

```golang
import "github.com/golang/protobuf/ptypes/ptypes"

mesh.DiscoveryRefreshDelay = ptypes.DurationProto(10 * time.Millisecond)
```

proxy/envoy/BUILD
```diff
@@ -59,8 +59,65 @@ genrule(
 )
 
 sh_test(
-    name = "config_test",
+    name = "envoy_config_v0_test",
```
ayj: Do these need a dependency on `envoy_bin` or does bazel figure that out via `data` attribute?

kyessenov: "envoy" data dependency is sufficient for genrules.

proxy/envoy/config_test.go
kyessenov: ack, that's better.

proxy/envoy/config_test.sh
kyessenov: removing them

PR #540, opened 2017-04-14
Improve ingress failure handling

ijsnellf:
Modifies the ingress Envoy config generation logic to avoid generating invalid Envoy configurations. This should prevent the hot restart issue described here: https://github.com/istio/manager/issues/508. However, this change does mean that failures will only be logged and no retries will be attempted.


ijsnellf:
@kyessenov: Yes, that is correct. This will just improve the behavior when it is flaky. The secret controller will be a separate PR.

PR #541, opened 2017-04-14
Egress support for https external services

GregHanson:
- as it is now, clients will have to call `http://httpbin:443` vs `http://httpbin` for external services of type https
 - adds empty ssl_context block to cluster definition in egress service if type https


platform/kube/conversion_test.go
```diff
@@ -148,7 +148,7 @@ func TestServiceConversion(t *testing.T) {
 			len(service.Ports), len(localSvc.Spec.Ports))
 	}
 
-	if service.ExternalName != "" {
+	if service.ExternalName != "" || service.External() {
```
ijsnellf: Should just be `service.External()`, since
```go
// External predicate checks whether the service is external
func (s *Service) External() bool {
	return s.ExternalName != ""
}
```

GregHanson: fixed

proxy/envoy/config.go
```diff
@@ -300,7 +300,31 @@ func buildOutboundHTTPRoutes(
 				http.VirtualHosts = append(http.VirtualHosts, host)
 				clusters = append(clusters, host.clusters()...)
 
-			case model.ProtocolTCP, model.ProtocolHTTPS:
+			case model.ProtocolHTTPS:
+				if service.External() {
+					routes := make([]*HTTPRoute, 0)
+
+					cluster := buildOutboundCluster(service.Hostname, servicePort, nil)
```
kyessenov: Can you make this a function `useEgressCluster(route)` on a default route and reduce code duplication with the above?

proxy/envoy/resources.go
```diff
@@ -410,10 +410,10 @@ type SSLContext struct {
 
 // SSLContextWithSAN definition, VerifySubjectAltName cannot be nil.
 type SSLContextWithSAN struct {
-	CertChainFile        string   `json:"cert_chain_file"`
-	PrivateKeyFile       string   `json:"private_key_file"`
+	CertChainFile        string   `json:"cert_chain_file,omitempty"`
+	PrivateKeyFile       string   `json:"private_key_file,omitempty"`
 	CaCertFile           string   `json:"ca_cert_file,omitempty"`
```
kyessenov: Do you know what root CAs is Envoy using? Are they packaged with the binary or do we need to add them to the base image?

proxy/envoy/testdata/all-cds.json.golden
```diff
@@ -74,6 +85,17 @@
      ]
     },
     {
+     "name": "out.httpsbin.default.svc.cluster.local|https",
```
kyessenov: curious, why do we have so many copies of this cluster? is `all-cds` computing clusters for every node?

proxy/envoy/testdata/cds-ssl-context.json.golden
```diff
@@ -9,8 +9,7 @@
     "ssl_context": {
      "cert_chain_file": "/etc/certs/cert-chain.pem",
      "private_key_file": "/etc/certs/key.pem",
-     "ca_cert_file": "/etc/certs/root-cert.pem",
-     "verify_subject_alt_name": []
```
kyessenov: @myidpt does this change semantics of the cluster? If I remember we want this to be an empty array rather than dropped since the two mean different things: missing implies all allowed, empty implies none allowed. We might need to define another schema resource to deal with this problem.

kyessenov:
#67

proxy/envoy/resources.go
rshriram: Needs to be added

proxy/envoy/testdata/all-cds.json.golden
GregHanson: That was my understanding of the get-all RDS and CDS API's

proxy/envoy/testdata/cds-ssl-context.json.golden
myidpt: Yes, you are right, Kuat.
My question is, is this cluster supposed to point to egress? If so, the egress should use Istio auth when talking to internal proxes, right? Then verify_subject_alt_name is required.

GregHanson:
@kyessenov do you want the documentation in this PR? Or should I just add it to the istio.io documentation?

kyessenov:
I think we want a quick task in istio.io for using HTTP and HTTPS external name services. The tests we have are ok for internal dev docs. Code cleanup and comments can follow up once we start integrating ingress/egress/sidecar config generation.

PR #550, opened 2017-04-17
Fetch Istio auth secret from Ingress proxy when mutual TLS is enabled

lookuptable:
This is part 1 for enabling Istio Auth between the Ingress and other
proxies.

### DIFF
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -326,19 +326,6 @@ func buildOutboundHTTPRoutes(
 		}
 
 		clusters.setTimeout(mesh.ConnectTimeout)
-
-		// apply SSL context to outbound clusters for authentication policy
-		switch mesh.AuthPolicy {
-		case proxyconfig.ProxyMeshConfig_NONE:
-		case proxyconfig.ProxyMeshConfig_MUTUAL_TLS:
-			serviceAccounts := accounts.GetIstioServiceAccounts(service.Hostname, service.Ports.GetNames())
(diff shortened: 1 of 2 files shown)


proxy/envoy/ingress.go
```diff
@@ -372,3 +387,20 @@ func extractPortAndTags(dst *config.DestinationWeight) (*model.Port, model.Tags,
 
 	return port, tags, nil
 }
+
+func writeIstioAuthSecret(secret *model.IstioAuthSecret, mesh *config.ProxyMeshConfig) error {
```
wattli: Add some comments to make it more readable?

wattli: Looks like we are manually fetch and write the secrets to dedicated path, just like Proxy, right?

model/auth.go
```diff
@@ -0,0 +1,39 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+const (
+	// IstioAuthCACertificate defines the name for CA root certificate file.
+	IstioAuthCACertificate = "root-cert.pem"
```
kyessenov: quick comment: can we use a more descriptive name for the in-cluster mutual TLS rather than "Istio Auth". Let's drop Istio to reduce stutter (all code has istio.io in the name already), and clarify which "auth" this refers to.

PR #552, opened 2017-04-17
Testing: add checks for duplicate request IDs in the access logs

kyessenov:


test/integration/access.go
```diff
@@ -0,0 +1,113 @@
...
+
+// add an access log entry for an app
+func (a *accessLogs) add(app, id, desc string) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	v, exists := a.logs[app]
+	if !exists {
+		v = make([]request, 0)
+	}
+	a.logs[app] = append(v, request{id: id, desc: desc})
```
ayj: You can append to a nil slice and avoid the extra conditional check here, e.g. 
```golang
a.logs[app] = append(a.logs[app], request{id: id, desc: desc})
```

test/integration/access.go
```diff
@@ -0,0 +1,113 @@
...
+
+				if strings.Contains(logs, "assert failure") {
+					glog.Errorf("assert failure in %s", pod)
+					return failure
+				}
+
+				// find all ids and counts
+				counts := make(map[string]int)
+				for _, request := range a.logs[app] {
+					i, exists := counts[request.id]
```
ayj: You can take advantage of the fact that non-existent keys have the value type's zero value, e.g. 

```golang
counts[request.id] = counts[request.id] + 1
```

test/integration/access.go
```diff
@@ -0,0 +1,113 @@
...
+					if !exists {
+						i = 0
+					}
+					counts[request.id] = i + 1
+				}
+				for id, count := range counts {
+					i := strings.Count(logs, id)
```
ayj: `i` isn't very meaningful here. Maybe use got/want instead? e.g.

```golang
if got := strings.Count(logs.id), got < want {
}
```

test/integration/access.go
kyessenov: done

test/integration/access.go
kyessenov: done

test/integration/access.go
kyessenov: done

kyessenov:
#533

PR #556, opened 2017-04-18
HTTP/2 (GRPC) tests

kyessenov:
More work on test framework and add reachability tests for GRPC.
Extend test client and server with an echo GRPC service.

#533

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -2,7 +2,7 @@ workspace(name = "com_github_istio_manager")
 
 git_repository(
     name = "io_bazel_rules_go",
-    commit = "87cdda3fc0fd65c63ef0316533be03ea4956f809",  # April 7 2017 (0.4.2)
+    commit = "78d030fc16e7c6e0a188714980db0b04086c4a5e",  # April 12 2017 (0.4.3)
     remote = "https://github.com/bazelbuild/rules_go.git",
 )
 
@@ -108,12 +108,6 @@ new_go_repository(
     name = "com_github_golang_groupcache",
     commit = "02826c3e79038b59d737d3b1c0a1d937f71a4433",
     importpath = "github.com/golang/groupcache",
-)
-
-new_go_repository(
-    name = "com_github_golang_protobuf",
-    commit = "8616e8ee5e20a1704615e6c8d7afcdac06087a67",
-    importpath = "github.com/golang/protobuf",
 )
 
 new_go_repository(
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -341,8 +341,8 @@ http_archive(
 
 new_go_repository(
     name = "org_golang_google_grpc",
+    commit = "8050b9cbc271307e5a716a9d782803d09b0d6f2d", # v1.2.1
     importpath = "google.golang.org/grpc",
-    tag = "v1.2.1",
 )
 
 new_git_repository(
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -324,9 +324,26 @@ http_file(
 ## Protobuf codegen rules
 ##
 
-load("@io_bazel_rules_go//proto:go_proto_library.bzl", "go_proto_repositories")
+# Note: do not use go_proto_repositories since it has old versions of protobuf and grpc-go
 
-go_proto_repositories()
+new_go_repository(
+    name = "com_github_golang_protobuf",
+    commit = "8ee79997227bf9b34611aee7946ae64735e6fd93",
+    importpath = "github.com/golang/protobuf",
+)
+
+http_archive(
+    name = "com_github_google_protobuf",
+    sha256 = "2a25c2b71c707c5552ec9afdfb22532a93a339e1ca5d38f163fe4107af08c54c",
+    strip_prefix = "protobuf-3.2.0",
+    url = "https://github.com/google/protobuf/archive/v3.2.0.tar.gz",
+)
+
+new_go_repository(
+    name = "org_golang_google_grpc",
+    importpath = "google.golang.org/grpc",
+    tag = "v1.2.1",
+)
 
 new_git_repository(
     name = "io_istio_api",
--- a/test/client/client.go
+++ b/test/client/client.go
@@ -19,23 +19,24 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
 	"strings"
 	"time"
 
+	"google.golang.org/grpc"
+
 	"crypto/tls"
 
 	"github.com/golang/sync/errgroup"
+	pb "istio.io/manager/test/grpcecho"
 )
 
 var (
-	count    int
-	parallel bool
-	insecure bool
-	timeout  time.Duration
-	client   *http.Client
+	count   int
+	timeout time.Duration
 
 	url       string
 	headerKey string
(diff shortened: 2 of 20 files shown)


test/client/client.go
```diff
@@ -44,83 +46,124 @@ var (
...
 }
 
-func makeRequest(i int) func() error {
-	return func() error {
-		req, err := http.NewRequest("GET", url, nil)
-		if err != nil {
-			return err
-		}
+func makeHTTPRequest(client *http.Client) func(int) func() error {
+	return func(i int) func() error {
```
andraxylia: Style comment: I noticed all over the code we over use these inline func. This is fine as long as the func is small, but this one is big. Also, there are 2 levels of "return func".

kyessenov: errgroup forces me use continuation passing style here. do we really need names for these intermediate functions?

kyessenov: I wouldn't call CPS an overuse :) golang has true functions and it's actually easier to parallelize CPS-style funcs

test/client/client.go
```diff
@@ -19,23 +19,25 @@ package main
...
+	"os"
 	"strings"
 	"time"
 
+	"google.golang.org/grpc"
+
 	"crypto/tls"
 
 	"github.com/golang/sync/errgroup"
+	pb "istio.io/manager/test/grpcecho"
```
andraxylia: Avoid renaming and name the package "grpc_echo"

ayj: Underscores are discouraged in package names (see https://blog.golang.org/package-names)

kyessenov: e.g. `strconv`

test/client/client.go
```diff
@@ -44,83 +46,124 @@ var (
...
+			// grpc-go sets incorrect authority header
+			grpc.WithAuthority(address),
+			grpc.WithBlock(),
+			grpc.WithTimeout(timeout))
+		if err != nil {
+			log.Fatalf("did not connect: %v", err)
+		}
+		client := pb.NewEchoTestServiceClient(conn)
+		f = makeGRPCRequest(client)
+		defer func() {
```
andraxylia: Move it up, right after conn is open.

test/client/client.go
```diff
@@ -44,83 +46,124 @@ var (
...
-		err := g.Wait()
-		if err != nil {
-			log.Printf("Error %s\n", err)
-		} else {
-			log.Println("All requests succeeded")
-		}
+	if err := g.Wait(); err != nil {
+		log.Printf("Error %s\n", err)
+		os.Exit(1)
+	} else {
```
andraxylia: No need for else.

test/integration/grpc.go
```diff
@@ -0,0 +1,92 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+
```
andraxylia: Remove extra line.

test/client/client.go
ayj: I've found the nested inline functions a bit difficult to parse at times. It might be easier to read if  makeXXXRequest are defined as a normal function that takes http.Client and the iteration count and use that directly (via function pointer).

```golang
func httpRequest(client *http.Client, iter count) error {
    req, err ;= http.NewRequest("GET", url, nil)
    log.Printf("[%d] Url=%s\n", i, url)
}

var request func(client*http.Client, iter int) error
switch {
case http:
    request = httpRequest 
case grpc:
    request = grpcRequest
default: // error 
}

for i := 0; i < count; i++ {
 	g.Go(request(client, i))
}  
```

ayj: Deleted my original comment because it was wrong. But I agree that this is a bit difficult to parse with the two levels of inline functions, though easier than the equivalent in C :)

Would it work if the XXXRequest functions were normal functions and the code at 161 simplified wrapped that in a function, e.g.
 
```golang
g.Go(func() { 
    makeHTTPRequest(client, iter) 
})
```

kyessenov: Go doesn't capture `for` loop variables for efficiency.

test/client/client.go
```diff
@@ -44,83 +46,124 @@ var (
...
-			if line != "" {
-				log.Printf("[%d body] %s\n", i, line)
+func makeGRPCRequest(client pb.EchoTestServiceClient) func(int) func() error {
+	return func(i int) func() error {
+		return func() error {
+			req := &pb.EchoRequest{Message: fmt.Sprintf("request #%d", i)}
+			log.Printf("[%d] grpcecho.Echo(%v)\n", i, req)
+			resp, err := client.Echo(context.Background(), req)
+			if err != nil {
+				log.Printf("[%d error] %v\n", i, err)
```
ayj: Anyone to return *and* log the error instead of just returning it?

test/client/client.go
ayj: "else" isn't need here since the previous error check will exit.

kyessenov: yeah, linter missed this.

test/client/client.go
kyessenov: yeah, no need to log the error since it's logged in main.

test/client/client.go
kyessenov: ack

WORKSPACE
```diff
@@ -330,9 +324,26 @@ http_file(
...
+    sha256 = "2a25c2b71c707c5552ec9afdfb22532a93a339e1ca5d38f163fe4107af08c54c",
+    strip_prefix = "protobuf-3.2.0",
+    url = "https://github.com/google/protobuf/archive/v3.2.0.tar.gz",
+)
 
-go_proto_repositories()
+new_go_repository(
+    name = "org_golang_google_grpc",
+    importpath = "google.golang.org/grpc",
+    tag = "v1.2.1",
```
ayj: Why the switch to "tag" instead of "commit' here? Weren't we using "commit" even if cases where tag was available because tag wasn't 100% immutable?

test/integration/driver.go
```diff
@@ -156,6 +156,7 @@ func runTests(envs ...infra) {
 
 		tests := []test{
 			&reachability{infra: &istio},
+			&grpc{infra: &istio},
```
ayj: Does the order matter here?

test/integration/infra.go
```diff
@@ -177,3 +184,42 @@ func (infra *infra) teardown() {
...
+func (infra *infra) clientRequest(app, url string, count int, extra string) response {
+	request, err := util.Shell(fmt.Sprintf("kubectl exec %s -n %s -c app -- client -url %s -count %d %s",
+		infra.apps[app][0], infra.Namespace, url, count, extra))
+	out := response{}
+	if err != nil {
+		glog.Errorf("client request error %v for %s in %s", err, url, app)
+		return out
+	}
+
+	ids := regexp.MustCompile("(?i)X-Request-Id=(.*)").FindAllStringSubmatch(request, -1)
```
ayj: clientRequest is called many times from nested loops. It would be good to move MustCompile out to a global to avoid recompiling regex each time.

rshriram:
Why remove the insecure option? I recall that ingress was using it. I haven't seen the code, so please ignore if you have added support for the same in another form. 

We also need insecure for egress right?

test/client/client.go
andraxylia: Ack. It also discourage renaming imports : https://github.com/golang/go/wiki/CodeReviewComments#imports

test/client/client.go
andraxylia: I will not make this a blocking comment if you cannot do it differently, but I do find it hard to read. Besides, it increases cyclomatic complexity.

test/client/client.go
kyessenov: copy paste error from grpc-go tutorial... I don't mind keeping the original name.

test/integration/driver.go
kyessenov: Each test should be self contained, so order is not important.

test/integration/infra.go
kyessenov: ack

WORKSPACE
kyessenov: was experimenting with it plays with rules_go, should use commit.

kyessenov:
`insecure` is always on (like curl).

test/integration/grpc.go
ijsnellf: I think that is goimports fault.

kyessenov:
PTAL

PR #557, opened 2017-04-19
Limit Istio CA to issue certs in its own namespace in integration test

lookuptable:


test/integration/testdata/ca.yaml.tmpl
```diff
@@ -14,3 +14,9 @@ spec:
       - name: istio-ca-container
         image: {{.CaImage}}
         imagePullPolicy: Always
+	env:
```
kyessenov: this is not valid YAML

PR #560, opened 2017-04-20
istioctl: add support for creating mixer rules

ayj:
This change adds support for creating mixer rules with istioctl. It
supports "create" and "get" as subcommands to "istioctl mixer
rule". The mixer API server address is set with the
ISTIO_MIXER_API_SERVER environment variable for easy integration into
istio demo and smoke test. The server address can also be set with "-m/--mixer" flag.

Example usage:

    $ export ISTIO_MIXER_API_SERVER=localhost:9094
    $ cat rules.yml
    subject: namespace:ns
    revision: "2022"
    rules:
    - aspects:
      - kind: denials

    $ istioctl mixer rule create global myservice.ns.svc.cluster.local -f rules.yml
    $ istioctl mixer rule get global myservice.ns.svc.cluster.local
    revision: "2022"
    rules:
    - aspects:
      - kind: denials
    subject: namespace:ns

There is much room for improvement in the istioctl and API server space which is not addressed in this PR:

1) Unify mixer and proxy API servers

2) Unify proxy and mixer resource file format so that
"istioctl create -f <file>" can be used to create any
istio resource, e.g. DestinationPolicy, RouteRule, MixerRule.

3) Move API server's gRPC and REST API definitions into istio.io/api
instead of spreading it across multiple component repos.

4) istioctl auth to API servers


cmd/istioctl/mixer.go
```diff
@@ -0,0 +1,166 @@
...
+
+	mixerRuleCmd = &cobra.Command{
+		Use:          "rule",
+		Short:        "Istio Mixer Rule configuration",
+		SilenceUsage: true,
+	}
+
+	mixerRuleCreateCmd = &cobra.Command{
+		Use:   "create <scope> <subject>",
+		Short: "Create an Istio Mixer rule",
```
mandarjog: Create a Istio Mixer rules document.

This document has multiple rules within in.

cmd/istioctl/mixer.go
```diff
@@ -0,0 +1,166 @@
...
+
+func mixerRulePath(host, scope, subject string) string {
+	if !strings.HasPrefix(host, "http://") {
+		host = "http://" + host
+	}
+	return fmt.Sprintf("%s/api/v1/scopes/%s/subjects/%s/rules", host, scope, subject)
+}
+
+func mixerRuleCreate(host, scope, subject string, rule []byte) error {
+	fmt.Println(mixerRulePath(host, scope, subject))
```
mandarjog: is this `fmt` on purpose here?

cmd/istioctl/mixer.go
```diff
@@ -0,0 +1,166 @@
...
+	request, err := http.NewRequest("PUT", mixerRulePath(host, scope, subject), bytes.NewReader(rule))
```
mandarjog: use `http.MethodPut`

cmd/istioctl/mixer.go
ayj: Nope, was leftover from debug - removed.

cmd/istioctl/mixer.go
ayj: Done

cmd/istioctl/mixer.go
```diff
@@ -0,0 +1,165 @@
...
+
+func mixerRulePath(host, scope, subject string) string {
+	if !strings.HasPrefix(host, "http://") {
+		host = "http://" + host
+	}
+	return fmt.Sprintf("%s/api/v1/scopes/%s/subjects/%s/rules", host, scope, subject)
+}
+
+func mixerRuleCreate(host, scope, subject string, rule []byte) error {
+	request, err := http.NewRequest(http.MethodPut, mixerRulePath(host, scope, subject), bytes.NewReader(rule))
```
kyessenov: the benefit of using `NewRequest` is that you can attach cancellation context...

ayj: Is there a better way to send a `put` request, or were you suggesting adding a cancellation context?  http.Client doesn't have a `Put` method.

bin/init.sh
```diff
@@ -17,7 +17,10 @@ ln -sf "$(pwd)/bazel-genfiles/external/io_istio_api/proxy/v1/config/cfg.pb.go" \
   vendor/istio.io/api/proxy/v1/config/cfg.pb.go
 mkdir -p vendor/istio.io/manager/test/grpcecho
 ln -sf "$(pwd)/bazel-genfiles/test/grpcecho/echo.pb.go" \
-  vendor/istio.io/manager/test/grpcecho/
+   vendor/istio.io/manager/test/grpcecho/
+mkdir -p vendor/github.com/googleapis/googleapis/google/rpc
+ln -sf $(pwd)/bazel-genfiles/external/com_github_googleapis_googleapis/google/rpc/*.pb.go \
```
kyessenov: can you also make sure `bin/gazelle` works with this change and do a fix if it doesn't?

cmd/istioctl/mixer.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return fmt.Errorf("failed creating request: %v", err)
+	}
+	request.Header.Set("Content-Type", "application/yaml")
+
+	var client http.Client
+	resp, err := client.Do(request)
```
kyessenov: set request timeouts throughout. default timeout is infinite.

cmd/istioctl/mixer.go
```diff
@@ -0,0 +1,164 @@
...
+		return "", fmt.Errorf("failed sending request: %v", err)
+	}
+	defer resp.Body.Close() // nolint: errcheck
+
+	if resp.StatusCode != http.StatusOK {
+		return "", errors.New(http.StatusText(resp.StatusCode))
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("failed processing response: %v", err)
```
kyessenov: nit: maybe wrap into a function and use multiline.Prefix?

bin/init.sh
ayj: Missed this - thanks

cmd/istioctl/mixer.go
ayj: I don't quite understand. Are you suggesting regex to match/extract response body contents?

ayj:
Anyone other comments or is this gtg?

cmd/istioctl/mixer.go
```diff
@@ -0,0 +1,170 @@
...
+
+// TODO This should come from something like istio.io/api instead of
+// being hand copied from istio.io/mixer.
+type mixerAPIResponse struct {
+	Data   interface{} `json:"data,omitempty"`
+	Status rpc.Status  `json:"status,omitempty"`
+}
+
+const (
+	requestTimeout = time.Second
```
kyessenov: 60 seconds?

cmd/istioctl/mixer.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	return fmt.Sprintf("%s/api/v1/scopes/%s/subjects/%s/rules", host, scope, subject)
+}
+
+func mixerRuleCreate(host, scope, subject string, rule []byte) error {
+	request, err := http.NewRequest(http.MethodPut, mixerRulePath(host, scope, subject), bytes.NewReader(rule))
+	if err != nil {
+		return fmt.Errorf("failed creating request: %v", err)
+	}
+	request.Header.Set("Content-Type", "application/yaml")
```
kyessenov: Is YAML a good format for client-server exchange? I guess it's to make error messages more useful?

kyessenov:
Let's merge this. I suspect we'll need to do several iterations to make it usable, but it's hard to tell now without a working rate-limit or smth else.

cmd/istioctl/mixer.go
kyessenov: Just want to avoid repetitions of `fmt.Errorf` with same prefix...

cmd/istioctl/mixer.go
kyessenov: ah, I misunderstood the original comment. There is `http.Put` method that we should avoid.

cmd/istioctl/mixer.go
ayj: Probably not, but the mixer API server accepts it and I wanted to pass the rule content through transparently with minimal client processing.

PR #562, opened 2017-04-20
Update ingress resource status

ijsnellf:
Resolves https://github.com/istio/manager/issues/214.


ijsnellf:
Once this merges, I'll update the docs to make use of it.

ijsnellf:
@rshriram: Yep.

andraxylia:
In addition, this Ingress IP is the standard way users should be accessing the HTTP LB and the bookinfo /productpage.
@ijsnellf I will update the bookinfo, I have a bunch of other changes to commit, but please feel free to update any other ingress related doc.

platform/kube/controller.go
```diff
@@ -131,6 +134,25 @@ func NewController(client *Client, options ControllerOptions) *Controller {
...
+		// 2. specific ingress class provided and same annotation on ingress resource
+		var ingressClass, defaultIngressClass string
+		switch options.IngressSyncMode {
+		case IngressStrict:
+			ingressClass = options.IngressClass
+		case IngressDefault:
+			defaultIngressClass = options.IngressClass
+		}
+
+		out.ingressStatus = status.NewStatusSyncer(status.Config{
```
kyessenov: Seems like the syncer is going to use `POD_IP` to set the IP in the status. Should we be using the ingress service instead? There is an option `PublishServicer' in the config here.

kyessenov: Actually, this is probably even necessary since the ingress controller is run by the central manager, not the ingress proxy.

platform/kube/controller.go
```diff
@@ -131,6 +134,25 @@ func NewController(client *Client, options ControllerOptions) *Controller {
...
+
+		// Ingress sync mode -> ingress status sync mode. The ingress status is synced in two cases:
+		// 1. default ingress class provided and no annotation on ingress resource
+		// 2. specific ingress class provided and same annotation on ingress resource
+		var ingressClass, defaultIngressClass string
+		switch options.IngressSyncMode {
+		case IngressStrict:
+			ingressClass = options.IngressClass
+		case IngressDefault:
+			defaultIngressClass = options.IngressClass
```
kyessenov: "default" means applies to non-annotated and annotated with "istio". Is it the same meaning in status syncer? I could not find documentation for the config parameter.

platform/kube/controller.go
kyessenov: Regarding the silly `POD_IP` env variable use by the library, we should file an issue and do a hack to set the env variable for now. I think maybe add it `cmd/main.go` since there are similar flags there already.

platform/kube/controller.go
ijsnellf: Good catch. The meaning in ingress syncer doesn't match up with that:
```go
// IsValid returns true if the given Ingress either doesn't specify
// the ingress.class annotation, or it's set to the configured in the
// ingress controller.
func IsValid(ing *extensions.Ingress, controller, defClass string) bool {
	ingress, err := parser.GetStringAnnotation(IngressKey, ing)
	if err != nil && !errors.IsMissingAnnotations(err) {
		glog.Warningf("unexpected error reading ingress annotation: %v", err)
	}

	// we have 2 valid combinations
	// 1 - ingress with default class | blank annotation on ingress
	// 2 - ingress with specific class | same annotation on ingress
	//
	// and 2 invalid combinations
	// 3 - ingress with default class | fixed annotation on ingress
	// 4 - ingress with specific class | different annotation on ingress
	if ingress == "" && controller == defClass {
		return true
	}

	return ingress == controller
}
```
That would be 3, which is considers invalid.

ijsnellf:
I will add the integration test check in a follow on PR. I removed what I had because it is a bit flaky.

ijsnellf:
@kyessenov: PTAL

ijsnellf:
@kyessenov: I had an integration test inside the e2e framework, but it was flaky because of timing issues. Since this is causing issues for docs (among other things), I'll merge this and add in the test in a follow-on PR.

kyessenov:
Yes, go ahead

kyessenov:

kyessenov:

Issue #486, opened 2017-04-09
Ingress controller: watch secrets referenced by ingresses


zcahana:
Extend the `platform/kube.Controller` to watch secrets referenced within ingresses, and provide change notifications so that the proxy could be reloaded when a referenced secret changes.

Extend the `proxy/envoy.ingressWatcher` to subscribe for these notifications and reload certs/Envoy accordingly.

This can be used for in-place updates of TLS certificates stored in kube secrets.

zcahana:
Priority for this issue is probably low at this point. I'm opening it mostly to capture this idea for later.

ijsnellf:
Resolved by https://github.com/istio/manager/pull/563

PR #563, opened 2017-04-20
Use RDS/CDS for ingress proxy and rework secret propagation

kyessenov:
Remove controller loop from the ingress watcher and add a new HTTP endpoint to poll for secret UIDs. Secrets are fetched locally using kubernetes API from the secret UID. The secrets are hashed and incorporated in the config to ensure restarts on the content change.


kyessenov:
#533, @lookuptable you can work on adding auth for ingress once this is done.

lookuptable:
Great. Thanks for pushing this through.

proxy/envoy/config.go
```diff
@@ -175,23 +180,24 @@ func buildHTTPListener(mesh *proxyconfig.ProxyMeshConfig, routeConfig *HTTPRoute
 		config.RouteConfig = routeConfig
 	}
 
-	listener := &Listener{
-		Address: fmt.Sprintf("tcp://%s:%d", ip, port),
+	return &Listener{
+		BindToPort: true,
```
ayj: Leftover from debug?

proxy/envoy/discovery_test.go
```diff
@@ -225,6 +235,32 @@ func TestRouteDiscoveryFault(t *testing.T) {
...
+	response = makeDiscoveryRequest(ds, "GET", url, t)
+	compareResponse(response, "testdata/rds-ingress-ssl.json", t)
+}
+
+func TestSecretDiscovery(t *testing.T) {
+	registry := mock.MakeRegistry()
+	addIngressRoutes(registry, t)
+	ds := makeDiscoveryService(t, registry)
+	url := fmt.Sprintf("/v1alpha/secret/%s/%s", ds.MeshConfig.IstioServiceCluster, ingressNode)
+	response := makeDiscoveryRequest(ds, "GET", url, t)
```
ayj: (nit): s/response/got and s/secret/want

proxy/envoy/ingress_test.go
```diff
@@ -71,49 +66,9 @@ func compareFile(filename string, golden []byte, t *testing.T) {
...
-	addIngressRoutes(r, t)
-	testIngressConfig(&IngressConfig{
-		CertFile:  ingressCertFile,
-		KeyFile:   ingressKeyFile,
-		Namespace: ingressNamespace,
-		Secrets:   s,
-		Registry:  r,
-		Mesh:      &mesh,
-	}, ingressEnvoyPartialSSLConfig, t)
+	testIngressOptions(&mesh, ingressTLSSecret, ingressEnvoyConfig, t)
```
ayj: It looks like `testIngressOptions` is only called once in this function? Were you planning to reuse it for additional test functions? Otherwise it might be good to merge it into TestIngressRouteSSL.

platform/kube/client.go
```diff
@@ -372,3 +372,33 @@ func (cl *Client) convertConfig(item *Config) (name, namespace, kind string, dat
...
+	if len(parts) != 2 {
+		return nil, fmt.Errorf("URI %q does not match <name>.<namespace>", uri)
+	}
+
+	secret, err := cl.client.CoreV1().Secrets(parts[1]).Get(parts[0], meta_v1.GetOptions{})
+	if err != nil {
+		return nil, multierror.Prefix(err, "failed to retrieve secret "+uri)
+	}
+
+	cert := secret.Data[secretCert]
```
ayj: Do we ever need to differentiate between non-existent and empty key/cert? Otherwise this is probably fine.

proxy/envoy/config.go
kyessenov: it's actually specifically unset above. I think it's worth keeping the default explicitly set (missing bindtoport is "true")

proxy/envoy/discovery_test.go
kyessenov: ack

proxy/envoy/ingress_test.go
kyessenov: leftover from older code...

platform/kube/client.go
kyessenov: I don't think we need to differentiate between missing and incorrect key/certs.

proxy/envoy/ingress.go
```diff
@@ -15,256 +15,222 @@
...
 	}
+	return out, nil
+}
 
-	if err := ctl.AppendConfigHandler(model.IngressRule, func(model.Key, proto.Message, model.Event) {
-		out.reload()
-	}); err != nil {
-		return nil, err
-	}
+func (w *ingressWatcher) Run(stop <-chan struct{}) {
```
ayj: (not for this PR) might be worth passing `context.Context` around as first parameter to unify timeout/cancel handling.

proxy/envoy/ingress.go
```diff
@@ -15,256 +15,222 @@
...
+	client := &http.Client{Timeout: convertDuration(w.mesh.ConnectTimeout)}
+	url := fmt.Sprintf("http://%s/v1alpha/secret/%s/%s",
+		w.mesh.DiscoveryAddress, w.mesh.IstioServiceCluster, ingressNode)
+	initial := true
+	for {
+		tls, err := fetchSecret(ctx, client, url, w.secrets)
+		if err != nil && !initial {
+			glog.Warning(err)
+		} else {
+			config := generateIngress(w.mesh, tls, certFile, keyFile)
```
ayj: I think this is trying to only log a warning on the first attempt and only use `tls` if fetchSecret is successfully? But the code here will use `tls` on non-initial attempts when `err != nil`. Maybe something like the following?

```golang
if err != nil {
    if !warned {
        warned = true
    } // else skip warning to avoid flooding logs, maybe add a counter?
} else {
    warned = false // reset warning log whenever fetchSecret is successfully.
    // schedule update
}
```

proxy/envoy/ingress.go
kyessenov: Yes, I was thinking the same. kubernetes library doesn't use contexts (yet) though...

proxy/envoy/ingress.go
```diff
@@ -15,256 +15,222 @@
...
+	}
+	resp, err := client.Do(req.WithContext(ctx))
+	if err != nil {
+		return nil, multierror.Prefix(err, "failed to fetch "+url)
+	}
+	uri, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, multierror.Prefix(err, "failed to read request body")
+	}
+	if err = resp.Body.Close(); err != nil {
```
ayj: Doesn't resp.Body.Close() need to be invoked regardless of whether the body contents is read, i.e. before line 102? 

You can add the following after line 100:

```golang
defer res.Body.Close() // nolint: errcheck
```

proxy/envoy/ingress.go
```diff
@@ -15,256 +15,222 @@
...
+	secret := string(uri)
+	if secret == "" {
+		glog.Info("no secret needed")
+		return nil, nil
+	}
+	out, err := secrets.GetTLSSecret(secret)
+	if err != nil {
+		return nil, multierror.Prefix(err, "failed to read secret from storage")
+	}
+	glog.Info("secret successfully fetched")
```
ayj: It seems like this should be logged by the caller instead where other information could also be included, e.g #attempts, total # of fetches.

proxy/envoy/ingress.go
```diff
@@ -360,15 +336,16 @@ func extractPortAndTags(dst *config.DestinationWeight) (*model.Port, model.Tags,
 	}
 
 	var tags model.Tags
-	if len(dst.Tags) > 3 {
+	if len(dst.Tags) > 4 {
 		tags = make(model.Tags, len(dst.Tags)-3)
```
ayj: `-4` here too? Or just set it to zero since this isn't performance critical?

proxy/envoy/config.go
```diff
@@ -147,7 +152,7 @@ func buildListeners(context *proxy.Context) (Listeners, Clusters) {
 // Use "0.0.0.0" IP address to listen on all interfaces
 // RDS parameter controls whether to use RDS for the route updates.
 func buildHTTPListener(mesh *proxyconfig.ProxyMeshConfig, routeConfig *HTTPRouteConfig,
-	ip string, port int, rds bool, inbound bool) *Listener {
+	ip string, port int, rds bool) *Listener {
 	filters := buildFaultFilters(routeConfig)
```
rshriram: Why eliminate inbound ? Or treat both as same ?

rshriram: Ignore. Reviewing over phone is pain.

proxy/envoy/discovery.go
```diff
@@ -439,18 +449,27 @@ func (ds *DiscoveryService) ListClusters(request *restful.Request, response *res
...
 
 		// service-node holds the IP address
-		ip := request.PathParameter(ServiceNode)
+		node := request.PathParameter(ServiceNode)
+
 		// CDS computes clusters that are referenced by RDS routes for a particular proxy node
 		// TODO: this implementation is inefficient as it is recomputing all the routes for all proxies
 		// There is a lot of potential to cache and reuse cluster definitions across proxies and also
 		// skip computing the actual HTTP routes
-		instances := ds.Discovery.HostInstances(map[string]bool{ip: true})
```
rshriram: Remove comment block above as we have caching ?

proxy/envoy/discovery.go
```diff
@@ -526,9 +545,14 @@ func (ds *DiscoveryService) ListRoutes(request *restful.Request, response *restf
...
-		services := ds.Discovery.Services()
-		httpRouteConfigs := buildOutboundHTTPRoutes(instances, services, ds.Accounts, ds.MeshConfig, ds.Config)
+		var httpRouteConfigs HTTPRouteConfigs
+		if node == ingressNode {
+			httpRouteConfigs, _ = buildIngressRoutes(ds.Config.IngressRules(""))
+		} else {
+			instances := ds.Discovery.HostInstances(map[string]bool{node: true})
+			services := ds.Discovery.Services()
+			httpRouteConfigs = buildOutboundHTTPRoutes(instances, services, ds.Accounts, ds.MeshConfig, ds.Config)
+		}
```
rshriram: Bit unclear on the logic here. Will the envoy config in ingress container use CDS instead of hard coded clusters? (This allows us to set circuit breakers, change load balancing algorithms, etc). 

Will the envoy config in ingress container use RDS or use the spec from ingress resource (which just allows path, prefix, virtual hosts) or merge both based on the backend service selected in ingress? The latter would be the best of both, as we could define route rules (header match, retry, timeout, weighted clusters) for the destination

proxy/envoy/ingress.go
```diff
@@ -15,256 +15,222 @@
...
+	}
 
-	// Initialize envoy according to the current model state,
-	// instead of waiting for the first event to arrive.
-	// Note that this is currently done synchronously (blocking),
-	// to avoid racing with controller events lurking around the corner.
-	// This can be improved once we switch to a mechanism where reloads
-	// are linearized (e.g., by a single goroutine reloader).
-	w.reload()
-	w.ctl.Run(stop)
```
rshriram: Where is this functionality moved to? Iirc this was a bug fix.

proxy/envoy/testdata/envoy-ingress.json.golden
```diff
@@ -1,30 +1,18 @@
...
-                      "prefix": "/bar",
-                      "host_rewrite": "foo.default.svc.cluster.local:80",
-                      "cluster": "out.foo.default.svc.cluster.local|http"
-                    }
-                  ]
-                }
-              ]
+            "rds": {
+              "cluster": "rds",
+              "route_config_name": "80",
```
rshriram: This looks great! Now, where does this RDS gets its routes from? The manager ? If so, does that manager listen to ingress resource and generate the Envoy route config accordingly ? (Assuming that, it's still incomplete right, as we have no route rule support. But that's for another PR).

bin/e2e.sh
```diff
@@ -32,10 +32,13 @@ if [[ "$hub" =~ ^gcr\.io ]]; then
...
-  bazel $BAZEL_ARGS run //docker:${image}_debug
-  docker tag istio/docker:${image}_debug $hub/$image:$tag
+for image in app init manager; do
+  bazel $BAZEL_ARGS run //docker:${image}
+  docker tag istio/docker:${image} $hub/$image:$tag
   docker push $hub/$image:$tag
 done
+bazel $BAZEL_ARGS run //docker:proxy_debug
+docker tag istio/docker:proxy_debug $hub/proxy:$tag
+docker push $hub/proxy:$tag
```
kyessenov: docker is getting slow so limited debug image to just proxy.

platform/kube/client.go
```diff
@@ -372,3 +372,33 @@ func (cl *Client) convertConfig(item *Config) (name, namespace, kind string, dat
...
+	key := secret.Data[secretKey]
+	if len(cert) == 0 || len(key) == 0 {
+		return nil, fmt.Errorf("Secret keys %q and/or %q are missing", secretCert, secretKey)
+	}
+
+	return &model.TLSSecret{
+		Certificate: cert,
+		PrivateKey:  key,
```
kyessenov: TLS secret is per route, not per host. Ingress does not prevent defining multiple secrets for the same host or secret/no secret combination for a host.

proxy/envoy/discovery.go
kyessenov: Still inefficient to compute routes for clusters.

proxy/envoy/discovery.go
kyessenov: There's no merging logic yet. Ingress just gets ingress-based routes.

proxy/envoy/ingress.go
kyessenov: Gone. Unnecessary since ingress is no longer a controller.

proxy/envoy/testdata/envoy-ingress.json.golden
kyessenov: Manager listens to ingress resources.

platform/kube/client.go
rshriram: Yes, that's the spec. But do we restrict the config generation to just wildcard or tls for all hosts?

rshriram:
Fixes #486

platform/kube/client.go
kyessenov: We just take the first secret in the ingress ignoring "host" field completely. There's no SNI on listeners.

platform/kube/conversion.go
```diff
@@ -177,21 +177,28 @@ func convertIngress(ingress v1beta1.Ingress, getService serviceGetter) map[model
 		}
 	}
 
+	tls := ""
+	if len(ingress.Spec.TLS) > 0 {
```
kyessenov: this is where TLS secret is set.

proxy/envoy/ingress.go
kyessenov: This is to guarantee that ingress always runs at least on port 80.

proxy/envoy/ingress.go
kyessenov: ack, applied your suggestion.

proxy/envoy/ingress.go
kyessenov: removed this line. it's printed as part of the config anyways.

proxy/envoy/ingress.go
kyessenov: ack

kyessenov:
Please take a look.

PR #571, opened 2017-04-23
Change istioctl to talk to API server rather than direct to kube

liamawhite:
This is the functionality to get istioctl to talk to the API server rather than directly to Kubernetes for the proxy.

It still needs linting, etc. but the functionality should all be there.


cmd/istioctl/main.go
```diff
@@ -271,7 +300,7 @@ func init() {
 	rootCmd.AddCommand(putCmd)
 	rootCmd.AddCommand(getCmd)
 	rootCmd.AddCommand(deleteCmd)
-	rootCmd.AddCommand(injectCmd)
+	// rootCmd.AddCommand(injectCmd)
```
kyessenov: @ayj how are we going to deal with kube-inject in the kube-independent istio client?

cmd/istioctl/main.go
```diff
@@ -98,18 +118,18 @@ var (
...
 				return errors.New("nothing to create")
 			}
-			for _, v := range varr {
-				if err = setup(v.Type, v.Name); err != nil {
+			for _, config := range varr {
+				if err = setup(config.Type, config.Name); err != nil {
 					return err
 				}
-				if err = model.IstioConfig.ValidateConfig(&key, v.ParsedSpec); err != nil {
+				if err = model.IstioConfig.ValidateConfig(&key, config.ParsedSpec); err != nil {
```
kyessenov: remove this. the server is validating configs, right?

cmd/istioctl/main.go
```diff
@@ -58,6 +64,20 @@ var (
 		Long: fmt.Sprintf("Istio configuration command line utility. Available configuration types: %v",
 			model.IstioConfig.Kinds()),
 		PersistentPreRunE: func(*cobra.Command, []string) (err error) {
+			// Get manager address
+			add := os.Getenv("ISTIO_MANAGER_ADDRESS")
```
kyessenov: call it `addr`, maybe?

cmd/istioctl/main.go
```diff
@@ -58,6 +64,20 @@ var (
...
+			if add == "" {
+				return errors.New("manager address environment variable is not set," +
```
kyessenov: can you add a flag and populate it with env variable value, just like other flags do?

cmd/istioctl/main.go
```diff
@@ -131,18 +151,18 @@ var (
...
 				return errors.New("nothing to replace")
 			}
-			for _, v := range varr {
-				if err = setup(v.Type, v.Name); err != nil {
+			for _, config := range varr {
+				if err = setup(config.Type, config.Name); err != nil {
 					return err
 				}
-				if err = model.IstioConfig.ValidateConfig(&key, v.ParsedSpec); err != nil {
+				if err = model.IstioConfig.ValidateConfig(&key, config.ParsedSpec); err != nil {
```
kyessenov: same here. validation should be done server side. ideally the client should be lightweight and future-compatible.

cmd/istioctl/main.go
```diff
@@ -131,18 +151,18 @@ var (
...
 					return err
 				}
-				err = config.Put(key, v.ParsedSpec)
+				err = apiClient.UpdateConfig(key, config)
 				if err != nil {
 					return err
 				}
-				fmt.Printf("Put %v %v\n", v.Type, v.Name)
+				fmt.Printf("Put %v %v\n", config.Type, config.Name)
```
kyessenov: Change message to "updated config"?

cmd/istioctl/main.go
kyessenov: Let's keep this in. I'd suggest moving kube inject variables to inject.go to make it easy to track dependencies.

cmd/istioctl/istioctl_test.go
```diff
@@ -15,159 +15,359 @@
...
-	"istio.io/manager/test/mock"
+	"strings"
+
+	"istio.io/manager/apiserver"
+	"istio.io/manager/model"
 )
 
-func rootSetup(t *testing.T) {
-	config = mock.MakeRegistry()
+type StubClient struct {
```
kyessenov: thought: we have `test/mock/config.go` for mocking the config registry. if it makes it easier might be able to wrap the client around it. what you have also works.

cmd/istioctl/istioctl_test.go
```diff
@@ -15,159 +15,359 @@
...
+		return fmt.Errorf("received unexpected config type: %s, wanted: %s", config.Type, wantConfig.Type)
 	}
+	return nil
+}
 
-	// YAML output
-	outputFormat = "yaml"
-	if err := getCmd.RunE(getCmd, []string{"route-rules"}); err != nil {
-		t.Fatalf("Could not list routes: %v", err)
+func TestClientSideValidation(t *testing.T) {
```
kyessenov: i'm assuming client side validation is just doing YAML to proto parsing?

cmd/istioctl/istioctl_test.go
```diff
@@ -15,159 +15,359 @@
...
+			configKeyMapReq: true,
+			outFormat:       "yaml",
+		},
+		{
+			name:            "TestGetNotYAMLOrShort",
+			command:         "get",
+			arg:             []string{"route-rule"},
+			configKeyMapReq: true,
+			outFormat:       "not-an-output-format",
+			wantError:       "unknown output format not-an-output-format. Types are yaml|short",
```
ayj: Comparing specific error message strings is brittle. Better to define a new error type if its important to check for a specific returned error. Or make `wantError` a `bool` and remove lines 369-371 below.

cmd/istioctl/main.go
ayj: inject.go and any other platform specific commands can themselves  to rootCmd in their own init() function. We can use build tags to selectively include the platform specific commands in the build if that is needed.

cmd/istioctl/istioctl_test.go
liamawhite: I like the idea in theory but it gives us less control at the unit test level and introduces a bit of test scope leakage. If we update the config registry interface it then may also break these tests.

cmd/istioctl/main.go
rshriram: We can ship istioctl with all platform specific sub commands. If the user tries to do kube-inject against a mesos cluster, the command would ultimately fail.

cmd/istioctl/istioctl_test.go
liamawhite: It also calls config.ParseSpec. Its all validated server side as well but it just saves a call.

rshriram: I would suggest we keep all config validation at server side only. Otherwise we risk having old clients providing a (locally valid) config that is invalid at a newer server.

cmd/istioctl/main.go
liamawhite: It is but this saves a request?

cmd/istioctl/istioctl_test.go
liamawhite: Removed the spec parsing client side. We still want to check for valid types/kinds client side though as they are part of the command. The remaining validation tests are that the YAMLs can be read and converted into JSON/proto so we have to do them client side as well anyway.

liamawhite:
retrigger

kyessenov:

kyessenov:

liamawhite:
Hold off on the merge I want to work out why my tests didn't pick up on the lack of content-type header.

cmd/istioctl/main.go
```diff
@@ -58,6 +60,25 @@ var (
 		Long: fmt.Sprintf("Istio configuration command line utility. Available configuration types: %v",
 			model.IstioConfig.Kinds()),
 		PersistentPreRunE: func(*cobra.Command, []string) (err error) {
+			// Get manager address
+			if managerAddr == "" {
+				if a := os.Getenv("ISTIO_MANAGER_ADDRESS"); a != "" {
```
kyessenov: you can just put `os.GetEnv` as the default value for `managerAddr` flag.

cmd/istioctl/main.go
```diff
@@ -58,6 +60,25 @@ var (
...
+					glog.V(2).Infof("Setting manager address from ISTIO_MANAGER_ADDRESS environment variable")
+					managerAddr = a
+				} else {
+					return errors.New("manager address environment variable is not set, " +
+						"please set ISTIO_MANAGER_ADDRESS to the location and port of you Istio manager")
+				}
+
+			}
+			url, err := url.Parse(managerAddr)
```
kyessenov: would be better to avoid using same name as the package name

cmd/istioctl/main.go
```diff
@@ -178,17 +193,26 @@ var (
 					return err
 				}
 
-				list, err := config.List(key.Kind, key.Namespace)
+				list, err := apiClient.ListConfig(key.Kind, key.Namespace)
 				if err != nil {
-					return fmt.Errorf("error listing %s: %v", key.Kind, err)
+					return err
+				}
+				printList := make(map[model.Key]proto.Message)
```
kyessenov: please use an array or sort a map. the list iteration should be deterministic.

cmd/istioctl/main.go
```diff
@@ -163,11 +178,11 @@ var (
...
 				}
-				item, exists := config.Get(key)
-				if !exists {
-					return fmt.Errorf("%q does not exist", key)
+				config, err := apiClient.GetConfig(key)
+				if err != nil {
+					return err
 				}
-				out, err := schema.ToYAML(item)
+				out, err := schema.ToYAML(config.ParsedSpec)
```
kyessenov: I'm still not sure why we need schema here. The API server can just send proto-decoded JSON and we simply convert JSON to YAML, right?

cmd/istioctl/inject.go
```diff
@@ -46,6 +48,10 @@ var (
 
 	inFilename  string
 	outFilename string
+
+	kubeconfig string
+	client     *kube.Client
+	config     model.ConfigRegistry
```
kyessenov: not used?

client/proxy/proxy.go
```diff
@@ -21,6 +21,15 @@ type ManagerClient struct {
 	client           *http.Client
 }
 
+// Client defines the interface for the proxy specific functionality of the manager client
+type Client interface {
+	GetConfig(model.Key) (*apiserver.Config, error)
```
kyessenov: I'm confused how come ParsedSpec is transferred over the wire if it's marked as `json:"-"`.

cmd/istioctl/main.go
liamawhite: Its checking that there isn't a flag set, similar to the kubeconfig.

cmd/istioctl/main.go
liamawhite: This was here previously, I assumed it was to verify what we retrieved was actually a legitimate entry.

client/proxy/proxy.go
liamawhite: Ah crap, this was all fine as I parsed it client side but then I removed it so we could reduce the client side validation...

cmd/istioctl/inject.go
liamawhite: It is used in `main.go` to setup the kube inject stuff

PR #572, opened 2017-04-24
Set cluster SSLContext for Ingress when Istio auth is enabled

lookuptable:


lookuptable:
@kyessenov Do we also need the SSLContext in the response for `ListAllClusters`

rshriram:
@ijsnellf FYI.

lookuptable:
@kyessenov In the first commit, SSLContext is generated only for `ListClusters`. But we also have `ListAllClusters`. So in the second commit I moved SSLContext generation into `buildIngressRoute` and `buildOutboundHTTPRoutes` so make sure we have SSLContext for both `ListClusters` and `ListAllClusters`.

kyessenov:
I liked the first method, since it's simpler. I think we need to refactor `ListAllClusters` to use `ListClusters`. That method is only used for debugging and does not affect functionality.

lookuptable:
Talked with @kyessenov offline - we're going with the first approach where we set SSLContext in `ListClusters`

test/integration/driver.go
```diff
@@ -125,7 +125,6 @@ func setAuth(params infra) infra {
 	out := params
 	out.Name = "(auth infra)"
 	out.Auth = proxyconfig.ProxyMeshConfig_MUTUAL_TLS
-	out.Ingress = false
```
rshriram: is this the only place that you need to test for ingress integration with auth? I vaguely remember auth being disabled in reachability tests.

kyessenov: yes, I refactored it so that tests are self-contained.

PR #573, opened 2017-04-24
prepare_proxy: add support for excluding external traffic

ayj:
(Note for reviewers: prepare_proxy.sh is ready for review. prepare_proxy_test.sh is functional with some hand edits required for the configuration. It could probably be rewritten in golang to make it more readable.)

Short term solution to #515 for external traffic.

Add generic support for including and excluding outbound traffic based
on CIDR blocks. Default behavior is opt-out where all outbound traffic
is redirected to proxy unless explicitly opted-opt with "-e"
option. When the "-i" option is used the behavior changes to opt-in
and only outbound traffic explicitly specified with "-i" is redirected
to the proxy. The "-i" option takes precedence over "-e" the option if
both are specified.

As part of this change the iptable rules were refactored to make it
easier differentiate redirect from bypass cases, e.g. '-j
ISTIO_REDIRECT' vs. '-j RETURN'.


rshriram:
Can we reduce the confusion on opt-out and opt-in and just have one? Like a set of -i flags. With no -i, all traffic goes through proxy. With -i, operators could restrict traffic capture to specific CIDR blocks.

docker/prepare_proxy.sh
```diff
@@ -40,20 +51,53 @@ if [[ -z "${ISTIO_PROXY_PORT-}" ]] || [[ -z "${ISTIO_PROXY_UID-}" ]]; then
...
+# istio and attach it the OUTPUT rule for all tcp traffic. '-j RETURN'
+# bypasses the proxy port; '-j ISTIO_REDIRECT' redirects to the proxy
+# port.
+iptables -t nat -N ISTIO_OUTPUT
+iptables -t nat -A OUTPUT -p tcp -j ISTIO_OUTPUT                                    -m comment --comment "istio/install-istio-output"
+
+# Locally routed traffic is not captured by PREROUTING chain. Redirect
+# loopback back traffic when the DST_IP is not explicitly the loopback
+# address to handles appN => proxy (client) => proxy (server) => appN.
+iptables -t nat -A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -j ISTIO_REDIRECT           -m comment --comment "istio/implicit-loopback"
```
rshriram: Probably best to describe the use case instead of the data flow. E.g., when kubelet accesses the app for health checks (??), or app calls itself via http://podIP/..

docker/prepare_proxy.sh
```diff
@@ -40,20 +51,53 @@ if [[ -z "${ISTIO_PROXY_PORT-}" ]] || [[ -z "${ISTIO_PROXY_UID-}" ]]; then
...
+
+# Avoid infinite loops. Don't redirect proxy traffic directly back to
+# the proxy for non-loopback traffic.
+iptables -t nat -A ISTIO_OUTPUT -m owner --uid-owner ${ISTIO_PROXY_UID} -j RETURN   -m comment --comment "istio/exclude-proxy"
+
+# Skip redirection for proxy-aware applications and
+# container-to-container traffic which explicitly use localhost.
+iptables -t nat -A ISTIO_OUTPUT -d 127.0.0.1/32 -j RETURN                           -m comment --comment "istio/explicit-loopback"
+
+# The default outobund redirection policy is opt-out. IP ranges may be
```
rshriram: Typo. Outbound

docker/prepare_proxy.sh
rshriram: And shouldn't this be restricted to tcp?

docker/prepare_proxy.sh
```diff
@@ -40,20 +51,53 @@ if [[ -z "${ISTIO_PROXY_PORT-}" ]] || [[ -z "${ISTIO_PROXY_UID-}" ]]; then
...
+if [ "${ISTIO_IP_RANGES_INCLUDE}" = "" ]; then
+    for cidr in ${ISTIO_IP_RANGES_EXCLUDE}; do
+        iptables -t nat -A ISTIO_OUTPUT -d ${cidr} -j RETURN                        -m comment --comment "istio/exclude-cidr-block-${cidr}"
+    done
+    iptables -t nat -A ISTIO_OUTPUT -j ISTIO_REDIRECT                               -m comment --comment "istio/default-opt-in-redirect"
+else
+    for cidr in ${ISTIO_IP_RANGES_INCLUDE}; do
+        iptables -t nat -A ISTIO_OUTPUT -d ${cidr} -j ISTIO_REDIRECT                -m comment --comment "istio/include-cidr-block-${cidr}"
+    done
+    iptables -t nat -A ISTIO_OUTPUT -j RETURN                                       -m comment --comment "istio/default-opt-out-redirect"
```
rshriram: If we follow the suggestion I made earlier (just have -i cidr), then this else block moves to the top doesn't it? We would no longer require the default catch all policy either.

docker/prepare_proxy_test.sh
```diff
@@ -0,0 +1,194 @@
...
+CLIENT_IP=
+while [ "$CLIENT_IP" = "" ]; do CLIENT_IP=$(kc get pod -l app=client -o jsonpath='{.items[0].status.podIP}'); done
+SERVER=
+while [ "$SERVER" = "" ]; do SERVER=$(kc get pod -l app=server -o jsonpath='{.items[0].metadata.name}'); done
+SERVER_IP=
+while [ "$SERVER_IP" = "" ]; do SERVER_IP=$(kc get pod -l app=server -o jsonpath='{.items[0].status.podIP}'); done
+
+if false; then
+    # exclude specific external traffic (opt-out)
+    EXCLUDE_CIDR=$(host httpbin.org | cut -f4 -d' ' | sed -e 's|$|/32|' | paste -sd ",")
```
rshriram: What's the point of this if block ?

docker/prepare_proxy_test.sh
```diff
@@ -0,0 +1,194 @@
...
+
+    # Enable proxy redirection on server only
+    kc exec ${SERVER} -c init -- \
+       /usr/local/bin/prepare_proxy.sh -u ${ISTIO_PROXY_UID} -p ${ISTIO_PROXY_PORT} -e ${EXCLUDE_CIDR}
+else
+    # include specific internal service and pod traffic (opt-in)
+    INCLUDE_CIDR=$(gcloud container clusters describe ${GCLOUD_CLUSTER_NAME} --zone=${GCLOUD_CLUSTER_ZONE} |
+		       grep -e clusterIpv4Cidr -e servicesIpv4Cidr |
+		       cut -f2 -d' ' | paste -sd ",")
+
```
rshriram: Could you perhaps abstract this function out so that people could test this on say minikube or their local k8s clusters?

docker/prepare_proxy.sh
```diff
@@ -40,20 +51,53 @@ if [[ -z "${ISTIO_PROXY_PORT-}" ]] || [[ -z "${ISTIO_PROXY_UID-}" ]]; then
...
-# This include any proxy generated traffic which is necessary to make
-# (app => proxy => proxy => app) work when two apps are same pod.
-iptables -t nat -A OUTPUT -p tcp -o lo -j REDIRECT \
-  ! -d 127.0.0.1/32 --to-port ${ISTIO_PROXY_PORT}
-
-# 3. Locally generated traffic not sent explicitly from loopback IP
-# (i.e. proxy aware) or not from the proxy itself is redirected proxy port.
-iptables -t nat -A OUTPUT -p tcp -j REDIRECT ! -s 127.0.0.1/32 \
-  --to-port ${ISTIO_PROXY_PORT} -m owner '!' --uid-owner ${ISTIO_PROXY_UID}
+iptables -t nat -F
```
ayj: We don't. It was useful for unit testing but can be removed since prepare_proxy_test.sh restarts the pods for each test iteration.

docker/prepare_proxy.sh
```diff
@@ -40,20 +51,53 @@ if [[ -z "${ISTIO_PROXY_PORT-}" ]] || [[ -z "${ISTIO_PROXY_UID-}" ]]; then
...
+iptables -t nat -A ISTIO_REDIRECT -p tcp -j REDIRECT --to-ports ${ISTIO_PROXY_PORT} -m comment --comment "istio/redirect-to-proxy-port"
+
+# Redirect all inbound traffic to the proxy.
+iptables -t nat -A PREROUTING -j ISTIO_REDIRECT                                     -m comment --comment "istio/install-istio-prerouting"
+
+# Create a new chain for selectively redirecting outbound packets to
+# istio and attach it the OUTPUT rule for all tcp traffic. '-j RETURN'
+# bypasses the proxy port; '-j ISTIO_REDIRECT' redirects to the proxy
+# port.
+iptables -t nat -N ISTIO_OUTPUT
```
ayj: The preceding comment was intended to cover lines 68 and 69 (i.e. create chain *and* attach to OUTPUT rule). I'll split this up to make it more clear.

docker/prepare_proxy.sh
ayj: We only jump to ISTIO_OUTPUT for tcp protocol traffic (line 69) so the `-p tcp` seemed redundant. I can add it back if that makes things more clear though.

docker/prepare_proxy_test.sh
ayj: It creates a comma-separated list of IP ranges for httpbin.org to be excluded from redirection.

````bash
$ host httpbin.org | cut -f4 -d' ' | sed -e 's|$|/32|' | paste -sd ","
107.21.206.81/32,50.19.93.247/32,54.225.109.133/32,54.243.85.55/32,23.23.128.123/32,23.23.223.197/32,23.21.77.86/32,23.23.117.228/32
```

rshriram: No, I meant the `if false` block. This is never going to be executed is it?

ayj: Ah, this was for manual testing. I'll added a flag at the top of control this as part of the refactoring to remove the EXCLUDE option.

docker/prepare_proxy.sh
```diff
@@ -45,59 +40,52 @@ while getopts ":p:u:e:i:h" opt; do
...
 
-# The default outobund redirection policy is opt-out. IP ranges may be
-# selectively excluded from redirection with ISTIO_IP_RANGES_EXCLUDE.
-#
-# If ISTIO_IP_RANGES_INCLUDE is defined the outbound redirection
-# policy is default opt-in. IP ranges may be selectively included for
-# redirection with ISTIO_IP_RANGES_INCLUDE.
+# Only redirect outbound traffic to selective IP ranges when
+# ISTIO_IP_RANGES_INCLUDE is non-empty. Otherwise redirect all
+# outbound traffic not already handled by the previous rules to Envoy.
```
rshriram: I would reuse a previous version of this comment as it looked better.
All outbound traffic will be redirected to Envoy by default. If ISTIO_IP_RANGES_INCLUDE is non-empty, only traffic bound for the destinations specified in this list will be captured.

docker/prepare_proxy_test.sh
```diff
@@ -11,11 +11,12 @@ TAG=test
...
 GCLOUD_CLUSTER_ZONE=us-central1-a
-ISTIO_PROXY_UID=1337
-ISTIO_PROXY_PORT=80
-SERVER_PORT=${ISTIO_PROXY_PORT}
+ENVOY_UID=1337
+ENVOY_PORT=80
+SERVER_PORT=${ENVOY_PORT}
 CLIENT_PORT=81
 OTHER_PORT=82
+TEST_IP_RANGE_INCLUDE=0
```
rshriram: add a variable: PLATFORM="gcloud"

docker/prepare_proxy_test.sh
```diff
@@ -67,28 +68,29 @@ while [ "$SERVER" = "" ]; do SERVER=$(kc get pod -l app=server -o jsonpath='{.it
...
-if false; then
-    # exclude specific external traffic (opt-out)
-    EXCLUDE_CIDR=$(host httpbin.org | cut -f4 -d' ' | sed -e 's|$|/32|' | paste -sd ",")
-    EXTERNAL=http://httpbin.org/status/418
-    EXTERNAL_MATCH=TEAPOT
+# Return the kubernetes service and pod IP ranges as a comma seperated
+# list, e.g. 10.0.0.1/32,10.2.0.1/16. This function is GKE specific
+# and must modified for other providers, e.g. minikube.
+function k8sClusterAndServiceIPRange() {
+    echo $(gcloud container clusters describe ${GCLOUD_CLUSTER_NAME} --zone=${GCLOUD_CLUSTER_ZONE} |
```
rshriram: case $platform in
 gcloud)
   echo..
 *)
   echo ""
esac

ayj:
Fixes #515 for alpha.

docker/prepare_proxy.sh
```diff
@@ -8,22 +8,28 @@ set -o pipefail
 usage() {
   echo "${0} -p PORT -u UID [-h]"
   echo ''
-  echo '  -p: Specify the proxy port to which redirect all TCP traffic'
+  echo '  -p: Specify the envoy port to which redirect all TCP traffic'
   echo '  -u: Specify the UID of the user for which the redirection is not'
   echo '      applied. Typically, this is the UID of the proxy container'
+  echo '  -i: Comma separated list of IP ranges in CIDR form to redirection to envoy (optional)'
```
kyessenov: "for redirection"?

docker/prepare_proxy.sh
```diff
@@ -8,22 +8,28 @@ set -o pipefail
...
     u)
-      ISTIO_PROXY_UID=${OPTARG}
+      ENVOY_UID=${OPTARG}
+      ;;
+    i)
+      ISTIO_IP_RANGES_INCLUDE=${OPTARG}
       ;;
     h)
-      usage
+      usagey
```
kyessenov: typo?

docker/prepare_proxy.sh
```diff
@@ -8,22 +8,28 @@ set -o pipefail
...
     p)
-      ISTIO_PROXY_PORT=${OPTARG}
+      ENVOY_PORT=${OPTARG}
       ;;
     u)
-      ISTIO_PROXY_UID=${OPTARG}
+      ENVOY_UID=${OPTARG}
+      ;;
+    i)
+      ISTIO_IP_RANGES_INCLUDE=${OPTARG}
```
kyessenov: why rename `ISTIO_PROXY_PORT` but keep this `ISTIO_`?

kyessenov: there's not much specific to envoy in this script btw

docker/prepare_proxy.sh
```diff
@@ -34,26 +40,53 @@ while getopts ":p:u:h" opt; do
...
-  ! -d 127.0.0.1/32 --to-port ${ISTIO_PROXY_PORT}
-
-# 3. Locally generated traffic not sent explicitly from loopback IP
-# (i.e. proxy aware) or not from the proxy itself is redirected proxy port.
-iptables -t nat -A OUTPUT -p tcp -j REDIRECT ! -s 127.0.0.1/32 \
-  --to-port ${ISTIO_PROXY_PORT} -m owner '!' --uid-owner ${ISTIO_PROXY_UID}
+# Create a new chain for redirecting inbound and outbound traffic to
+# the common Envoy port.
+iptables -t nat -N ISTIO_REDIRECT                                             -m comment --comment "istio/redirect-common-chain"
+iptables -t nat -A ISTIO_REDIRECT -p tcp -j REDIRECT --to-ports ${ENVOY_PORT} -m comment --comment "istio/redirect-to-envoy-port"
```
kyessenov: "to-ports"?

docker/prepare_proxy_test.sh
```diff
@@ -0,0 +1,238 @@
+#!/bin/sh
```
kyessenov: I'd really prefer not have these long bash test scripts. They are close to unmaintainable...

docker/prepare_proxy.sh
```diff
@@ -34,26 +40,53 @@ while getopts ":p:u:h" opt; do
...
+iptables -t nat -A ISTIO_OUTPUT -o lo ! -d 127.0.0.1/32 -j ISTIO_REDIRECT     -m comment --comment "istio/redirect-implicit-loopback"
+
+# Avoid infinite loops. Don't redirect Envoy traffic directly back to
+# Envoy for non-loopback traffic.
+iptables -t nat -A ISTIO_OUTPUT -m owner --uid-owner ${ENVOY_UID} -j RETURN   -m comment --comment "istio/bypass-envoy"
+
+# Skip redirection for Envoy-aware applications and
+# container-to-container traffic both of which explicitly use
+# localhost.
+iptables -t nat -A ISTIO_OUTPUT -d 127.0.0.1/32 -j RETURN                     -m comment --comment "istio/bypass-explicit-loopback"
```
kyessenov: I have difficulty understanding all implications of these rules, so I trust @rshriram carefully reviewed these

ayj: Agreed. Most of the setup/teardown code can be shared with the other integration tests. Maybe after release we can split that out into its own test package and have individual tests consume that that?

docker/prepare_proxy.sh
ayj: Either will work, but changed to "to-port" to avoid confusion.

docker/BUILD
```diff
@@ -48,9 +48,17 @@ pkg_tar(
     tags = ["manual"],
 )
 
-docker_build(
-    name = "init",
-    base = "@docker_ubuntu//:xenial",
+manager_docker_build(
```
kyessenov: please put those back in jenkinsfile if you need them.

PR #580, opened 2017-04-24
istioctl: improve error message when istio ConfigMap isn't found

### LINKED ISSUE #569: istioctl kube-inject in a namespace
If user wants to deploy in a different namespace than default, they have to specify -n <ns> in the kube-inject command, in addition to specifying in kubectl command. Otherwise, the  error "Cannot find ConfigMap" is displayed, which does not offer a good indication of the problem.

Can we avoind having to specify -n, or at a minimum, display a more relevant error and a usage example?

ayj:
Fixes #569


cmd/istioctl/inject.go
```diff
@@ -100,7 +100,7 @@ Example usage:
 
 			mesh, err := cmd.GetMeshConfig(client.GetKubernetesClient(), namespace, meshConfig)
 			if err != nil {
-				return err
+				return fmt.Errorf("Please install Istio system into kubernetes before running kube-inject: %v", err)
```
mandarjog: The error message should be
`could not find config map   ${namespace}/${meshConfig}, is Istio system installed in ${namespace} ?`

cmd/istioctl/inject.go
```diff
@@ -100,7 +100,8 @@ Example usage:
 
 			mesh, err := cmd.GetMeshConfig(client.GetKubernetesClient(), namespace, meshConfig)
 			if err != nil {
-				return err
+				return fmt.Errorf("could not find config map %s/%s. Is Istio system installed in %s?",
```
kyessenov: maybe drop first namespace (that's not standard configmap name) and use quotes %q?

PR #581, opened 2017-04-24
Validate proxy mesh config

### LINKED ISSUE #539: need validation for proxymeshconfig
We currently have validation for the routing proto, but lack validation for the ProxyMeshConfig which also touches some aspects of Envoy config. For example, users can set the cluster connect timeout via the proxy mesh config. These fields need to be validated/sanitized to ensure that incorrect values ( negative timeouts) don't creep in (this would cause Envoy to crash).

kyessenov:
Fixes #539


model/validation_test.go
```diff
@@ -0,0 +1,696 @@
...
+		},
+			valid: false},
+	}
+	for _, c := range cases {
+		if got := ValidateDestinationPolicy(c.in); (got == nil) != c.valid {
+			t.Errorf("Failed: got valid=%t but wanted valid=%v: %v", got == nil, c.valid, got)
+		}
+	}
+}
+
```
kyessenov: everything above just got moved (git failed to identify a diff)

model/validation.go
```diff
@@ -669,3 +661,89 @@ func ValidateDestinationPolicy(msg proto.Message) error {
...
+		errs = multierror.Append(errs, multierror.Prefix(err, "invalid refresh delay:"))
+	}
+
+	if err := validateDuration(mesh.ConnectTimeout); err != nil {
+		errs = multierror.Append(errs, multierror.Prefix(err, "invalid connect timeout:"))
+	}
+
+	if mesh.AuthCertsPath == "" {
+		errs = multierror.Append(errs, errors.New("invalid auth certificates path"))
+	}
```
lookuptable: Can we also verify `mesh.AuthPolicy`? Then we can probably get rid of the check here: https://github.com/istio/manager/blob/master/proxy/envoy/discovery.go#L489

model/validation.go
```diff
@@ -669,3 +661,89 @@ func ValidateDestinationPolicy(msg proto.Message) error {
 
 	return errs
 }
+
+func validateProxyAddress(hostAddr string) error {
```
ayj: https://golang.org/pkg/net/#SplitHostPort could be used here.

model/validation.go
```diff
@@ -669,3 +661,89 @@ func ValidateDestinationPolicy(msg proto.Message) error {
...
+	}
+	return nil
+}
+
+func validateDuration(pd *duration.Duration) error {
+	dur, err := ptypes.Duration(pd)
+	if err != nil {
+		return err
+	}
+	if dur <= 0 {
```
ayj: Maybe a good idea to check max duration here as well? This would need to be specified as paramter to this function but would help catch fat-fingering Xs instead of Xs when hand editing config maps.

kyessenov: @rshriram what are you good upper bounds for durations in this config?

model/validation.go
kyessenov: thanks, good call.

model/validation.go
kyessenov: hm, not sure about ipv6 support in envoy.

ayj: The current code doesn't validate the host part either. Maybe something like the following?

```golang
host, port, err := net.SplitHostPort(hostAddr)
if err != { return err }

if _, err := net.ResolveIPAddr("ip6", host); err == nil {
    return fmt.Errorf("ipv6 not supported: %v", err)
}
// port validation, etc.
```

kyessenov: I couldn't quite see if golang checks for DNS names. Envoy can use either IPv4 or DNS host but not IPv6.

ayj: Yup. That's why I reversed the check to be anything *not* ipv6. The documentation says checking against hostname isn't recommending since DNS can resolve to more than one address.

kyessenov: I don't think we should be resolving DNS names in validation. First, DNS resolvers might be different between golang and envoy. Second, DNS can change the IP assignment.

ayj: Agreed. Maybe the following? Either way, host validation is a different issue from splitting up the hostAddr into host/port tuple./

```golang
if ip := net.ParseIP(host); ip != nil && ip.To4() == nil {
    return fmt.Errors("ipv6 address: %v, host)
}
```

kyessenov:

model/validation.go
kyessenov: Not sure what you suggest. "my-name" and "10.0.0.10" are both valid envoy hosts. This will just detect the second case.

ayj: The snippet above returns an error if the host is IPv6. Anything else (IPv4 or DNS name) should be okay. I believe that is what you wanted to validate?

See https://play.golang.org/p/zsWMAoutrB

kyessenov: since we already had functions to validate both host and ipv4, I added host validation.

proxy/envoy/config.go
```diff
@@ -196,10 +196,7 @@ func applyInboundAuth(listener *Listener, mesh *proxyconfig.ProxyMeshConfig) *Li
 	case proxyconfig.ProxyMeshConfig_NONE:
 	case proxyconfig.ProxyMeshConfig_MUTUAL_TLS:
 		listener.SSLContext = buildListenerSSLContext(mesh.AuthCertsPath)
-	default:
```
andraxylia: Not a good idea to remove the default case.

model/validation.go
rshriram: upper and lower bounds.
drain/parent shutdown duration: >1s <1hr?
refresh delay: >1s <10m ?
connect timeout: >1ms <30s

PR #588, opened 2017-04-25
Use RDS for egress proxy and enable mutual TLS auth for egress

kyessenov:


cmd/manager/main.go
```diff
@@ -160,7 +160,7 @@ var (
 
 	ingressCmd = &cobra.Command{
 		Use:   "ingress",
-		Short: "Istio Proxy ingress controller",
+		Short: "Istio Proxy ingress agent",
```
rshriram: Envoy ingress agent

cmd/manager/main.go
```diff
@@ -177,18 +177,10 @@ var (
 		Use:   "egress",
 		Short: "Istio Proxy external service agent",
 		RunE: func(c *cobra.Command, args []string) error {
-			controller := kube.NewController(client, mesh, flags.controllerOptions)
```
rshriram: Envoy external service proxy

test/integration/egress.go
```diff
@@ -31,6 +34,13 @@ func (t *egress) String() string {
 }
 
 func (t *egress) setup() error {
+	if !t.Egress {
+		return nil
+	}
+	if err := util.Run(fmt.Sprintf(
+		"kubectl -n %s apply -f test/integration/testdata/external-service.yaml", t.Namespace)); err != nil {
+		return err
+	}
```
ayj: (nit) you can return from util.Run() directly, e.g.

```golang
cmd := mt.Sprintf("kubectl -n %s apply -f test/integration/testdata/external-service.yaml", t.Namespace)
return util.Run(cmd)
```

test/integration/egress.go
```diff
@@ -65,4 +75,13 @@ func (t *egress) run() error {
 }
 
 func (t *egress) teardown() {
+	if !t.Egress {
+		return
+	}
+	if err := client.CoreV1().Services(t.Namespace).Delete("httpbin", &meta_v1.DeleteOptions{}); err != nil {
```
ayj: (optional): you could use "kubectl delete" to avoid hardcoding the individual service names, e.g. 

```golang
if err := util.Run(mt.Sprintf("kubectl -n %s delete -f %s", externalServiceFile, t.Namespace)); err != nil {
    glog.Warning(err)
}
```

proxy/envoy/egress.go
```diff
@@ -26,113 +28,60 @@ import (
...
-	// instead of waiting for the first event to arrive.
-	// Note that this is currently done synchronously (blocking),
-	// to avoid racing with controller events lurking around the corner.
-	// This can be improved once we switch to a mechanism where reloads
-	// are linearized (e.g., by a single goroutine reloader).
-	w.reload()
-	w.ctl.Run(stop)
+func getEgressProxyPort(mesh *config.ProxyMeshConfig) int {
+	addr := mesh.EgressProxyAddress
+	port, _ := strconv.Atoi(addr[strings.Index(addr, ":")+1:])
```
lookuptable: Just to confirm: we verify the addr format in the validation process so it's safe to ignore the error here?

test/integration/driver.go
```diff
@@ -125,7 +125,6 @@ func setAuth(params infra) infra {
 	out := params
 	out.Name = "(auth infra)"
 	out.Auth = proxyconfig.ProxyMeshConfig_MUTUAL_TLS
-	out.Egress = false
```
lookuptable: We can now remove both `Ingress` and `Egress` flags since they're always true. I can do it in a follow-up PR.

proxy/envoy/discovery.go
```diff
@@ -471,19 +474,23 @@ func (ds *DiscoveryService) ListClusters(request *restful.Request, response *res
...
+			// apply custom policies for HTTP clusters
 			for _, cluster := range clusters {
-				ports := model.PortList{cluster.port}.GetNames()
-				serviceAccounts := ds.Accounts.GetIstioServiceAccounts(cluster.hostname, ports)
-				cluster.SSLContext = buildClusterSSLContext(ds.MeshConfig.AuthCertsPath, serviceAccounts)
+				insertDestinationPolicy(ds.Config, cluster)
+			}
+
+			// apply auth policies
+			switch auth := ds.MeshConfig.AuthPolicy; auth {
```
lookuptable: nit: just `switch ds.MeshConfig.AuthPolicy` (previously assigned it to a var since it's used in the default clause for logging).

cmd/manager/main.go
kyessenov: ack

cmd/manager/main.go
kyessenov: ack. fairly soon, we should merge the two edge proxies into one.

test/integration/egress.go
kyessenov: I'd like to keep it as is since it follows the same pattern for multi-step setups. it will be easier to add another step.

proxy/envoy/egress.go
kyessenov: yes, it's validated on start-up.

test/integration/egress.go
kyessenov: yeah not sure what's better - hardcoding file name or service names. both are referenced in the file already.

test/integration/driver.go
kyessenov: i'd like to keep those flags. we don't know if we need them again and they are not hard to maintain.

proxy/envoy/discovery.go
kyessenov: ok, done

PR #590, opened 2017-04-25
Fault injection, hostname, and env var output for echo server

esnible:
This modifies the 'echo' server.

First, in addition to outputting HTTP headers, it outputs the hostname of the server responding.  This is handy for verifying load balancing interactively.

Second it outputs environment variables beginning with ECHO.  This is handy for verifying label-specific routing interactively if the instance is started with env variables beginning with ECHO_ and matching the label.  For example if I set ECHO_version=v1 I can verify the routing worked as expected.

Third it introduces a ?codes= query parameter.  The format is int[,int]*.  This causes the server to respond with HTTP error codes other than 200 if supplied.  For example, if I supply "?codes=500,200" in the URL then the output will cycle between 500 and 200.  This is useful for verifying HTTP retries *but only when the number of instances is 1*.


test/server/echo.go
```diff
@@ -53,20 +56,40 @@ type handler struct {
...
+
+	hostname, err := os.Hostname()
+	if err == nil {
+		body.WriteString(fmt.Sprintf("Server instance hostname=%v\n", hostname))
+	}
+
+    for _, envExpr := range os.Environ() {
+    	if strings.HasPrefix(envExpr, "ECHO") {
+			body.WriteString(fmt.Sprintf("Server environment variable %v\n", envExpr))
+		}
```
kyessenov: There is a flag called version that tries to do the same thing. Should we just use that one or there is some advantage?

test/server/echo.go
```diff
@@ -53,20 +56,40 @@ type handler struct {
...
 	for name, headers := range r.Header {
 		for _, h := range headers {
 			body.WriteString(fmt.Sprintf("%v=%v\n", name, h))
 		}
 	}
+
+	hostname, err := os.Hostname()
+	if err == nil {
+		body.WriteString(fmt.Sprintf("Server instance hostname=%v\n", hostname))
+	}
```
kyessenov: Can you make prefix to be a single word like Hostname?

test/server/echo.go
```diff
@@ -120,3 +143,61 @@ func main() {
...
+	if codestrings == "" {
+		// Consider no codes to be a repeating sequence of 200,
+		codestrings = strconv.Itoa(http.StatusOK)
+	}
+
+	aCodestrings := strings.Split(codestrings, ",")
+	codes := make([]int, len(aCodestrings), len(aCodestrings))
+
+	for i, codestring := range aCodestrings {
+		code, err := validateCode(codestring)
```
kyessenov: I think it would be simpler to maintain a global request count and then just do a modulus in the request array.

test/server/echo.go
```diff
@@ -39,6 +39,9 @@ var (
 	ports     []int
 	grpcPorts []int
 	version   string
+
+	// Record how many attempts have been made using a particular sequence of codes
+	state map[string]int = make(map[string]int)
```
ayj: The type declaration isn't needed.
```golang
state = make(map[string]int)
```

test/server/echo.go
```diff
@@ -53,20 +56,40 @@ type handler struct {
 
 func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	body := bytes.Buffer{}
+	r.ParseForm()
+
+	// If the request has form ?codes=int[,int]* return those codes, in sequence, rather than 200
+	err := setResponseFromCodes(r, w)
```
ayj: Use an initialization statement here to limit scope of `err`.
```golang
if err := setResponseFromCodes(r, w); err != nil {
    body.WriteString("codes error: " + err.Error() + "\n")
}
```

test/server/echo.go
```diff
@@ -53,20 +56,40 @@ type handler struct {
...
 	body.WriteString("RemoteAddr=" + r.RemoteAddr + "\n")
 	body.WriteString("Host=" + r.Host + "\n")
+
 	for name, headers := range r.Header {
 		for _, h := range headers {
 			body.WriteString(fmt.Sprintf("%v=%v\n", name, h))
 		}
 	}
+
+	hostname, err := os.Hostname()
```
ayj: Use an initialization statement here to limit scope of `hostname`.

```golang
if hostname, err := os.Hostname(); err == nil {
    body.WriteString(fmt.Sprintf("Server instance hostname=%v\n", hostname))
}
```

test/server/echo.go
```diff
@@ -120,3 +143,61 @@ func main() {
...
+func setResponseFromCodes(request *http.Request, response http.ResponseWriter) error {
+	responseCodes := request.FormValue("codes");
+
+	codes, err := validateCodes(responseCodes)
+	if err != nil {
+		return err
+	}
+
+	// Keep a cursor for each sequence of response codes
+	position, ok := state[responseCodes]
```
ayj: Looks like `state` should be protected with sync.Mutex since ServeHTTP can be invoked from multiple goroutines.

test/server/echo.go
rshriram: I agree. The version flag serves the same purpose.

test/server/echo.go
```diff
@@ -120,3 +143,61 @@ func main() {
...
+
+// code must be HTTP response code
+func validateCode(code string) (int, error) {
+	n, err := strconv.Atoi(code)
+	if err != nil {
+		return n, err
+	}
+
+	if n < 200 || n > 600 {
+		return 400, fmt.Errorf("invalid HTTP response code %v", n)
```
ayj: s/400/http.StatusBadRequest

test/server/echo.go
```diff
@@ -120,3 +143,61 @@ func main() {
...
+}
+
+// code must be HTTP response code
+func validateCode(code string) (int, error) {
+	n, err := strconv.Atoi(code)
+	if err != nil {
+		return n, err
+	}
+
+	if n < 200 || n > 600 {
```
ayj: n >= 600?

test/server/echo.go
esnible: I found using a map gave me a lot of flexibility.

When writing test cases against a single instance of a server I could be assured if I started a new sequence of codes I would be starting at 0.

On a previous project I spent a lot of time verifying fencepost errors.  If my server generates 17 errors, and I retry 17 times, am I guaranteed to fail?  It is easy to make that happen if each test uses a unique sequence like "501,500,500,500,200" and "502,500,500,500,500,200".  Is it worth giving that up to use an int instead of a map?

test/server/echo.go
esnible: I didn't want to add a dependency because the output of this server when invoked by multiple callers or when implemented with multiple instances is not deterministic.  Therefore it doesn't matter if state gets corrupt.

If you disagree let me know and I will add the mutex.

esnible:
>>This is useful for verifying HTTP retries but only when the number of instances is 1.
>Is this by design or artifact of current implementation?

I felt a very simple server with a convention not to expect the "codes" responses to be deterministic was better than introducing a resource that would be shared across multiple pod instances to coordinate.  If you do not agree please suggest your preferred mechanism for sharing state across pods.

test/server/echo.go
ayj: Please add a lock. Behavior of concurrent read/write to a map is undefined behavior (see https://blog.golang.org/go-maps-in-action).

ayj:
It depends if you want the response code sequence applied globally or per source. You're already sharing state across pods with `state` map. This state could be extended to track sequence by RemoteAddr, e.g. `map[string]map[string][]int`.

test/server/echo.go
kyessenov: We cannot control the lifecycles of pods. I think assuming request 0 is consistent is unrealistic and will lead to flaky tests.

esnible:
But I'm *NOT* sharing the `state` variable.  Each pod has it's own copy, so to get consistent results the number of pods must be 1.

Tracking the RemoteAddr will not work for what I suspect will be the most common use case, manual interactive testing with _curl_.  Each _curl_ invocation will use a different local port.  Try curling the existing echo server and you will see the RemoteAddr be something like 127.0.0.1:50756 the first time and  127.0.0.1:50757 the second time.  If we desire to use the ?codes= feature in automated tests that run in parallel we should introduce a transaction ID header for those tests to supply.

costinm:
Maybe the codes could be specified as a flag ?

kyessenov:

esnible:
@costinm By specifying `codes` as part of the URL it allows me to keep a separate sequence for each URL.

For example, I can do

```
curl echo/uuid-test1?codes=500,500,200 # returns 500
curl echo/uuid-test2?codes=400,200 # returns 400
curl echo/uuid-test1?codes=500,500,200 # returns 500
curl echo/uuid-test2?codes=400,200 # returns 200
curl echo/uuid-test1?codes=500,500,200 # returns 200
```

I wrote this feature because popular testing servers like _httpbin_ have URLs that always return a particular error, but there are no servers with stable URLs that give me a particular error.  But I want to test things like different retry counts so it is desirable to get more than one deterministic sequence out of the same server instance.

esnible:

esnible:

test/server/echo.go
```diff
@@ -39,6 +40,9 @@ var (
 	ports     []int
 	grpcPorts []int
 	version   string
+
+	// Record how many attempts have been made using a particular sequence of codes
+	state = make(map[string]int)
```
kyessenov: Ed, I worry that expecting the pod to respond in an exact sequence is likely to lead to flaky tests. Kubernetes does not guarantee that pods are not killed/restarted in a long test. This is more true for our test cluster which gets overloaded sometimes and fails to schedule pods on time.

PR #594, opened 2017-04-26
Add k8s auth plugins

kyessenov:
Add "gcp" and "oidc" auth plugins. Verified that "gcp" plugin works for GKE. We still want to use client certs for testing since bazel does not allow invoking random commands from tests.


kyessenov:

kyessenov:

PR #595, opened 2017-04-26
Add gdb, curl and vim to debug image to chase envoy crashes

andraxylia:


rshriram:
If you are going to put vim, then can you please put nano in the image? Its the simplest notepad style editor that you can find, for linux. (is it too much to ask for emacs in the debug image?)

andraxylia:
Added vim and nano, removed the second curl.

PR #597, opened 2017-04-26
istioctl: auto-generate markdown and bash completion files.

ayj:
Auto-generate reference documentation for istioctl commands. This will be
used to auto-generate raw reference content for
istio.io. Additional post-processing is needed to get files into the
proper formatting expected by the istio.io jekyll templates. That job
will be handled by a separate script in istio/istio.github.io repo.

  istioctl collateral markdown -dir <dir>

Generate bash completion for istioctl commands, subcommands, and
flags.

  source <(istioctl collateral completion)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -253,6 +253,24 @@ new_go_repository(
 )
 
 new_go_repository(
+    name = "com_github_cpuguy83_go_md2man",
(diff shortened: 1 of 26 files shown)


andraxylia:
Maybe it's just me... how about ancillary/supplementary as a synonym for collateral?

Issue #214, opened 2017-02-27
Ingress controller: update ingress resource status


zcahana:
Ingress resources have a status field, which indicates the external IP addresses/hostname for which the ingress is mapped to. This is normally updated by the ingress controller itself, and is set to the IP address(es) of the node(s) in which the ingress controller(s) run, or the external IP of a LoadBalancer-type service which exposes the ingress controller.

This can be implemented using the [`github.com/kubernetes/ingress/core/pkg/ingress/status`](https://github.com/kubernetes/ingress/tree/master/core/pkg/ingress/status) package, which already does something similar.

zcahana:
Once this issue is resolved, we should change the bookinfo demo so that the the application is accessed via the IP address assigned to the ingress resource (as returned by `kubectl get ingress <ingress_name>`), and not via the LoadBalancer IP and/or hostIP:NodePort of the ingress controller service.

kyessenov:
It's not fully done since the syncer sets the wrong IP.

kyessenov:
Tested with GKE, the ingress status IP is a list of node IPs (3 out of my 5 nodes). Trying to access a node IP on port 80 does not work.

ijsnellf:
This can be closed, right? The syncer either updates the IPs to the node IPs or the associated service IP.

PR #605, opened 2017-04-27
Ingress status publish service

ijsnellf:
Should enable the ingress status syncer to obtain the load balancer IP address.

### DIFF
--- a/platform/kube/ingressstatus.go
+++ b/platform/kube/ingressstatus.go
@@ -15,6 +15,8 @@
 package kube
 
 import (
+	"fmt"
+
 	"istio.io/api/proxy/v1/config"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


platform/kube/ingressstatus.go
```diff
@@ -57,11 +59,16 @@ func NewIngressStatusSyncer(mesh *config.ProxyMeshConfig, client *Client,
 		&v1beta1.Ingress{}, options.ResyncPeriod, cache.Indexers{},
 	)
 
+	var publishService string
+	if mesh.IngressService != "" {
```
kyessenov: i think it's better not to run status syncer when the field is not defined, WDYT?

ijsnellf: If this field isn't defined the status syncer will update the ingress resource with the node IP. This is somewhat useful if you are using a NodePort setup, as in Minikube. Though in that case you'll still need to use the 30000+ port.

ijsnellf:

platform/kube/ingressstatus.go
kyessenov: ok, fair enough.

ijsnellf:
Contributes to #214.

PR #610, opened 2017-04-27
e2e ingress status IP check

ijsnellf:
Ensures ingress status is given an IP. Since we are running without `mesh.ingressService` set, the IP will be the node IP.


ijsnellf:
Contributes to #214.

ijsnellf:
The status takes a bit of time to sync, so this typically tries 10-15 times before it succeeds.

Issue #601, opened 2017-04-26
compress cluster names


kyessenov:
envoy set a hard limit of 60 characters to the cluster name which must be unique per proxy. a typical cluster name looks like this:
```
outbound:world.default.svc.cluster.local:http:version=v0
```
We need to update config gen to hash these names. This will likely make stats difficult to interpret without some reverse mapping.

@costinm (since you're a heavy user of stats)

PR #613, opened 2017-04-27
Rebase envoy and use sha-1 cluster names

kyessenov:


kyessenov:
fixes #601

kyessenov:
Not necessary mostly since "service key" in each cluster contains the inverse.

PR #614, opened 2017-04-27
changed from double to duration

GregHanson:


rshriram:
Need to merge PR in istio/API first before merging this.

kyessenov:
I'll hold on to my API change till you merge yours first.

proxy/envoy/testdata/cb-policy.yaml.golden
```diff
@@ -2,10 +2,10 @@ destination: world.default.svc.cluster.local
 circuit_breaker:
   simple_cb:
     max_connections: 100
-    sleep_window_seconds: 15.5
+    sleep_window: 15.5s
```
rkpagadala: Any reason why this is specified in seconds vs ms? is dest-policy.yaml we use 15500ms for the duration.

kyessenov: i think either is fine with protobuf. i don't think this config is actually tested end-to-end

proxy/envoy/policy.go
```diff
@@ -105,18 +106,28 @@ func insertDestinationPolicy(config *model.IstioRegistry, cluster *Cluster) {
...
+				cluster.OutlierDetection.IntervalMS = protoDurationToMS(cbconfig.HttpDetectionInterval)
 			}
 			if cbconfig.HttpMaxEjectionPercent > 0 {
 				cluster.OutlierDetection.MaxEjectionPercent = int(cbconfig.HttpMaxEjectionPercent)
 			}
 		}
 	}
 }
+
+func protoDurationToMS(dur *duration.Duration) int {
```
kyessenov: please use convertDuration to avoid manipulating time directly. You can divide by unit to get MS.

rkpagadala: Could be rewritten to something like
func protoDurationToMS(dur *duration.Duration) int {
	var d time.Duration = convertDuration(dur)
	ms := int(d / time.Millisecond)
	return ms
}

proxy/envoy/testdata/cb-policy.yaml.golden
GregHanson: @rkpagadala these are used for testing the envoy config generation.  I figured it would be useful to test one case using `15.5s` and another `15500ms` to verify the desired config is achieved with both formats.

geeknoid:
camelCase all the way...

model/validation.go
```diff
@@ -509,24 +515,35 @@ func ValidateLoadBalancing(lb *proxyconfig.LoadBalancing) (errs error) {
 func ValidateCircuitBreaker(cb *proxyconfig.CircuitBreaker) (errs error) {
 	if simple := cb.GetSimpleCb(); simple != nil {
 		if simple.MaxConnections < 0 {
-			errs = multierror.Append(errs, fmt.Errorf("circuit_breaker max_connections must be in range [0..]"))
+			errs = multierror.Append(errs,
+				fmt.Errorf("circuit_breaker max_connections must be in range [0..]"))
```
kyessenov: Can you please rename fields to follow camelCase, e.g. maxConnections here? Proto allows both versions but we are trying to standardize on camelCase.

model/validation.go
```diff
@@ -310,8 +312,9 @@ func ValidateDestinationWeight(dw *proxyconfig.DestinationWeight) (errs error) {
 // ValidateHTTPTimeout validates HTTP Timeout
 func ValidateHTTPTimeout(timeout *proxyconfig.HTTPTimeout) error {
 	if simple := timeout.GetSimpleTimeout(); simple != nil {
-		if simple.TimeoutSeconds < 0 {
-			return fmt.Errorf("timeout_seconds must be in range [0..]")
+		err := validateDuration(simple.Timeout)
+		if err != nil {
+			return fmt.Errorf("timeout must be in range [0..]")
```
kyessenov: you can use multierror.Prefix to add field name to err returned from validateDuration. See examples in the code.

model/validation.go
```diff
@@ -731,12 +782,8 @@ func ValidateProxyMeshConfig(mesh *proxyconfig.ProxyMeshConfig) (errs error) {
 		errs = multierror.Append(errs, errors.New("Istio service cluster must be set"))
 	}
 
-	if err := validateDuration(mesh.DrainDuration); err != nil {
-		errs = multierror.Append(errs, multierror.Prefix(err, "invalid drain duration:"))
-	}
-
-	if err := validateDuration(mesh.ParentShutdownDuration); err != nil {
-		errs = multierror.Append(errs, multierror.Prefix(err, "invalid parent shutdown duration:"))
+	if err := validateParentAndDrain(mesh.DrainDuration, mesh.ParentShutdownDuration); err != nil {
```
kyessenov: Nice, thanks.

model/validation.go
GregHanson: nice catch, looks like I made the same mistake a few times.  fixed other occurrences as well

model/validation.go
GregHanson: updated all error messages to use camelCase

kyessenov:
The race is on!

PR #626, opened 2017-04-28
Kill all proxies on error exit status

kyessenov:


proxy/envoy/watcher.go
```diff
@@ -115,51 +114,67 @@ func configFile(config string, epoch int) string {
...
-
-func cleanupEnvoy(mesh *proxyconfig.ProxyMeshConfig) func(int) {
-	return func(epoch int) {
-		path := configFile(ConfigPath, epoch)
-		if err := os.Remove(path); err != nil {
-			glog.Warningf("Failed to delete config file %s for %d, %v", path, epoch, err)
-		}
+func runEnvoy(mesh *proxyconfig.ProxyMeshConfig, node string) proxy.Proxy {
+	return proxy.Proxy{
+		Run: func(config interface{}, epoch int, abort <-chan error) error {
```
ayj: Where does Run() return errAbort? It see it handled in various places and returned in the test case but am missing it in the actual runtime code.

kyessenov: It returns whatever is passed on the abort channel. Agent sends and discards abort to avoid infinite loop.

PR #632, opened 2017-05-01
New ?codes= option to simulate HTTP errors from echo server

esnible:
This is a re-write of the ?codes= injection feature from last week.

The echo server output includes its hostname
If a request to the echo server HTTP includes ?codes=code[:chances],... the server might send a response other than HTTP 200.  This is for testing code that retries, but where we don't need to track the exact number of retries.

For example, ?codes=500:90,200:10 returns 500 90% of times and 200 10% of times
For example, ?codes=500:1,200:1 returns 500 50% of times and 200 50% of times
For example, ?codes=501:999,401:1 returns 500 99.9% of times and 401 0.1% of times.
For example, ?codes=500,200 returns 500 50% of times and 200 50% of times

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,3 +1,15 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_prefix")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_prefix")
 
 go_prefix("istio.io/manager")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["config.pb.go"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "@com_github_golang_protobuf//proto:go_default_library",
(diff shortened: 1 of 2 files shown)


BUILD
```diff
@@ -1,3 +1,15 @@
-load("@io_bazel_rules_go//go:def.bzl", "go_prefix")
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_prefix")
 
 go_prefix("istio.io/manager")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["config.pb.go"],
```
kyessenov: this seems unnecessary? there is no config.pb.go file around

esnible: Because I needed to add math/rand I ran `bin/gazelle` and it did this.  I will try removing it and see if everything still tests.

PR #633, opened 2017-05-01
Proxy agent: panic after retries, broadcast abort, minor fixes

### LINKED ISSUE #400: Istio-proxy should exit when envoy cannot be started or restart fails
This is to recover by letting kubelet restart the istio-proxy pod.

kyessenov:
fixes #400

### DIFF
--- a/proxy/agent.go
+++ b/proxy/agent.go
@@ -125,9 +125,9 @@ type Proxy struct {
 	// Cleanup command for an epoch
 	Cleanup func(int)
 
-	// Panic command is invoked when all retries to start the proxy fail
-	// just before the agent terminating
-	Panic func()
+	// Panic command is invoked with the desired config when all retries to
+	// start the proxy fail just before the agent terminating
+	Panic func(interface{})
 }
 
 type agent struct {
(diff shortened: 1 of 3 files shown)


proxy/agent.go
```diff
@@ -99,28 +97,37 @@ type Retry struct {
...
+	// Run command for a config, epoch, and abort channel
 	Run func(interface{}, int, <-chan error) error
 
-	// Cleanup command
+	// Cleanup command for an epoch
 	Cleanup func(int)
+
+	// Panic command is invoked when all retries to start the proxy fail
+	// just before the agent terminating
+	Panic func()
```
andraxylia: Can we give a different name than Panic? It seems like a very controller shut down of all epochs.

kyessenov: I think we need to wave a red flag here since it's a very bad error case. Any proposal?

andraxylia: KillThemAll

kyessenov: Ah, but we don't kill all proxies here. We just exit the process and let panic also panic. I guess we should kill all proxies before panicking?

ayj: Can you pass the current (bad?) configuration to Panic so it can be logged? Would be useful to debug / triage later.

andraxylia: +1

kyessenov: added as an argument. the config is large and is printed ten times before panic, maybe let's not do that in sidecar yet.

ayj: 10 times? Shouldn't Panic() only be called once?

kyessenov: There are 10 retries before panic. Each retry prints out the whole config to the log.

PR #636, opened 2017-05-02
Update config proto for destination policy

kyessenov:
Needs a follow-up to validate destination policy name against the destination field value.
As-is sufficient to pass the tests with no functional change.

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -109,7 +109,7 @@ const (
 	// RouteRule defines the kind for the route rule configuration
 	RouteRule = "route-rule"
 	// RouteRuleProto message name
-	RouteRuleProto = "istio.proxy.v1alpha.config.RouteRule"
+	RouteRuleProto = "istio.proxy.v1.config.RouteRule"
 
 	// IngressRule kind
 	IngressRule = "ingress-rule"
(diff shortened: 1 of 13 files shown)


kyessenov:
@sebastienvas we need to do `bazel clean` to cleanup genfiles. don't know why bazel doesn't do that automatically.

kyessenov:

model/config.go
```diff
@@ -234,15 +234,16 @@ func (i *IstioRegistry) PoliciesByNamespace(namespace string) []*proxyconfig.Des
 	return out
 }
 
-// DestinationPolicies lists all policies for a service version.
-// Policies are not inherited by tags inclusion. The policy tags must match the tags precisely.
-func (i *IstioRegistry) DestinationPolicies(destination string, tags Tags) []*proxyconfig.DestinationPolicy {
-	out := make([]*proxyconfig.DestinationPolicy, 0)
+// DestinationPolicy returns a policy for a service version.
```
rshriram: Can you make a note here about renaming this to Destination Rule?

rshriram:
Feel free to merge if you think these need to be handled elsewhere.

model/config.go
kyessenov: one thing at a time...

kyessenov:
I found only single use of policy in this repo. The semantics is that there should be only one dest-policy object per destination. I'll have to deal with naming YAMLs appropriate in a next PR.
For example:
```
destination: a.default.svc.cluster.local
policy:...
```
should a dest-policy named "a" in "default" namespace.

kyessenov:

PR #642, opened 2017-05-02
Zipkin support

ijsnellf:
Add Zipkin tracing support.

TODO:
1. ~~Needs changes to `istio/api` to add ZipkinAddress field to the mesh.~~
1. Envoy Zipkin PR needs to be merged.
1. ~~Add a Zipkin integration test that makes requests and ensures that they show up in Zipkin via Zipkin's REST API.~~
1. Set Zipkin trace context from mixer attribute values.

edit:
2 and 4 don't need to hold up this PR.


kyessenov:
Nice, I would add:
4. set zipkin trace context from mixer attribute values (since service-cluster in Istio envoy is meaningless)

ijsnellf:
@kyessenov: What would that entail?

kyessenov:
Mixer is supposed to return service membership information back to proxy after check call. We need to modify mixer filter in the proxy to set the zipkin metadata right after the check call. For the initial cut, let's use service-node (which is pod IP), then iterate to get more useful information out of traces.

rshriram:
@kyessenov I am working on that. We (me and @fabolive) have devised a clean solution by which the mixer can plug in the service name into Zipkin spans. As such, you cannot use the service-node (envoy will insist on using service-cluster). But from our past experiments, this doesn't affect the demo either. We were able to see end to end traces, with individual service labels in the trace. When my PR and fabio's last PR goes into Envoy, we can change the mixer (its a two line change) to overwrite the service cluster to the service name, on a per API call basis.

WORKSPACE
```diff
@@ -296,12 +302,12 @@ PROXY = "67a24c284d9e75ae957833b1c33548f94a28f647"  # May 1, 2017
 
 http_file(
     name = "istio_proxy",
-    url = "https://storage.googleapis.com/istio-build/proxy/envoy-alpha-" + PROXY + ".tar.gz",
+    url = "https://github.com/amalgam8/envoy/raw/envoy-bin/envoy-binary/envoy-alpha-5.tar.gz",
 )
```
rshriram: @ijsnellf make sure to undo this before finalizing the PR.

WORKSPACE
```diff
@@ -296,12 +302,12 @@ PROXY = "67a24c284d9e75ae957833b1c33548f94a28f647"  # May 1, 2017
...
 
 http_file(
     name = "istio_proxy_debug",
-    url = "https://storage.googleapis.com/istio-build/proxy/envoy-debug-" + PROXY + ".tar.gz",
+    url = "https://github.com/amalgam8/envoy/raw/envoy-bin/envoy-binary/envoy-debug-5.tar.gz",
 )
```
rshriram: And this

ijsnellf:
This PR works with a Envoy proxy without Zipkin support as long as Zipkin is disabled in the manager. I'll switch back to the normal Envoy proxy and disable Zipkin so we can merge the PR.

WORKSPACE
ijsnellf: Yep. Done.

ijsnellf:

ijsnellf:
@rshriram @kyessenov: PTAL

Smoketest is failing, but it doesn't look like this PR is causing it.

test/integration/testdata/zipkin.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: zipkin
+  labels:
```
rshriram: question: will this PR work if the user names the zipkin service as anything else? like istio-zipkin etc?

PR #651, opened 2017-05-03
Fix 'panic: shorthand redefinition' for istioctl mixer subcommand

ayj:
Immediate fix is to remove mixer `-m` short flag option to avoid
conflict with the manager API server address. Slightly longer term fix
is to fetch manager/mixer API server address from a config map instead
of depending on environment variables or flags.


ayj:
This fixes the following panic:

```
$ /home/vagrant/go/bin/istioctl mixer
panic: shorthand redefinition

goroutine 1 [running]:
istio.io/manager/vendor/github.com/spf13/pflag.(*FlagSet).AddFlag(0xc42008fd40, 0xc420216e60)
```

PR #652, opened 2017-05-03
Add logging of HTTP traffic

esnible:
Add logging of HTTP traffic from istioctl to the Discovery Manager

### DIFF
--- a/client/proxy/proxy.go
+++ b/client/proxy/proxy.go
@@ -45,14 +45,14 @@ func NewManagerClient(base url.URL, apiVersion string, client *http.Client) *Man
 
 func (m *ManagerClient) toCurl(request *http.Request, body string) string {
 
-	headers := ""
+	var headers string
 	for key, values := range request.Header {
 		for _, value := range values {
 			headers += fmt.Sprintf(` -H %q`, fmt.Sprintf("%s: %s", key, value))
 		}
 	}
 
-	bodyOption := ""
+	var bodyOption string
 	if body != "" {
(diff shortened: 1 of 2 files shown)


client/proxy/proxy.go
```diff
@@ -41,6 +43,23 @@ func NewManagerClient(base url.URL, apiVersion string, client *http.Client) *Man
 	}
 }
 
+func (m *ManagerClient) toCurl(request *http.Request, body string) string {
+
+	headers := ""
```
ayj: (nit): `var headers string`

client/proxy/proxy.go
```diff
@@ -41,6 +43,23 @@ func NewManagerClient(base url.URL, apiVersion string, client *http.Client) *Man
...
+	for key, values := range request.Header {
+		for _, value := range values {
+			headers += fmt.Sprintf(` -H %q`, fmt.Sprintf("%s: %s", key, value))
+		}
+	}
+
+	bodyOption := ""
```
ayj: (nit): `var bodyOption string`

client/proxy/proxy.go
```diff
@@ -161,5 +184,11 @@ func (m *ManagerClient) doConfigCRUD(key model.Key, method string, inBody []byte
 	if err != nil {
 		return nil, err
 	}
-	return m.do(request)
+
+	response, err := m.do(request)
+
+	// Log after the call to m.do() so that the full hostname is present
```
ayj: (optional)

```golang
// Log after the call to m.do() so that the full hostname is present
defer glog.V(2).Infof("%s", m.toCurl(request, string(inBody)))
return m.do(request)
```

PR #661, opened 2017-05-04
Combine ingress paths and route rules

kyessenov:
The goal is to enable route rules on ingress rule destinations. I had to make concessions since it is not trivial to combine two Envoy HTTP routes. The ingress routes have the following restrictions:

1. no weighted destination - can be added later by multiplying weights from ingress and destination rule and balancing them to 100.
2. no header matching in ingress - it is not trivial to combine two sets of match conditions on headers, once they utilize regexes.
3. no additional tags on destinations in ingress - we would need to apply the tags to destinations on destination rules, also non-trivial

These restrictions are not serious since we do not generate ingress rules that violate any of these from the ingress controller. 

Additionally, I cleaned up some of the logic related to header matching: quoting prefix in a regex, detecting if a route is a catch-all, resolve service ports during route generation rather than rule generation.

Added a test to combine ingress with a route rule on "c".


proxy/envoy/config.go
```diff
@@ -227,6 +226,61 @@ func buildOutboundListeners(instances []*model.ServiceInstance, services []*mode
...
+				// When such a route is encountered, we stop building further routes for the
+				// destination and we will not add the default route after the for loop.
+				if httpRoute.CatchAll() {
+					useDefaultRoute = false
+					break
+				}
+			}
+		}
+
+		if useDefaultRoute {
```
rshriram: I think we also need to add rules with a single route (same destination, 100 weight) as a catch all route.
examples of catch all routes:
1. rule which has prefix /, no http headers and no path
2. rule which has no match, with single route (weight 100)
3. any thing else @frankbu ??

proxy/envoy/resources.go
```diff
@@ -175,12 +186,17 @@ type HTTPRoute struct {
 	faults []*HTTPFilter
 }
 
+// CatchAll returns true if the route matches all requests
+func (route *HTTPRoute) CatchAll() bool {
+	return len(route.Headers) == 0 && route.Path == "" && route.Prefix == "/"
+}
```
rshriram: if my http route rule does not provide path/prefix/headers, will route.Prefix still be set to / ?

proxy/envoy/testdata/rds-ingress-weighted.json.golden
```diff
@@ -0,0 +1,28 @@
...
+      "weighted_clusters": {
+       "clusters": [
+        {
+         "name": "out.c76febe0f151b2f8abe0f377d2052c0fbbfb959d",
+         "weight": 75
+        },
+        {
+         "name": "out.66fcc955b8875b19844f9eaf6cfda47c778c609e",
+         "weight": 25
+        }
```
rshriram: nice work!

rshriram: How do we match headers as well here?

proxy/envoy/resources.go
```diff
@@ -47,22 +54,27 @@ const (
...
 func convertDuration(d *duration.Duration) time.Duration {
 	dur, err := ptypes.Duration(d)
 	if err != nil {
-		glog.Warningf("error converting duration %#v: %v", d, err)
+		glog.Warningf("error converting duration %#v, using 0: %v", d, err)
 	}
 	return dur
 }
 
+func protoDurationToMS(dur *duration.Duration) int64 {
```
ijsnellf: :+1:

proxy/envoy/config.go
kyessenov: The logic is same as it was before. I just moved catchAll to a function to simplify the control flow. Is there a specific example when the default rule is added or not added?

proxy/envoy/resources.go
kyessenov: The function assumes the route is valid (one of path or prefix is set). It looks like we set path and prefix in every route soon after instantiation. It's probably a good idea to move it to a factory method so that no invalid routes are created.

proxy/envoy/testdata/rds-ingress-weighted.json.golden
kyessenov: If there is a route rule on destination that uses headers, there should be two routes - one with header match, and another for catch-all. Both should be added, as long as they are compatible.

kyessenov:

proxy/envoy/config.go
```diff
@@ -246,99 +300,38 @@ func buildOutboundHTTPRoutes(
...
+			}
 
-				if !catchAll {
-					// default route for the destination
-					cluster := buildOutboundCluster(service.Hostname, servicePort, nil)
-					routes = append(routes, buildDefaultRoute(cluster))
-				}
+			routes := buildDestinationHTTPRoutes(service, servicePort, rules)
 
+			if len(routes) > 0 {
```
ayj: (optional): invert conditional check and continue to reduce nesting depth

proxy/envoy/header.go
```diff
@@ -18,20 +18,46 @@ package envoy
...
 	"github.com/golang/glog"
 
 	proxyconfig "istio.io/api/proxy/v1/config"
+	"istio.io/manager/model"
 )
 
+func buildURIPathPrefix(matches *proxyconfig.MatchCondition) (path string, prefix string) {
+	path = ""
+	prefix = "/"
+	if matches != nil {
```
ayj: (optional): invert conditional check and continue to reduce nesting depth

kyessenov:

kyessenov:
I'll follow up with a smaller PR to address some header/and code style things I noticed.

kyessenov:

kyessenov:

kyessenov:

kyessenov:

PR #663, opened 2017-05-04
Restart sidecar proxy when Istio auth secret is changed

lookuptable:


proxy/envoy/watcher.go
```diff
@@ -88,14 +91,44 @@ func NewWatcher(ctl model.Controller, context *proxy.Context) (Watcher, error) {
...
+	fw, err := fsnotify.NewWatcher()
+	if err != nil {
+		glog.Warning("failed to create a watcher for certificate files")
+	}
+	defer fw.Close()
+
+	certsDir := w.context.MeshConfig.GetAuthCertsPath()
+	fw.Watch(certsDir)
+
+	go func() {
```
ayj: The stop and fw.Event channel can be handled in the same loop with select, i.e.

```golang
for {
    select {
        case <-fw.Event:
			glog.V(2).Infof("Change to %q is detected, reload the proxy if necessary", certsDir)
     		w.reload()
        case <-stop:
            glog.V(2).Info("Certificate watcher is terminated")
            return
    }
}
```

lookuptable:
@rshriram this is more like a stop gap solution to make sure the rotated certs are picked up by the proxies. Eventually we want Envoy to reload certs without actually restart itself. @PiotrSikora is helping us to drive the envoy effort.

ayj:
@rshriram, are you concerned with proxy restart races or system-wide load on RDS/CDS/SDS when all proxies in the network restart? The latter issue can be mitigated by distributing proxy reload over time via randomized delay between cert-change detection and proxy reload.

proxy/envoy/watcher.go
```diff
@@ -102,23 +102,28 @@ func (w *watcher) watchCerts(stop <-chan struct{}) {
 	if err != nil {
 		glog.Warning("failed to create a watcher for certificate files")
 	}
-	defer fw.Close()
+	defer func() {
```
ayj: Flow cannot proceed to this point if NewWatcher returned non-nil error. The function should at least return though there is no easy way to propagate errors back in these watchers. Possibly later on we could plumb through context.Context for that purpose, but that is outside the scope of this PR.

proxy/envoy/watcher.go
```diff
@@ -225,7 +230,9 @@ func generateCertHash(certsDir string) []byte {
 			glog.Warningf("failed to read file %q", filename)
 			continue
 		}
-		h.Write(bs)
+		if err := h.Write(bs); err != nil {
```
ayj: How bad is it if cert-read or hash-write fails here since that would likely indicate some other major problem in the system? In other words, should generateCertHash return an error in addition to hash?

lookuptable: I think we have two options: 1) log and ignore the error in `generateCertHash()` and the caller will try to schedule a config update. The update may fail (e.g. Envoy cannot read the certs). 2) return an error in `generateCertHash()`. The caller sees the error and stop config update and Envoy continues to use an old config/certs.

I'm leaning towards 1) to fail early.

lookuptable:
@kyessenov can you take another look?

proxy/envoy/ingress.go
```diff
@@ -137,14 +147,27 @@ func generateIngress(mesh *proxyconfig.ProxyMeshConfig, tls *model.TLSSecret, ce
 	}
 
 	config := buildConfig(listeners, nil, mesh)
+
+	h := sha256.New()
+	hashed := false
```
ayj: General question: should the cert/key writes here also be guarded with AuthPolicy==MUTAL_TLS check?

lookuptable: We need to write the hash in two cases for Ingress: 1) when `tls` is not nil, which defines the cert/key for external traffic and 2) when `AuthPolicy` is `MUTUAL_TLS` and this takes care of certs used for traffic between proxies.

proxy/envoy/cert.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"crypto/sha256"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/golang/glog"
+	"github.com/howeyc/fsnotify"
+)
+
+// watchCerts watches a certificate directory and calls the provided
```
kyessenov: nit: be explicit that the watcher does not watch subdirectories.

proxy/envoy/cert_test.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"bytes"
+	"crypto/sha256"
+	"io/ioutil"
+	"os"
+	"testing"
+	"time"
+)
+
+func TestWathCerts(t *testing.T) {
```
lookuptable: Done

proxy/envoy/watcher.go
```diff
@@ -88,6 +88,9 @@ func NewWatcher(ctl model.Controller, context *proxy.Context) (Watcher, error) {
 func (w *watcher) Run(stop <-chan struct{}) {
 	// must start consumer before producer
 	go w.agent.Run(stop)
+	if mesh := w.context.MeshConfig; mesh.AuthPolicy == proxyconfig.ProxyMeshConfig_MUTUAL_TLS {
+		go watchCerts(mesh.AuthCertsPath, stop, func() { w.reload() })
```
lookuptable: Done

proxy/envoy/ingress.go
```diff
@@ -137,14 +147,27 @@ func generateIngress(mesh *proxyconfig.ProxyMeshConfig, tls *model.TLSSecret, ce
...
 		if _, err := h.Write(tls.Certificate); err != nil {
 			glog.Warning(err)
 		}
 		if _, err := h.Write(tls.PrivateKey); err != nil {
 			glog.Warning(err)
 		}
+	}
+
+	if mesh.AuthPolicy == proxyconfig.ProxyMeshConfig_MUTUAL_TLS {
+		hashed = true
```
kyessenov: why not just take XOR of two hashes?

PR #673, opened 2017-05-05
Validate in apiserver to surface correct response code

liamawhite:
This will allow us to return 400 errors when validation fails

Fixes - https://github.com/istio/manager/issues/648#issuecomment-299529254


liamawhite:
This isn't a sensible place to validate as we lack the context of the other rules. We should use error typing in the registry to solve this issue instead.

Issue #660, opened 2017-05-04
k8s tests for istioctl + apiserver


esnible:
Currently _test/integration/testdata/manager.yaml.tmpl_ does not include the apiserver container.

Probably _test/integration/apiserver.go_ should be created which would make HTTP CRUD requests to verify the apiserver functions as intended when running on manager pod.

Probably a _test/integration/istioctl.go_ should be created that makes CRUD requests using _istioctl_ to verify that it can talk to the apiserver.

esnible:
This work item should include tests of GET config/route-rule/

esnible:
Closing because the REST-based apiserver tests are complete.

The e2e tests that use the istioctl binary will be tracked by https://github.com/istio/istio/issues/215

PR #676, opened 2017-05-05
e2e test for API server.

esnible:
First part of fix for #660 

driver.go will issue requests against apiserver and verify the responses are as expected.

Allow API server to run outside K8s with KUBECONFIG env.  Test template should include apiserver container

A future PR will add additional tests so istioctl create/get/put/delete is used as well.

driver.go now takes an argument to just run one test.  I have been testing with

```
./test/integration/integration --tag vagrant_20170504_192054 -hub docker.io/esnible \
   --testtype "apiserver" --auth disable
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -82,9 +82,6 @@ var (
 			}
 			if flags.controllerOptions.Namespace == "" {
 				flags.controllerOptions.Namespace = os.Getenv("POD_NAMESPACE")
-				if flags.controllerOptions.Namespace == "" {
-					return fmt.Errorf("export POD_NAMESPACE or supply --namespace flag")
-				}
 			}
 			glog.V(2).Infof("flags %s", spew.Sdump(flags))
 
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -58,6 +58,13 @@ var (
 		Short: "Istio Manager",
 		Long:  "Istio Manager provides management plane functionality to the Istio service mesh and Istio Mixer.",
 		PersistentPreRunE: func(*cobra.Command, []string) (err error) {
+			if flags.kubeconfig == "" {
+				if v := os.Getenv("KUBECONFIG"); v != "" {
+					glog.V(2).Infof("Setting configuration from KUBECONFIG environment variable")
+					flags.kubeconfig = v
+				}
+			}
+
 			client, err = kube.NewClient(flags.kubeconfig, model.IstioConfig)
 			if err != nil {
 				return multierror.Prefix(err, "failed to connect to Kubernetes API.")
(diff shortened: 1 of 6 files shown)


test/integration/apiserver.go
```diff
@@ -0,0 +1,359 @@
...
+	server := apiserver.NewAPI(apiserver.APIServiceOptions{
+		Version:  kube.IstioResourceVersion,
+		Port:     8081,
+		Registry: &model.IstioRegistry{ConfigRegistry: controller},
+	})
+	r.stopChannel = make(chan struct{})
+	go controller.Run(r.stopChannel)
+	go server.Run()
+
+	// Wait until apiserver is ready.  (As far as I can see there is no ready channel)
```
liamawhite: I'm writing a healthcheck soonish that this should be able to call

test/integration/apiserver.go
```diff
@@ -0,0 +1,359 @@
...
+			expectedBody:         jsonRule2,
+		},
+		// Can still GET after update
+		{
+			method: "GET", url: testURL,
+			expectedResponseCode: net_http.StatusOK,
+			expectedBody:         jsonRule2,
+			retryOn:              map[int]bool{net_http.StatusNotFound: true},
+			retryIfMatch:         []interface{}{jsonRule},
+		},
```
liamawhite: Add a second config then do a GET `config/route-rules/namespace` as well because it's a separate codepath

test/integration/apiserver.go
```diff
@@ -0,0 +1,359 @@
+// Copyright 2017 Istio Authors
```
liamawhite: It should probably be `test/integration/apiserver_test.go` just to ensure that it doesn't accidentally get pulled into a binary.

liamawhite: also maybe put it under `test/integration/kube/apiserver_test.go` because it is kube integration

cmd/manager/main.go
```diff
@@ -75,6 +82,9 @@ var (
 			}
 			if flags.controllerOptions.Namespace == "" {
 				flags.controllerOptions.Namespace = os.Getenv("POD_NAMESPACE")
+				if flags.controllerOptions.Namespace == "" {
```
kyessenov: namespace "" has a special meaning - it means all namespaces. why are we removing the option to listen on all namespaces?

esnible: This was intended to be guidance to the developer testing apiserver.  Perhaps it works inside the cluster.  From outside the cluster it was failing with an error message that confused me because the default namespace for _kubectl_ and _istioctl_ is "default" but the default namespace for _manager_ is "".

```
$ KUBECONFIG=~/.kube/config ./manager apiserver --namespace default
I0508 19:25:07.190941   32404 apiserver.go:96] Starting api at :8081
^C
$ KUBECONFIG=~/.kube/config ./manager apiserver --namespace ""
Error: failed to retrieve mesh configuration. an empty namespace may not be set when a resource name is provided
```

test/integration/apiserver.go
```diff
@@ -0,0 +1,408 @@
...
+			expectedResponseCode: net_http.StatusInternalServerError,
+		},
+		//  Appears in the namespace after creation
+		{
+			method: "GET", url: testNamespaceURL,
+			expectedResponseCode: net_http.StatusOK,
+			expectedBody:         []interface{}{jsonRule},
+			retryOn:              map[int]bool{net_http.StatusNotFound: true},
+			retryIfMatch:         []interface{}{[]interface{}{}},
+		},
```
liamawhite: Should also do a GET `config/route-rule/`

PR #681, opened 2017-05-07
Header validation

### LINKED ISSUE #662: HTTP header validation in route rules
There is no validation for http headers. I think we should decide upon the case sensitivity in the rules since the logic queries by lower-case versions (e.g. "uri", not "URI"). Also, block regexes on "uri" or improve the generation logic to handle it.

kyessenov:
fixes #662

### DIFF
--- a/model/validation.go
+++ b/model/validation.go
@@ -49,7 +49,7 @@ func IsDNS1123Label(value string) bool {
 	return len(value) <= dns1123LabelMaxLength && dns1123LabelRex.MatchString(value)
 }
 
-// ValidatePort ...
+// ValidatePort checks that the network port is in range
 func ValidatePort(port int) error {
 	if 1 <= port && port <= 65535 {
 		return nil
@@ -215,7 +215,7 @@ func (t Tags) Validate() error {
 	return errs
 }
 
-// ValidateFQDN ...
+// ValidateFQDN checks a fully-qualified domain name
 func ValidateFQDN(fqdn string) error {
 	if len(fqdn) > 255 {
 		return fmt.Errorf("domain name %q too long (max 255)", fqdn)
@@ -326,7 +326,7 @@ func ValidateL4MatchAttributes(ma *proxyconfig.L4MatchAttributes) (errs error) {
 	return
 }
 
-// ValidatePercent ...
+// ValidatePercent checks that percent is in range
 func ValidatePercent(val int32) error {
 	if val < 0 || val > 100 {
 		return fmt.Errorf("must be in range 0..100")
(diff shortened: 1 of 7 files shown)


model/validation.go
```diff
@@ -49,7 +49,8 @@ func IsDNS1123Label(value string) bool {
 	return len(value) <= dns1123LabelMaxLength && dns1123LabelRex.MatchString(value)
 }
 
-func validatePort(port int) error {
+// ValidatePort ...
+func ValidatePort(port int) error {
```
ayj: Why export all of Validate*() functions? I don't see them used outside of the model package. If this is prep work for a future CL then the function comments should be completed.

model/validation.go
```diff
@@ -256,46 +258,94 @@ func ValidateMatchCondition(mc *proxyconfig.MatchCondition) (errs error) {
...
+			errs = multierror.Append(errs, multierror.Prefix(err, fmt.Sprintf("header name %q invalid: ", name)))
+		}
+		if err := ValidateStringMatch(value); err != nil {
+			errs = multierror.Append(errs, multierror.Prefix(err, fmt.Sprintf("header %q value invalid: ", name)))
+		}
+
+		// validate special `uri` header:
+		// absolute path must be non-empty (https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.1.2)
+		if name == HeaderURI {
+			switch m := value.MatchType.(type) {
```
ayj: Should this switch be moved into ValidateStringMatch?

model/validation.go
kyessenov: We will need to export these functions to the common API repo?
I tried to normalize the code style and updated the names to be more consistent.

model/validation.go
kyessenov: StringMatch can be an empty string for other headers. StringMatch can also be used for non-headers (no examples though yet).

model/validation.go
ayj: Sounds good. Were you going to finish filling in the exported function comments in this CL or the CL that moves this all to istio/api?

kyessenov: I left out only dummy comments like "validateport validates network port". We can write lengthier comments if necessary but this keeps the linter happy.

ayj: Might be good to set precedent here and start with proper comments, otherwise it's more likely external contributions may copy existing incorrect style.

kyessenov: Yeah, good point. I borrowed this style from k8s ;) I think this `Validate` should just be an interface which explains why comments feel dummy; too bad we can't define them easily for proto-generated sources.

model/validation.go
ayj: I suppose `Exact` with empty string could be used to match empty headers, but when would it be valid to specify `Prefix` or `Regex` with no corresponding prefix or regex value?

kyessenov: Regex empty means "path empty", prefix empty means "any". This string match is not specific to headers.

PR #685, opened 2017-05-08
redirect routing rule for e2e

GregHanson:


test/integration/routing.go
```diff
@@ -174,3 +193,48 @@ func (t *routing) verifyFaultInjection(src, dst, headerKey, headerVal string,
...
 }
+
+// verifyRedirect verifies if the http redirect was setup properly
+func (t *routing) verifyRedirect(src, dst, targetHost, targetPath, headerKey, headerVal string, respCode int) error {
+	url := fmt.Sprintf("http://%s/%s", dst, src)
+	glog.Infof("Making 1 request (%s) from %s...\n", url, src)
+
+	resp := t.clientRequest(src, url, 1, fmt.Sprintf("-key %s -val %s", headerKey, headerVal))
+
+	statusCode := ""
```
ijsnellf: Don't need to convert the code to an integer:
`if len(resp.code) == 0 || resp.code[0] != fmt.Sprint(respCode) { ...return error... }`

test/integration/routing.go
```diff
@@ -174,3 +193,48 @@ func (t *routing) verifyFaultInjection(src, dst, headerKey, headerVal string,
...
+	}
+
+	if strconv.Itoa(respCode) != statusCode {
+		return fmt.Errorf("redirect verification failed: "+
+			"response status code: %v, expected %v",
+			statusCode, respCode)
+	}
+
+	exp := regexp.MustCompile("(?i)Host=(.*)")
+	hosts := exp.FindAllStringSubmatch(resp.body, -1)
```
ijsnellf: You should be able to make the logic a bit simpler since you only care about the first match:
https://golang.org/pkg/regexp/#Regexp.FindString
(or one of the ones right below it)

test/integration/routing.go
```diff
@@ -174,3 +193,48 @@ func (t *routing) verifyFaultInjection(src, dst, headerKey, headerVal string,
...
+		host = hosts[0][1]
+	}
+	if host != targetHost {
+		return fmt.Errorf("redirect verification failed: "+
+			"response body contains Hostname=%v, expected Hostname=%v",
+			hosts[0], targetHost)
+	}
+
+	exp = regexp.MustCompile("(?i)URL=(.*)")
+	paths := exp.FindAllStringSubmatch(resp.body, -1)
```
ijsnellf: Same comment as above

GregHanson: in this case however there are two occurrences of "URL" in the response, the original URL and the redirected URL

kyessenov:
please update the title to be about "redirect routing rule"

test/integration/routing.go
```diff
@@ -174,3 +193,43 @@ func (t *routing) verifyFaultInjection(src, dst, headerKey, headerVal string,
...
 	return nil
 }
+
+// verifyRedirect verifies if the http redirect was setup properly
+func (t *routing) verifyRedirect(src, dst, targetHost, targetPath, headerKey, headerVal string, respCode int) error {
+	url := fmt.Sprintf("http://%s/%s", dst, src)
+	glog.Infof("Making 1 request (%s) from %s...\n", url, src)
+
+	resp := t.clientRequest(src, url, 1, fmt.Sprintf("-key %s -val %s", headerKey, headerVal))
+
```
ijsnellf: super nit: remove newline

test/integration/routing.go
```diff
@@ -174,3 +193,43 @@ func (t *routing) verifyFaultInjection(src, dst, headerKey, headerVal string,
...
+	if len(resp.code) == 0 || resp.code[0] != fmt.Sprintf(respCode) {
+		return fmt.Errorf("redirect verification failed: "+
+			"response status code: %v, expected %v",
+			resp.code, respCode)
+	}
+
+	exp := regexp.MustCompile("(?i)Host=(.*)")
+	hosts := exp.FindAllStringSubmatch(resp.body, -1)
```
ijsnellf: something like this?
```golang
var host string
if matches := regexp.MustCompile("(?i)Host=(.*)").FindStringSubmatch(resp.body); len(matches) >= 2 {
   host = matches[1]
}
```

test/integration/routing.go
```diff
@@ -174,3 +193,43 @@ func (t *routing) verifyFaultInjection(src, dst, headerKey, headerVal string,
...
+		return fmt.Errorf("redirect verification failed: "+
+			"response body contains Host=%v, expected Host=%v",
+			host, targetHost)
+	}
+
+	exp = regexp.MustCompile("(?i)URL=(.*)")
+	paths := exp.FindAllStringSubmatch(resp.body, -1)
+	path := ""
+	if len(paths) > 1 {
+		path = paths[1][1]
```
ijsnellf: Potentially dangerous since `len(paths[1])` is never checked

test/integration/routing.go
```diff
@@ -174,3 +193,43 @@ func (t *routing) verifyFaultInjection(src, dst, headerKey, headerVal string,
...
+	}
+	if path != targetPath {
+		return fmt.Errorf("redirect verification failed: "+
+			"response body contains URL=%v, expected URL=%v",
+			paths[0], targetPath)
```
ijsnellf: meant to be...?
`path, targetPath)`

Could panic, because this is outside of the `len(paths) > 1` check

test/integration/routing.go
ijsnellf: Actually, because of the regex, `len(paths[x])` will always be 2, so check is unnecessary.

PR #692, opened 2017-05-09
istioctl: Fix path for Mixer Config server

ayj:
Manager and mixer API server paths are not consistent. Manager path is
prefixed with Istio resource version (i.e. v1alpha1) and mixer path is
not. Short term workaround is to special case this behavior. Long term
solution is to unify API scheme and server implementations.

See https://github.com/istio/api/issues/94 for long term tracking.


rshriram:
Good catch. Thanks for fixing this @ayj!

PR #702, opened 2017-05-09
Use proxy_debug image by default

andraxylia:


sebastienvas:
I need this change for the release as right now istioctl is using the proxy image not the proxy_debug. I ll finish 0.1.0 and will need to create a 0.1.1 in order to get debug proxy.

PR #703, opened 2017-05-09
Add capacity argument to make map/slice if it is known

rkpagadala:


mandarjog:
@rkpagadala did you find this thru linters or code inspections?

PR #709, opened 2017-05-10
Remove NodePort from Zipkin service definition in integration tests

ijsnellf:
Accidentally leftover from manual testing.


ijsnellf:
Doesn't seem to have been causing issues because it didn't specify a specific node port (and there is generally free ports on the node for it to use). But it makes bringing the infrastructure up and down slower.

ijsnellf:
@kyessenov: I think the "-" is a typo. `spec` isn't a list, so what I have now is valid.

PR #714, opened 2017-05-10
added validation for refreshDelay and connectTimeout

GregHanson:


model/validation.go
```diff
@@ -838,6 +838,55 @@ func ValidateParentAndDrain(drainTime, parentShutdown *duration.Duration) (errs
 				parentShutdownDuration.Seconds(), drainDuration.Seconds()))
 	}
 
+	if drainDuration > (time.Hour * 1) {
```
ijsnellf: nit: parenthesis not necessary

model/validation.go
```diff
@@ -838,6 +838,55 @@ func ValidateParentAndDrain(drainTime, parentShutdown *duration.Duration) (errs
...
+		errs = multierror.Append(errs,
+			fmt.Errorf("Istio drain time %v must be less than 1 hour", drainDuration.Minutes()))
```
ijsnellf: `.Seconds()`? Or include unit in error msg

model/validation.go
```diff
@@ -838,6 +838,55 @@ func ValidateParentAndDrain(drainTime, parentShutdown *duration.Duration) (errs
...
+	}
+
+	if parentShutdownDuration > (time.Hour * 1) {
+		errs = multierror.Append(errs,
+			fmt.Errorf("Istio parent shutdown time %v must be less than 1 hour",
+				parentShutdownDuration.Minutes()))
```
ijsnellf: see above

model/validation.go
ayj: General comment: can you use explicit `return err` throughout instead of named parameters? There is only one return value `errs` so I don't think the named part adds much.

model/validation.go
```diff
@@ -838,6 +838,55 @@ func ValidateParentAndDrain(drainTime, parentShutdown *duration.Duration) (errs
...
+	}
+	if errs != nil {
+		return
+	}
+
+	refreshDuration, _ := ptypes.Duration(refresh)
+	if refreshDuration > (time.Minute*10) || refreshDuration < time.Second {
+		errs = multierror.Append(errs,
+			fmt.Errorf("Istio discoveryRefreshDelay time %v must be >1 second and <10 minutes",
+				refreshDuration.Minutes()))
```
ijsnellf: see above

model/validation.go
```diff
@@ -838,6 +838,55 @@ func ValidateParentAndDrain(drainTime, parentShutdown *duration.Duration) (errs
...
+				parentShutdownDuration.Minutes()))
+	}
+
+	return
+}
+
+// ValidateRefreshDelay validates the discovery refresh delay time
+func ValidateRefreshDelay(refresh *duration.Duration) (errs error) {
+	if err := ValidateDuration(refresh); err != nil {
+		errs = multierror.Append(errs, err)
```
ayj: You can return the error directly here since `errs` should be empty.

model/validation.go
```diff
@@ -838,6 +838,55 @@ func ValidateParentAndDrain(drainTime, parentShutdown *duration.Duration) (errs
...
+func ValidateConnectTimeout(timeout *duration.Duration) (errs error) {
+	if err := ValidateDuration(timeout); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	if errs != nil {
+		return
+	}
+
+	timeoutDuration, _ := ptypes.Duration(timeout)
+	if timeoutDuration > (time.Second*30) || timeoutDuration <= (time.Millisecond) {
```
ijsnellf: nit: parenthesis unnecessary

model/validation.go
```diff
@@ -838,6 +838,55 @@ func ValidateParentAndDrain(drainTime, parentShutdown *duration.Duration) (errs
...
+// ValidateRefreshDelay validates the discovery refresh delay time
+func ValidateRefreshDelay(refresh *duration.Duration) (errs error) {
+	if err := ValidateDuration(refresh); err != nil {
+		errs = multierror.Append(errs, err)
+	}
+	if errs != nil {
+		return
+	}
+
+	refreshDuration, _ := ptypes.Duration(refresh)
```
ayj: Eventually goal is for these functions to be moved into istio/api. As such, it would be good to check the return error here instead of assuming the duration is a valid protobuf.

model/validation.go
ayj: Maybe define `const` for these, e.g. `MaxDrainDuration` and `MaxParentShutdownDuration`?

model/validation.go
```diff
@@ -880,11 +929,11 @@ func ValidateProxyMeshConfig(mesh *proxyconfig.ProxyMeshConfig) (errs error) {
 		errs = multierror.Append(errs, multierror.Prefix(err, "invalid parent and drain time combination"))
 	}
 
-	if err := ValidateDuration(mesh.DiscoveryRefreshDelay); err != nil {
+	if err := ValidateRefreshDelay(mesh.DiscoveryRefreshDelay); err != nil {
 		errs = multierror.Append(errs, multierror.Prefix(err, "invalid refresh delay:"))
 	}
 
-	if err := ValidateDuration(mesh.ConnectTimeout); err != nil {
+	if err := ValidateConnectTimeout(mesh.ConnectTimeout); err != nil {
```
ijsnellf: Generic function to validate a time is within a certain range?

model/validation.go
kyessenov: I think just `String()` is quite nice for durations.

ayj: [time.Duration](https://golang.org/src/time/time.go#L472) should print the proper units with String().

model/validation.go
GregHanson: @ayj the valid protobuf check is performed above and returns if an error is found - so only valid protobuf should make it to this point

model/validation.go
GregHanson: I agree with you @ayj, from a stylistic standpoint the only reason I went with the defined variable is because this was the way it was done throughout the rest of the file - so I opted for uniformity

GregHanson:

GregHanson:

GregHanson:

GregHanson:

PR #766, opened 2017-05-16
istioctl: compile in default hub and tag versions for kube-inject

ayj:
Introduce a new file 'kube-inject-versions` to set the hardcoded
default hub and tag info for kube-inject. Order of precedence is flag,
env-var, and then built-in defaults.

Existing documentation that use `source istio.VERSION` should continue
to work until istio.VERSION is phased out.

For debug purposes `istioctl version --kube-inject` shows the default
docker hub and tag. This flag is hidden from help text to avoid
confusing users as kube-inject is planned to be moved server-side
post-alpha.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,3 +1,5 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_prefix")
 
 go_prefix("istio.io/manager")
+
+exports_files(["kube-inject-versions"], visibility = ["//visibility:public"])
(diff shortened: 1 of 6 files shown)


ayj:
cc @sebastienvas, is this something you would want to hook-up to your release process?

andraxylia:
Thanks for adding this, and looking forward to remove the step "source istio.VERSION" part from the docs.

Issue #771, opened 2017-05-17
`istioctl delete` should not give up when deleting multiple rules on the first failure


esnible:
@frankbu says "Like _kubectl_, I think it should delete whatever it can, and print out messages about the ones that are not found"

To reproduce:

```
istioctl create -f route-rule-all-v1.yaml
istioctl delete route-rule foo productpage-default bar reviews-default
istioctl get route-rules
```

frankbu:
Another thought, in addition to this, it might be nice to add a `--all` option like kubectl
```
istioctl  delete route-rule --all
```

esnible:
When I want to delete all of the rules I do ```istioctl get route-rules | xargs istioctl delete route-rule`.  Similar to what I do deleting images in _docker_.  If you desire `--all` file it in a new Issue.

PR #773, opened 2017-05-17
Keep going when a delete fails during a multi-rule delete

esnible:
Fix for #771


cmd/istioctl/main.go
```diff
@@ -299,16 +299,22 @@ istioctl delete route-rule productpage-default
 					c.Println(c.UsageString())
 					return fmt.Errorf("provide configuration type and name or -f option")
 				}
+				var errs error
 				for i := 1; i < len(args); i++ {
 					if err := setup(args[0], args[i]); err != nil {
+						// If the user specified an invalid argument on the CLI,
```
ayj: s/argument/kind? That might make it more obvious why this case is handled differently than the file case below.

Issue #728, opened 2017-05-11
Ingress controller: robust validation


kyessenov:
The ingress controller accepts many inputs from the user to be put into envoy config: a list of hosts is arbitrary, TLS certs are just files. Some errors can be caught during validation, but some are only discovered at the ingress proxy. We should make sure that ingress proxy does not crash on a bad ingress to provide ingress resource isolation. There are two places where this is lacking:

- envoy requires hosts to be distinct
- envoy crashes on invalid cert files

kyessenov:
not as big of an issue with LDS/RDS

PR #776, opened 2017-05-17
Validate ingress certs

ijsnellf:
Contributes to https://github.com/istio/manager/issues/728


proxy/envoy/ingress.go
```diff
@@ -117,7 +117,7 @@ func fetchSecret(ctx context.Context, client *http.Client, url string,
 	}
 	secret := string(uri)
 	if secret == "" {
-		glog.Info("no secret needed")
+		glog.V(4).Info("no secret needed")
```
ijsnellf: Made this `.V(4)` because it is very noisy when looking at the logs.

kyessenov:
contributes to #728

PR #779, opened 2017-05-18
/version endpoint for apiserver

esnible:
This is needed to support the `istioctl version` command.  The functionality is similar to the /version endpoint on K8s api and the `kubectl version` command.  This is useful for supporting users -- let them tell us what build of client and server they are using.


apiserver/apiserver.go
```diff
@@ -93,6 +93,12 @@ func (api *API) Register(container *restful.Container) {
 		To(api.Status).
 		Doc("Returns 200 when the container is up"))
 
+	// Return information about the version of API server running
+	ws.Route(ws.
+		GET("/version").
+		To(api.Version).
+		Doc("Returns version information"))
+
```
liamawhite: Should be explicit in what it is writing e.g. `.Writes(Version{})`. Maybe make the Version struct private though.

esnible: I wasn't exactly sure what we *should* be returning.  I looked at what Kubernetes returns and I liked at the _istioctl_ client side.  I will go with what our client side defines.

PR #783, opened 2017-05-18
Make service account ID format SPIFFE-compatible

lookuptable:


myidpt:
I have prepared a PR for this, but it's also OK if you do the change. Please also change:
https://github.com/istio/manager/blob/master/model/service.go#L200
https://github.com/istio/manager/blob/master/model/service.go#L202
https://github.com/istio/manager/blob/master/proxy/envoy/testdata/cds-ssl-context.json.golden
https://github.com/istio/manager/blob/master/test/mock/service.go#L195
https://github.com/istio/manager/blob/master/test/mock/service.go#L196

platform/kube/controller.go
```diff
@@ -579,10 +579,10 @@ func (c *Controller) HostInstances(addrs map[string]bool) []*model.ServiceInstan
...
 
 const (
-	// IstioServiceAccountPrefix is always the prefix for Istio service accounts.
-	IstioServiceAccountPrefix = "istio:"
-	// LocalDomain is the domain name Istio service account uses for internal traffic.
-	LocalDomain = "cluster.local"
+	// the URI scheme used to encode a Kubernetes service account
+	uriScheme = "spiffe"
+	// the domain name Istio service account uses for internal traffic
+	localDomain = "cluster.local"
```
kyessenov: why are we not following kubernetes standard: `<name>.<namespace>.svc.<dns-domain>`?

myidpt: We are following the SPIFFE standard which is for the cert SAN. So it makes more sense to be a URI path format. The two formats are 1-1 mapped, so I won't see a problem.

kyessenov: It's yet another standard, so I'm wondering what the motivation is not to use service FQDN. I'm particularly interested how this is going to work in environments with no namespaces.

lookuptable: @kyessenov this encoded a service account, not a service. We are actually following how object path conversion used by the apiserver like https://kubernetes.io/docs/api-reference/v1.5/#create-482

model/service.go
```diff
@@ -194,12 +194,8 @@ type ServiceDiscovery interface {
...
 // ServiceAccounts exposes Istio service accounts
 type ServiceAccounts interface {
-	// Gets all the Istio service accounts mapped from service hostname, in istio identity format.
-	// For example,
-	// GetIstioServiceAccounts(catalog.myservice.com, 80) ->
-	//      --> [istio:serviceaccount1.ns1.cluster.local, istio:serviceaccount2.ns2.cluster.local]
-	// GetIstioServiceAccounts(backend.myservice.com, 8080) ->
-	//      --> [istio:serviceaccount3.ns3.cluster.local]
+	// GetIstioServiceAccounts returns a list of service accounts looked up from
+	// the specified service hostname and ports.
```
lookuptable: Removed comments here since this is the model package and I don't think we should give k8s specified examples. Also the interface has no way to enforce the implementation to return certain format.

kyessenov:

kyessenov:
@sebastienvas e2e smoketest is super flaky (3/4 failed, 1/4 succeeded)

sebastienvas:
@kyessenov there was a bunch of undeleted namespace which I cleaned. I also changed the jobs to not abort previously running jobs which was exacerbating this.

kyessenov:
jenkins stop, you're drunk

PR #784, opened 2017-05-19
print no resources found, use path.Join

### LINKED ISSUE #753: istioctl get route-rules should print "No resources found."
When there are no rules the output of `istioctl get route-rules` (and similar list commands) has no output.

Most users would like "No resources found." to be written to stderr.  That is what kubectl does when there are no resources of that type.

rkpagadala:
fixes #753


kyessenov:

PR #786, opened 2017-05-19
Make domain suffix configurable

kyessenov:
Allow a different DNS domain suffix than "cluster.local"

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -207,6 +207,8 @@ func init() {
 		"Select a namespace for the controller loop. If not set, uses ${POD_NAMESPACE} environment variable")
 	rootCmd.PersistentFlags().DurationVar(&flags.controllerOptions.ResyncPeriod, "resync", time.Second,
 		"Controller resync interval")
+	rootCmd.PersistentFlags().StringVar(&flags.controllerOptions.DomainSuffix, "domainSuffix", "cluster.local",
+		"Kubernetes DNS domain suffix")
 	rootCmd.PersistentFlags().StringVar(&flags.meshConfig, "meshConfig", cmd.DefaultConfigMapName,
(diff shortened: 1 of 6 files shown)


cmd/manager/main.go
```diff
@@ -207,6 +207,8 @@ func init() {
 		"Select a namespace for the controller loop. If not set, uses ${POD_NAMESPACE} environment variable")
 	rootCmd.PersistentFlags().DurationVar(&flags.controllerOptions.ResyncPeriod, "resync", time.Second,
 		"Controller resync interval")
+	rootCmd.PersistentFlags().StringVar(&flags.controllerOptions.DomainSuffix, "domainSuffix", "cluster.local",
```
lookuptable: The flag used by kubeadm to specify this is called `--service-dns-domain`. Shall we just use the same thing?

platform/kube/controller.go
```diff
@@ -606,7 +607,7 @@ func (c *Controller) GetIstioServiceAccounts(hostname string, ports []string) []
 		}
 
 		pod, _ := item.(*v1.Pod)
-		sa := generateServiceAccountID(pod.Spec.ServiceAccountName, pod.GetNamespace(), localDomain)
+		sa := generateServiceAccountID(pod.Spec.ServiceAccountName, pod.GetNamespace(), c.domainSuffix)
```
lookuptable: CA needs to support similar flag.

Do you think service and service account should always have the same domain?

cmd/manager/main.go
kyessenov: Localkube calls it `dns-domain`, kubelet calls it `cluster-domain`. Seems like a lot of options :)

platform/kube/controller.go
kyessenov: You should be using service account domain name, not service domain name. Not sure if they are the same thing.

PR #788, opened 2017-05-19
istioctl version must report remote as well as local version info

esnible:


esnible:

PR #789, opened 2017-05-19
Make ingress election ID more unique

ijsnellf:
Make the ingress election ID more unique to reduce the chance of conflicts with other ingress controllers. Currently the election ID is the same as many other ingress controllers, which means the Istio ingress controller is competing with other controllers.


kyessenov:

PR #825, opened 2017-06-02
Sweeping rename "manager" -> "pilot"

kyessenov:


kyessenov:
jenkins rebuild istio/pilot

kyessenov:
we should wait for Seb to come back to fix all pipelines. I made presubmit/postsubmit work, but others are broken.

PR #830, opened 2017-06-05
Add support for statsd export

douglas-reid:
The goal of this PR is to enable collection of proxy stats. This PR focuses on generating envoy configs to direct statsd output.

This PR updates the proxy agent and the manager to support a new field in the mesh config: `statsd_udp_address`. The initial default option is to generate this address from the IP address of Mixer and the configured statsd-udp port (defaults to 9125).

Tested on new deployments. A PR to istio/istio that updates the deployment configs for mixer, pilot, and prometheus will follow.

This has a current weakness: if mixer service is deleted and restarted, export will stop working, as the `statsd_udp_address` is not regenerated in response.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 
 	"github.com/golang/glog"
@@ -57,6 +58,24 @@ func GetMeshConfig(kube kubernetes.Interface, namespace, name string) (*proxycon
 	mesh := proxy.DefaultMeshConfig()
 	if err = model.ApplyYAML(yaml, &mesh); err != nil {
 		return nil, multierror.Prefix(err, "failed to convert to proto.")
+	}
+
+	if mesh.MixerAddress != "" && mesh.StatsdUdpAddress == "" {
+		colon := strings.Index(mesh.MixerAddress, ":")
+		mixerAddr := mesh.MixerAddress[:colon]
+		if err := model.ValidateIPv4Address(mixerAddr); err == nil {
+			mesh.StatsdUdpAddress = mesh.MixerAddress
+		} else {
+			if mixerSvc, err := kube.CoreV1().Services(namespace).Get(mixerAddr, v1.GetOptions{}); err == nil {
+				port := int32(9125)
+				for _, sp := range mixerSvc.Spec.Ports {
+					if sp.Protocol == "UDP" && strings.Contains(sp.Name, "statsd") {
+						port = sp.Port
+					}
+				}
+				mesh.StatsdUdpAddress = fmt.Sprintf("%s:%d", mixerSvc.Spec.ClusterIP, port)
+			}
+		}
 	}
 
 	if err = model.ValidateProxyMeshConfig(&mesh); err != nil {
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"strings"
 	"syscall"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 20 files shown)


kyessenov:
There is a linter warning (run `bin/check.sh`)

model/validation.go
```diff
@@ -806,6 +806,28 @@ func ValidateProxyAddress(hostAddr string) error {
 	return nil
 }
 
+// ValidateIPv4AddressAndPort checks that a network address is well-formed.
+// This works just like ValidateProxyAddress, but without support for FQDNs.
```
kyessenov: why is FQDN not supported?

cmd/cmd.go
```diff
@@ -59,6 +60,24 @@ func GetMeshConfig(kube kubernetes.Interface, namespace, name string) (*proxycon
 		return nil, multierror.Prefix(err, "failed to convert to proto.")
 	}
 
+	if mesh.MixerAddress != "" && mesh.StatsdUdpAddress == "" {
+		colon := strings.Index(mesh.MixerAddress, ":")
+		mixerAddr := mesh.MixerAddress[:colon]
+		if err := model.ValidateIPv4Address(mixerAddr); err == nil {
+			mesh.StatsdUdpAddress = mesh.MixerAddress
+		} else {
+			if mixerSvc, err := kube.CoreV1().Services(namespace).Get(mixerAddr, v1.GetOptions{}); err == nil {
```
kyessenov: cmd/ should not be depending on k8s. I'd prefer we did not automatically set values in the config since it obscures problems and makes debugging difficult.

kyessenov: (regarding the fetch above, it is going away once mesh config becomes a proper config resource)

model/validation.go
douglas-reid: This was needed because the statsd udp address (as currently processed by Envoy) has to be <IP>:<Port>. So, the thinking was to have some validation of something that can only be an IP Addr.

cmd/cmd.go
douglas-reid: Without this, it makes configuring the export much harder. I'm open to good ideas, however.

Envoy wants an IP, not a service name. So, to get this to work, something needs to map from say mixer_server to ip. It seems easier to have cmd do this than it is to have an operator deploy mixer first, get the IP, then update the mesh file, then deploy pilot.

But, yes, if there is a better way, let's shift to that.

model/validation.go
kyessenov: I see

cmd/cmd.go
kyessenov: Can you do DNS resolution yourself here? Please don't fallback onto mixer address though, we should just change the value of statsd address to IP:PORT until Envoy gets DNS support for statsd setting. You can use go's `LookupHost` to parse out and replace host.

proxy/envoy/config.go
```diff
@@ -111,6 +115,21 @@ func buildConfig(listeners Listeners, clusters Clusters, mesh *proxyconfig.Proxy
 		},
 	}
 
+	if mesh.StatsdUdpAddress != "" {
```
kyessenov: config gen is supposed to be a pure function since its output is cached. Can you please move this chunk to `proxy/envoy/watcher.go` close to where agent is instantiated? You'd also need to resolve DNS in `proxy/envoy/ingress.go` and `proxy/envoy/egress.go`. 

If you want to reload Envoy on DNS change, you'd need to add a polling routine and trigger. I suggest not do that and just wait till Envoy starts using UDP clusters.

rshriram: @douglas-reid silly question: why the choice of udp? why not use TCP?

douglas-reid: @rshriram the idea is to take advantage of the `statsd_udp_ip_address` field recently added to envoy (in conjunction with the official https://github.com/prometheus/statsd_exporter). This is all to quickly get some (shared) unified visibility into the envoy metrics.

douglas-reid:
@kyessenov I've updated to do resolve before agent instantiation and added to ingress/egress per your comments. I believe this matches what you were asking for. PTAL.

proxy/envoy/egress.go
```diff
@@ -41,6 +41,14 @@ func NewEgressWatcher(mesh *proxyconfig.ProxyMeshConfig) (Watcher, error) {
 	if mesh.EgressProxyAddress == "" {
 		return nil, errors.New("egress proxy requires address configuration")
 	}
+	if mesh.StatsdUdpAddress != "" {
+		if addr, err := resolveStatsdAddr(mesh.StatsdUdpAddress); err == nil {
+			mesh.StatsdUdpAddress = addr
+		} else {
+			glog.V(2).Infof("Error resolving statsd address; clearing to prevent bad config: %v", err)
```
kyessenov: Shouldn't this be a warning?

douglas-reid: done.

kyessenov:
Thanks!

douglas-reid:
Thanks for the good review, and the merge!

PR #831, opened 2017-06-05
use_remote_address for ingress

kyessenov:
Enable updating `x-forwarded-for` on the ingress proxy.
The sidecar proxy passes the header unmodified (documented in Envoy and validated with a test).

cc @saturnism


rshriram:
Any idea why this wasn't being passed by ingress alone ?

kyessenov:
@saturnism - Envoy should pass the header along even without this setting. Are you sure the downstream proxy is setting it?

PR #834, opened 2017-06-06
Generic config interface

kyessenov:
Separate storage from domain specific queries to retrieve config objects. The usable functions are just shims on top of the underlying storage library.

Add a key function to generalize name and namespace. The function derives a unique string (or maybe a path) from the content of the config.

Thoughts?


model/config.go
```diff
@@ -61,34 +44,68 @@ func (k Key) String() string {
...
-	// Get retrieves a configuration element, bool indicates existence
-	Get(key Key) (proto.Message, bool)
+	// Get retrieves a configuration element.
+	Get(kind, key string) (config proto.Message, exists bool, revision Revision)
 
-	// List returns objects for a kind in a namespace.
-	// Use namespace "" to list resources across namespaces
-	List(kind string, namespace string) (map[Key]proto.Message, error)
+	// List returns objects by type indexed by the key
+	List(kind string) (map[string]proto.Message, error)
```
mandarjog: Listing is typically 1 level deep. based on the resource scheme `/route-rule/{destination-service}/{rule-name}`    List("route-rule")  needs to be 2 levels deep., recurse, depth should be args?

IstioRegistry should resolve it to a key and call down into ConfigRegistry.
Or the config registry needs to know the details of individual message types, which seems incorrect.

kyessenov: This is a bare minimum. TPR backend only has one level of subkey nesting which is a namespace, and it's not right to use namespace for a subkey. IstioRegistry definitely needs to know how keys are derived and possibly implement subkey listing.

model/config.go
```diff
@@ -61,34 +44,68 @@ func (k Key) String() string {
...
 	// Delete removes an object from the distributed store by key.
-	Delete(key Key) error
+	Delete(kind, key string) error
+}
+
+// IstioRegistry is an interface to access config registry using Istio
+// configuration types
+type IstioRegistry interface {
+	// RouteRules lists all routing rules indexed by keys
+	RouteRules() map[string]*proxyconfig.RouteRule
```
kyessenov: For a route rule, that key would work as long as it satisfies key schema in the backend (e.g. size limit and no slash). The keys are unique per rules, not the name.

model/config.go
```diff
@@ -61,34 +44,68 @@ func (k Key) String() string {
...
+
+	// RouteRulesBySource selects routing rules by source service instances.
+	// A rule must match at least one of the input service instances since the proxy
+	// does not distinguish between source instances in the request.
+	// The rules are sorted by precedence (high first) in a stable manner.
+	RouteRulesBySource(instances []*ServiceInstance) []*proxyconfig.RouteRule
+
+	// IngressRules lists all ingress rules
+	IngressRules() map[string]*proxyconfig.RouteRule
```
kyessenov: It's same as route rule. The API for ingress is not done.

model/config.go
```diff
@@ -61,34 +44,68 @@ func (k Key) String() string {
...
+
+	// DestinationRules lists all destination rules
+	DestinationRules() []*proxyconfig.DestinationPolicy
```
kyessenov: Policy being renamed to Rule. This is not a query, the route rule and destination rule are very different objects.

model/config.go
rshriram: If we are doing revisions, I thikn having a list with revision numbers or a get with revision numbers is useful.

model/config.go
```diff
@@ -61,34 +44,68 @@ func (k Key) String() string {
...
 	// key already exists, the operation fails with no side effects.
-	Post(key Key, v proto.Message) error
+	Post(kind string, v proto.Message) (Revision, error)
 
 	// Put updates a configuration object in the distributed store.
 	// Put requires that the object has been created.
-	Put(key Key, v proto.Message) error
+	// Revision prevents overriding a value that has been changed
+	// between prior _Get_ and _Put_ operation to achieve optimistic concurrency.
+	Put(kind string, v proto.Message, revision Revision) (Revision, error)
```
rshriram: Same as post.
I am not really comfortable with specifying revision as part of the update. Its going to make UX painful.

And why return the same revision ?

model/config.go
```diff
@@ -61,34 +44,68 @@ func (k Key) String() string {
...
 
 	// Delete removes an object from the distributed store by key.
-	Delete(key Key) error
+	Delete(kind, key string) error
+}
```
rshriram: delete by revision as well, to stay consistent. Allows reverting a config to older version.

model/config.go
```diff
@@ -61,34 +44,68 @@ func (k Key) String() string {
...
 type ProtoSchema struct {
 	// MessageName refers to the protobuf message type name
 	MessageName string
+
 	// Validate configuration as a protobuf message
 	Validate func(o proto.Message) error
+
+	// Key function derives the unique key from the configuration content
+	Key func(o proto.Message) string
+
```
rshriram: This gets confusing. If I want some user readable names for managing configs of different types, why shoudl we create our own key generator? It makes it hard to operate on get/delete/put

model/config.go
```diff
@@ -61,34 +44,68 @@ func (k Key) String() string {
...
+	Put(kind string, v proto.Message, revision Revision) (Revision, error)
 
 	// Delete removes an object from the distributed store by key.
-	Delete(key Key) error
+	Delete(kind, key string) error
+}
+
+// IstioRegistry is an interface to access config registry using Istio
+// configuration types
+type IstioRegistry interface {
```
rshriram: may be IstioConfigStore ? or would it involve lots of renaming?

model/config.go
kyessenov: agreed about List. Get in the past won't work since etcd compactifies old revisions. It's not meant to be a linear sequence of revisions since there's no common API to query a range across various KV backends.

model/config.go
```diff
@@ -61,34 +44,68 @@ func (k Key) String() string {
...
-	// List returns objects for a kind in a namespace.
-	// Use namespace "" to list resources across namespaces
-	List(kind string, namespace string) (map[Key]proto.Message, error)
+	// List returns objects by type indexed by the key
+	List(kind string) (map[string]proto.Message, error)
 
 	// Post creates a configuration object. If an object with the same
 	// key already exists, the operation fails with no side effects.
-	Post(key Key, v proto.Message) error
+	Post(kind string, v proto.Message) (Revision, error)
```
kyessenov: key is a function of kind and message that's pre-built into the implementation of this interface. This is due to inability to mark a subset of properties of the value as metadata in proto land. If you think of "v" as having "name field", the key is derived from "v" using the right prototype ("kind").

model/config.go
kyessenov: Deliberate for optimistic concurrency of writes. We don't want to lose updates for concurrent `Put` operation. That's why you first need to fetch the last revision you know using `Get` then supply to `Put`. One of the updates will be blocked and require a retry. The returned revision is a new revision corresponding to the successful store write.

model/config.go
kyessenov: There's no going back to history since KV don't keep the whole history. This is not an infinite log of revisions. In fact, the interface is deliberately not allowing revision ranges since it's not well defined for various implementations of KV.

model/config.go
kyessenov: Sure. `IstioConfig`. This change is already causing massive repercussions throughout the code.

model/config.go
kyessenov: Key depends on the content of the message and specifically schema-defined metadata fields in the message. If I mark "destination" and "name" as the primitive composite key, this function will be return "{destination}/{name}". We can refine it to be an actual field query + composition, not an arbitrary function, but the idea stays the same.

model/config.go
```diff
@@ -61,34 +44,68 @@ func (k Key) String() string {
...
+	IngressRules() map[string]*proxyconfig.RouteRule
+
+	// DestinationRules lists all destination rules
+	DestinationRules() []*proxyconfig.DestinationPolicy
+
+	// DestinationPolicy returns a policy for a service version.
+	DestinationPolicy(destination string, tags Tags) *proxyconfig.DestinationVersionPolicy
 }
 
+// Revision is an opaque identifier for auditing updates to the config registry.
```
rshriram: In the light of comments above, I think this comment needs to be either removed or clarified. This threw me off, as I started assuming that the object revision in GETs/PUTs referred to their version and that the revisions would be kept until the user deleted them explicitly. Whereas, in reality, what we are doing is simply abstracting the underlying KV store (etcd, redis, memcached, etc.) where the revision is any arbitrary id to identify different (non-semantic) variants of the object (revision != git commit ID). 

Appreciate all the comments above (may be just stick them into the PR as well, as they explain the rationale well.

Secondly, please refine the comment below to indicate that the "Revision field pertains to the object revision of the underlying Data store, and has no bearing on audit logs, etc. The lifetime of an object of a particular revision depends on the underlying data store (e.g., Etcd's compaction of revisions). The config client should not make any assumptions about revisions and rely only on exact equality to implement optimistic concurrency control of read-write operations."

kyessenov: Thanks. I will move the explanations to the code comments. The revision can be used for auditing updates, but we don't place a requirement on the config storage to preserve all records of updates. There should be a separate audit system that logs all mutations together with the revision ID that can work together with the config store.

model/config.go
mandarjog: Why is key construction responsibility of `ConfigRegistry` ?
That should be the responsibility of the `IstioRegistry` which can use the `Key` function.
That would mean that `ConfigRegistry` can go back to being KV style interface.

What is the benefit of `ConfigRegistry` understanding all this?

PR #841, opened 2017-06-07
Generic config store interface

kyessenov:
(re-opening PR, somehow couldn't update the other one)

I apologize for the size of the PR. Start from `model/config.go`, the rest were necessary changes to make the new model fit.


model/config.go
```diff
@@ -24,75 +23,79 @@ import (
...
+	// KindMap exposes the configuration type schema known by the config store.
+	// The type schema defines the bidrectional mapping between configuration
+	// types (also called kinds) and the protobuf message names.
+	KindMap() KindMap
 
-	// List returns objects for a kind in a namespace.
-	// Use namespace "" to list resources across namespaces
-	List(kind string, namespace string) (map[Key]proto.Message, error)
+	// Get retrieves a configuration element by a kind and a key
+	Get(kind, key string) (config proto.Message, exists bool, revision Revision)
```
kyessenov: The key uniquely identifies a config. The structured keys are not part of this design. The two can be unified as: `List(kind string, keyPrefix []string) ([]Config, error)`

model/config.go
```diff
@@ -24,75 +23,79 @@ import (
...
-	return fmt.Sprintf("%s/%s-%s", k.Namespace, k.Kind, k.Name)
-}
-
-// ConfigRegistry defines the basic API for retrieving and storing
-// configuration artifacts.
+// The storage interface presented here assumes that the underlying storage
+// layer supports _GET_ (list), _PUT_ (update), _POST_ (create) and _DELETE_
+// semantics but does not guarantee any transactional semantics.
+//
+// Configuration objects possess a type property (also called kind throughout)
```
rshriram: if you have the appetite for it, how about kind -> type ?

kyessenov: I'd love to, but type is a reserved word. Maybe go the other way, type -> kind?

kyessenov: Done

kyessenov:
@rshriram added a cache interface with a note about grace period for full sync

kyessenov:
PTAL. This intentionally breaks istioctl since the PR is already way too large. I'll fix istioctl later.

adapter/config/memory/config.go
```diff
@@ -0,0 +1,137 @@
...
+	return val, exists, cr.revs[typ][key]
+}
+
+// List implements config registry method
+func (cr *store) List(typ string) ([]model.Config, error) {
+	_, ok := cr.data[typ]
+	if !ok {
+		return nil, nil
+	}
+	out := make([]model.Config, 0)
```
ijsnellf: We know the output array size:
`out := make([]model.Config, 0, len(cr.data[typ]))`

adapter/config/memory/config.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, typ := range descriptor.Types() {
+		out.data[typ] = make(map[string]proto.Message)
+		out.revs[typ] = make(map[string]string)
+	}
+	return &out
+}
+
+type store struct {
+	descriptor model.ConfigDescriptor
+	data       map[string]map[string]proto.Message
```
ayj: Do these need to be protected with a mutex?

kyessenov: I don't see concurrent use in our code. By default, nothing is thread-safe, so it's ok to skip mutex for now.

kyessenov:
presubmit passed but jenkins failed to post status

kyessenov:
merging, will do follow up PRs

adapter/config/memory/config.go
```diff
@@ -0,0 +1,137 @@
...
+	_, ok := cr.data[typ]
+	if !ok {
+		return nil, false, ""
+	}
+	val, exists := cr.data[typ][key]
+	return val, exists, cr.revs[typ][key]
+}
+
+// List implements config registry method
+func (cr *store) List(typ string) ([]model.Config, error) {
```
kyessenov: We need to return messages together with revisions. Get could also return a single Config and let the client unpack the struct. This is better than `[]revision, []proto.Message`

ldemailly:
build pilot/e2e-smoketest

sebastienvas:
build pilot/e2e-smoketest

PR #842, opened 2017-06-07
Cherry-picks from master

kyessenov:


sebastienvas:
Jenkins job pilot/presubmit passed

sebastienvas:
Jenkins job pilot/presubmit passed

ldemailly:
so the difference between release-0.1 and master is absence of unary in 0.1 ?
(can this be seen through the github UI somehow ? or that the diff between the 2 is the same as the unary commit ?)

kyessenov:
seems like it passed but status update is not working @sebastienvas

PR #852, opened 2017-06-09
ThirdPartyResource config store backend

kyessenov:
Mostly code shuffling...
Splitting the big kube controller into pieces.


kyessenov:
This was intentional since I'll split ingress away as well later, once we can aggregate across config sources.

PR #860, opened 2017-06-12
Move istioctl out of pilot to istio/config

kyessenov:


kyessenov:
no idea why jenkins fails, builds and tests fine locally

PR #861, opened 2017-06-12
Remove cfg editor dead code (to be moved to galley)

kyessenov:


kyessenov:
jenkins cached removed code and doesn't compile anymore

PR #863, opened 2017-06-12
Add version to all commands

kyessenov:


kyessenov:
@sebastienvas jenkins can you clean the code cache please?

Issue #866, opened 2017-06-13
Rule validation (abort 0% crashes envoy)


esnible:
If I create a rule like this:

```
destination: ratings.default.svc.cluster.local
httpFault:
  abort:
    httpStatus: 500
```

Then Envoy will fail with 

```
[critical][main] error initializing configuration '/etc/envoy/envoy-rev0.json': fault filter must have at least abort or delay specified in the config.
```

According to _istio/api/proxy/v1/config/http_fault.proto_:92 the percentage of requests to abort can be in the range 0-100.  If that claim is true the code that generates the Envoy configuration needs to special-case 0% and suppress creating the fault.  If that claim is false and Istio only supports the range 1-100 then the rule validation should be improved.

Note that a rules like

```
destination: ratings.default.svc.cluster.local
httpFault: {}
```
and

```
destination: ratings.default.svc.cluster.local
```

induce a similar crash.

kyessenov:
Thanks for catching this!

PR #874, opened 2017-06-15
Fix invalid Envoy fault bug

ijsnellf:
Fixes bug where special case was causing an invalid fault with no abort or delay to be generated. Resolves https://github.com/istio/pilot/issues/866.


proxy/envoy/fault.go
```diff
@@ -40,14 +40,19 @@ func buildFaultFilters(routeConfig *HTTPRouteConfig) []HTTPFilter {
 
 // buildFaultFilter builds a single fault filter for envoy cluster
 func buildHTTPFaultFilter(cluster string, faultRule *proxyconfig.HTTPFaultInjection, headers Headers) *HTTPFilter {
+	abort, delay := buildAbortConfig(faultRule.Abort), buildDelayConfig(faultRule.Delay)
+	if abort == nil && delay == nil {
```
rkpagadala: should this be || instead of &&

proxy/envoy/fault.go
```diff
@@ -40,14 +40,19 @@ func buildFaultFilters(routeConfig *HTTPRouteConfig) []HTTPFilter {
 
 // buildFaultFilter builds a single fault filter for envoy cluster
 func buildHTTPFaultFilter(cluster string, faultRule *proxyconfig.HTTPFaultInjection, headers Headers) *HTTPFilter {
+	abort, delay := buildAbortConfig(faultRule.Abort), buildDelayConfig(faultRule.Delay)
```
rkpagadala: Also the following reads better to me :)
abort := buildAbortConfig(faultRule.Abort)
delay := buildDelayConfig(faultRule.Delay)

proxy/envoy/fault.go
ijsnellf: No, it is valid if you have just 1 abort or 1 delay. The only invalid case is having neither.

proxy/envoy/fault.go
ijsnellf: Sure, bit easier to read.

PR #871, opened 2017-06-15
Some missing manager->pilot and getting presubmit to work again

ldemailly:
leftovers/breaking changes from #860 and #825 and #863

PR #875, opened 2017-06-15
updating doc to avoid build error

ldemailly:


ldemailly:
I'd have prefered you didn't merge this until after I'm done fixing the build in #871 it was seconds from being done

kyessenov:
This seems to be a doc change? Would be there be a problem fast-forwarding? Unfortunately, doesn't automatically fast forward for us here.

Issue #876, opened 2017-06-15
pilot jenkins code coverage in presubmit takes 9 minutes


ldemailly:
pilot jenkins code coverage in presubmit takes 9 minutes

mandar said for mixer splitting -race and coverage in 2 jobs made the sum faster

kuat said it takes 30s on his machine

we should try to break down so we know how the time is spent ?

![screen shot 2017-06-15 at 3 31 54 pm](https://user-images.githubusercontent.com/3664595/27204394-d16e461e-51df-11e7-9fee-b9ce8096db13.png)

sebastienvas:
@nlandolfi already split the code coverage to another job in Prow. So I ll just assign to him.

sebastienvas:
Why did you close it ?

sebastienvas:
we ll fix it once Prow is in use.

kyessenov:
#878 fixed it by removing the race test

PR #878, opened 2017-06-15
Remove race test

kyessenov:
Important tests already run in parallel with `t.Parallel`.


kyessenov:
fixes #876

ldemailly:
build pilot/presubmit

ldemailly:
(hope we won't regret not having the race test - should that be done at least weekly or some such?)

kyessenov:
go race test is quite naive. it simply spawns tests in parallel and tries to catch unprotected writes. since our unit tests allocate resources (namespaces) this native approach is harmful as it generates  a lot of load on the cluster.

PR #885, opened 2017-06-19
Filter ingress ports

ijsnellf:
Filter out non-HTTP ingress ports. This should fix the Envoy crashing observed in https://github.com/istio/issues/issues/20. Instead, Pilot will log a warning.

### DIFF
--- a/proxy/envoy/ingress.go
+++ b/proxy/envoy/ingress.go
@@ -276,6 +276,9 @@ func buildIngressRoute(ingress *proxyconfig.IngressRule,
 	if err != nil {
 		return nil, "", err
 	}
+	if servicePort.Protocol != model.ProtocolHTTP {
+		return nil, "", fmt.Errorf("unsupported protocol %q for %q", servicePort.Protocol, service.Hostname)
+	}
 
 	// unfold the rules for the destination port
 	routes := buildDestinationHTTPRoutes(service, servicePort, rules)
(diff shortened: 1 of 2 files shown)


proxy/envoy/resolve.go
```diff
@@ -10,13 +10,13 @@ import (
 	"github.com/golang/glog"
 )
 
-func resolveStatsdAddr(statsdAdrr string) (string, error) {
-	if statsdAdrr == "" {
+func resolveStatsdAddr(statsdAddr string) (string, error) {
```
ijsnellf: fixed typo

rshriram: ? Why remove the if block?

proxy/envoy/ingress.go
```diff
@@ -276,6 +276,9 @@ func buildIngressRoute(ingress *proxyconfig.IngressRule,
 	if err != nil {
 		return nil, "", err
 	}
+	if servicePort.Protocol != model.ProtocolHTTP {
```
rshriram: Https?

proxy/envoy/resolve.go
rshriram: Ignore

proxy/envoy/ingress.go
kyessenov: Ingress is only for HTTP traffic in-cluster. It may expose HTTPS on the front side, but that's irrelevant here.

kyessenov:
Disregard e2e failures, the test is broken.

PR #891, opened 2017-06-20
consul sds adapter impl

GregHanson:
looking for feedback a few items since there is not as nice of a 1x1 mapping from consul service definitions to those currently defined in model.Service

* External service info
* loadbalancer IP vs service instance IP
* Port and port naming conversion
* constructing service hostname -> important for envoy config generation


platform/consul/conversion.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+func convertTags(tags []string) model.Tags {
+	out := make(model.Tags, len(tags))
+	for _, tag := range tags {
+		vals := strings.Split(tag, "=")
+		if len(vals) > 1 {
+			out[vals[0]] = vals[1]
+		} else {
+			// TODO safe to assume all tags are of form "key=value"?
```
ijsnellf: remove comment. we should allow tags that aren't of the form "key=value"

platform/consul/conversion.go
```diff
@@ -0,0 +1,155 @@
...
+	return &model.Port{
+		Name:     name,
+		Port:     port,
+		Protocol: convertProtocol(name),
+	}
+}
+
+func convertService(endpoints []*api.CatalogService) *model.Service {
+	name, addr, external := "", "", ""
+
```
ijsnellf: nit: remove random newlines

kyessenov:
I'll need some time to research consul to provide useful feedback :)

platform/consul/controller.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (c *Controller) getCatalogService(name string, q *api.QueryOptions) []*api.CatalogService {
+	endpoints, meta, err := c.client.Catalog().Service(name, "", q)
+	if err != nil {
+		glog.Warningf("Could not retrieve service catalogue from consul: %v", err)
+		return []*api.CatalogService{}
+	}
+
+	//TODO log or process query metadata info?
+	glog.V(4).Infof("Response time: %v", meta.RequestTime)
```
ijsnellf: we don't care about this information, do we?

platform/consul/conversion.go
```diff
@@ -0,0 +1,155 @@
...
+	name = parts[0]
+	datacenter = parts[2]
+	return
+}
+
+func convertProtocol(name string) model.Protocol {
+	// TODO default TCP or HTTP?
+	out := model.ProtocolTCP
+
+	switch name {
```
ijsnellf: use constants for "udp" etc, since they are used in multiple places.

platform/consul/conversion.go
```diff
@@ -0,0 +1,155 @@
...
+	return out
+}
+
+func convertPort(port int, name string) *model.Port {
+
+	// TODO default HTTP/TCP?
+	defaultName := "tcp"
+	if name == "" {
+		name = defaultName
+	}
```
ijsnellf: nit: `defaultName` unnecessary
```golang
if name == "" {
   name = "tcp" 
}
```

platform/consul/conversion.go
```diff
@@ -0,0 +1,155 @@
...
+			// TODO ExternalName come from metadata?
+			ExternalName: inst.NodeMeta[externalTagName],
+		},
+		Tags: tags,
+	}
+}
+
+// serviceHostname produces FQDN for a consul service
+func serviceHostname(name, node, datacenter string) string {
+	// TODO include consul node in Hostname?
```
ijsnellf: Istio is service centric, so the node should be omitted. Consul supports [node based](https://www.consul.io/docs/agent/dns.html#node-lookups) and [service based](https://www.consul.io/docs/agent/dns.html#service-lookups) DNS lookup. We should only support service based.

platform/consul/conversion.go
```diff
@@ -0,0 +1,155 @@
...
+	node, datacenter := "", ""
+
+	ports := model.PortList{}
+	for _, endpoint := range endpoints {
+		name = endpoint.ServiceName
+		ports = append(ports, convertPort(endpoint.ServicePort, endpoint.NodeMeta[protocolTagName]))
+
+		// TODO This will not work if service is a mix of external and local services
+		// or if a service has more than one external name
+		if endpoint.NodeMeta[externalTagName] != "" {
```
ijsnellf: It looks like external services in Consul are designated by using a hostname instead of an IP address (see https://www.consul.io/api/catalog.html#register-entity and https://www.consul.io/docs/guides/external.html). If that is the case, we should use ip/hostname to decide whether a service is external instead of extra meta data.

platform/consul/conversion.go
```diff
@@ -0,0 +1,155 @@
...
+			external = endpoint.NodeMeta[externalTagName]
+		}
+
+		// TODO how to handle if a service spans datacenters/nodes?
+		node = endpoint.Node
+		datacenter = endpoint.Datacenter
+
+		// TODO where should the loadbalancer IP be stored in consul?
```
ijsnellf: Consul doesn't have load balancers natively, like k8s. Instead it does simple random load balancing by randomizing the results of DNS queries (see https://www.consul.io/docs/agent/dns.html#standard-lookup). Since Consul doesn't have native load balancers, we don't need to support them in our adapter (at least to start).

platform/consul/controller.go
```diff
@@ -0,0 +1,146 @@
...
+
+package consul
+
+import (
+	"github.com/golang/glog"
+	"github.com/hashicorp/consul/api"
+	"istio.io/pilot/model"
+)
+
+type Controller struct {
```
ijsnellf: Since the plan is to have an Istio control plane per cluster (see https://groups.google.com/forum/#!topic/istio-dev/IZyAP17LVok), I believe it would make sense for the Consul registry adapter to bind to a particular data center.

Consul's definition of a data center:
> Datacenter - While the definition of a datacenter seems obvious, there are subtle details that must be considered. For example, in EC2, are multiple availability zones considered to comprise a single datacenter? *We define a datacenter to be a networking environment that is private, low latency, and high bandwidth.* This excludes communication that would traverse the public internet, but for our purposes multiple availability zones within a single EC2 region would be considered part of a single datacenter.
- https://www.consul.io/docs/internals/architecture.html#glossary

platform/consul/conversion.go
```diff
@@ -0,0 +1,155 @@
...
+			// TODO safe to assume all tags are of form "key=value"?
+			out[tag] = ""
+		}
+	}
+	return out
+}
+
+func convertPort(port int, name string) *model.Port {
+
+	// TODO default HTTP/TCP?
```
kyessenov: I believe an empty port name is allowed, so not necessary to set the name.

platform/consul/conversion.go
kyessenov: I'd assume a single datacenter for now. The model API does not have provisions for AZ/DCs yet, so we cannot make use of that information just yet.

platform/consul/conversion.go
ijsnellf: According to the `model` package, you need to set a name if you have multiple ports for a single service:
```go
type Port struct {
	// Name ascribes a human readable name for the port object. When a
	// service has multiple ports, the name field is mandatory
	Name string `json:"name,omitempty"
...
```

platform/consul/conversion.go
saumoh: at pilot won't we need to know all the nodes that make up a service? For example, reviews (rev. v1, v2, v3) and their node ip's inside k8s would be known?
otherwise how will a micro service hosted in consul be able to apply a destination based policy?

platform/consul/controller.go
```diff
@@ -0,0 +1,219 @@
...
+func (c *Controller) AppendInstanceHandler(f func(*model.ServiceInstance, model.Event)) error {
+	c.monitor.AppendInstanceHandler(func(obj interface{}, event model.Event) error {
+		f(convertInstance(&(*obj.(*api.CatalogService))), event)
+		return nil
+	})
+	return nil
+}
+
+// HasSynced implements model.ConfigStoreCache operation TODO
+func (c *Controller) HasSynced() bool {
```
kyessenov: I'd suggest removing ConfigStore/Cache out of this PR. It's not necessary for service discovery, and is a separate interface.

platform/consul/controller.go
```diff
@@ -0,0 +1,219 @@
...
+	return out
+}
+
+// Run all controllers until a signal is received
+func (c *Controller) Run(stop <-chan struct{}) {
+	c.monitor.Start(stop)
+}
+
+// GetIstioServiceAccounts implements model.ServiceAccounts interface TODO
+func (c *Controller) GetIstioServiceAccounts(hostname string, ports []string) []string {
```
kyessenov: Similar thing. This is not needed to implement ServiceRegistry interface.

WORKSPACE
```diff
@@ -331,6 +331,30 @@ go_repository(
     importpath = "google.golang.org/genproto",
 )
 
+go_repository(
+    name = "com_github_hashicorp_go_rootcerts",
+    commit = "6bb64b370b90e7ef1fa532be9e591a81c3493e00",
```
kyessenov: If you don't mind, can you put dates on these commits? We're not enforcing this practice, but strongly recommend to keep track of how old the dependencies are.

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+type consulServices map[string][]string
+type consulServiceInstances []*api.CatalogService
+
+// Handler specifies a function to apply on an object for a given event type
+type Handler func(obj interface{}, event model.Event) error
+
+// Monitor handles service and instance changes
+type Monitor interface {
+	Start(<-chan struct{})
+	Stop()
```
kyessenov: can you please document the concurrency requirements? can these two functions be called from two separate threads? what's the purpose of stop() is the stop channel can be used.

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (m *consulMonitor) Start(stop <-chan struct{}) {
+	m.isStopped = false
+	m.tickChan = time.NewTicker(m.period).C
+	m.run(stop)
+}
+
+func (m *consulMonitor) Stop() {
+	m.isStopped = true
```
kyessenov: It looks like Start is blocking. That means Stop would be called from another thread, so the shared variable should be protected by a mutex, right?

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+
+package consul
+
+import (
+	"reflect"
+	"sort"
+	"time"
+
+	"github.com/golang/glog"
+
```
kyessenov: The typical pattern is to group golang standard deps first, then third party packages, then istio packages. That's not enforced, but looks nicer.

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+}
+
+func (m *consulMonitor) UpdateServiceRecord() error {
+	svcs, _, err := m.discovery.Catalog().Services(nil)
+	if err != nil {
+		glog.Warningf("Error:%s in fetching service result", err)
+		return err
+	}
+	newRecord := consulServices(svcs)
```
kyessenov: I suggest normalizing the list by sorting it. That's a more reliable comparison.

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+	if !reflect.DeepEqual(newRecord, m.serviceCachedRecord) {
```
kyessenov: End-to-end, this is only used to trigger global cache invalidation in pilot service. So triggering _some_ event would work. However, we should try to build a smarter diff between two sets of services. The diff project can be done separately, as long as you document TODO item here well.

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+
+	insts := []*api.CatalogService{}
+	for name := range svcs {
+		endpoints, _, err := m.discovery.Catalog().Service(name, "", nil)
+		if err != nil {
+			glog.Warningf("Could not retrieve service catalogue from consul: %v", err)
+			continue
+		}
+		insts = append(insts, endpoints...)
+
```
kyessenov: Let's avoid empty spaces that are not needed.

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+func (m *consulMonitor) AppendServiceHandler(h Handler) {
+	m.serviceHandlers = append(m.serviceHandlers, h)
+}
+
+func (m *consulMonitor) AppendInstanceHandler(h Handler) {
+	m.instanceHandlers = append(m.instanceHandlers, h)
+}
+
+func (list consulServiceInstances) normalize() {
```
kyessenov: if this is only used in one place, i wouldn't bother with a function.

platform/consul/controller.go
```diff
@@ -0,0 +1,219 @@
...
+	return nil
+}
+
+// AppendServiceHandler implements a service catalog operation
+func (c *Controller) AppendServiceHandler(f func(*model.Service, model.Event)) error {
+	c.monitor.AppendServiceHandler(func(obj interface{}, event model.Event) error {
+		f(convertService(*obj.(*[]*api.CatalogService)), event)
```
kyessenov: Can we change the type of handler function to limit the unsafe casts?

platform/consul/conversion.go
```diff
@@ -0,0 +1,149 @@
...
+		if len(vals) > 1 {
+			out[vals[0]] = vals[1]
+		}
+	}
+	return out
+}
+
+func convertPort(port int, name string) *model.Port {
+	if name == "" {
+		name = "http"
```
kyessenov: model.ProtocolHTTP (avoid use of hard-coded strings)

platform/consul/conversion.go
```diff
@@ -0,0 +1,149 @@
...
+	parts := strings.Split(hostname, ".")
+	if len(parts) < 1 {
+		err = fmt.Errorf("missing service name from the service hostname %q", hostname)
+		return
+	}
+	name = parts[0]
+	return
+}
+
+func convertProtocol(name string) model.Protocol {
```
kyessenov: can we use the same function as in k8s to convert port name to protocol?

platform/consul/controller.go
```diff
@@ -0,0 +1,219 @@
...
+		glog.V(2).Infof("parseHostname(%s) => error %v", hostname, err)
+		return nil
+	}
+
+	endpoints := c.getCatalogService(name, nil)
+
+	instances := []*model.ServiceInstance{}
+	for _, endpoint := range endpoints {
+		inst := convertInstance(endpoint)
+		if tags.HasSubsetOf(inst.Tags) && portMatch(inst, ports) {
```
ijsnellf: `portMatch` could be more efficient. check against a `map[string]bool` populated with the port names.

platform/consul/controller.go
```diff
@@ -0,0 +1,219 @@
...
+	for svcName := range data {
+		for addr := range addrs {
+			endpoints := c.getCatalogService(svcName, nil)
+			for _, endpoint := range endpoints {
+				if addr == endpoint.ServiceAddress {
+					out = append(out, convertInstance(endpoint))
+				}
+			}
+		}
+	}
```
ijsnellf: use `addrs` map
```golang
for svcName := range data {
  endpoints := c.getCatalogService(svcName, nil)
  for _, endpoint := range endpoints {
     if addrs[endpoint.ServiceAddress] {
        out = append(out, convertInstance(endpoint))
     }
  }
}
```

platform/consul/controller_test.go
```diff
@@ -0,0 +1,219 @@
...
+			ServiceAddress: "172.19.0.8",
+			ServicePort:    9080,
+			NodeMeta:       map[string]string{protocolTagName: "tcp"},
+		},
+	}
+)
+
+func newServer() *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
```
ijsnellf: nit: whitespace

platform/consul/controller_test.go
```diff
@@ -0,0 +1,219 @@
...
+			fmt.Println(r.URL.Path)
+		}
+	}))
+}
+
+func TestInstances(t *testing.T) {
+	ts := newServer()
+	defer ts.Close()
+	controller, err := NewController(ts.URL, "datacenter")
+
```
ijsnellf: nit: whitespace before `if err != nil` check is confusing

platform/consul/controller_test.go
```diff
@@ -0,0 +1,219 @@
...
+		t.Errorf("GetService() incorrect service returned => %q, want %q",
+			service.Hostname, serviceHostname("productpage", "node", "datacenter"))
+	}
+}
+
+func TestServices(t *testing.T) {
+	ts := newServer()
+	defer ts.Close()
+	controller, err := NewController(ts.URL, "datacenter")
+
```
ijsnellf: nit: whitespace

platform/consul/controller_test.go
```diff
@@ -0,0 +1,219 @@
...
+			ServiceName:    "reviews",
+			ServiceTags:    []string{"version=v3"},
+			ServiceAddress: "172.19.0.8",
+			ServicePort:    9080,
+			NodeMeta:       map[string]string{protocolTagName: "tcp"},
+		},
+	}
+)
+
+func newServer() *httptest.Server {
```
ijsnellf: could reduce duplicate code by passing in `t *testing.T` and `t.Fail(err)` if `err != nil` inside `newServer`

platform/consul/conversion.go
```diff
@@ -0,0 +1,149 @@
...
+	protocolTagName = "protocol"
+	externalTagName = "external"
+)
+
+func convertTags(tags []string) model.Tags {
+	out := make(model.Tags, len(tags))
+	for _, tag := range tags {
+		vals := strings.Split(tag, "=")
+
+		// Tags not of form "key=value" are ignored to avoid possible collisions
```
ijsnellf: log when these tags are found?

platform/consul/monitor.go
ijsnellf: remove the Stop() function, the stop channel is sufficient.

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+func (m *consulMonitor) Stop() {
+	m.isStopped = true
+}
+
+func (m *consulMonitor) run(stop <-chan struct{}) {
+	var err error
+	for {
+		select {
+		case <-stop:
+			m.Stop()
```
ijsnellf: just `return` here

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+		// This is only a work-around solution currently
+		// Since Handler functions generally act as a refresher
+		// regardless of the input, thus passing in meaningless
+		// input should make functionalities work
+		//TODO
+		obj := &[]*api.CatalogService{}
+		var event model.Event
+		for _, f := range m.serviceHandlers {
+			if err := f(obj, event); err != nil {
+				glog.Warningf("Error:%s in executing handler function", err)
```
ijsnellf: `glog.Warningf("Error executing handler function: %v", err)`

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (m *consulMonitor) run(stop <-chan struct{}) {
+	var err error
+	for {
+		select {
+		case <-stop:
+			m.Stop()
+		case <-m.tickChan:
+			if err = m.UpdateServiceRecord(); err != nil {
```
ijsnellf: instead of halting the monitor, just log an error and continue

platform/consul/monitor.go
ijsnellf: oh, and `ticker.Stop()`

platform/consul/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+	newRecord.normalize()
+	if !reflect.DeepEqual(newRecord, m.instanceCachedRecord) {
+		// This is only a work-around solution currently
+		// Since Handler functions generally act as a refresher
+		// regardless of the input, thus passing in meaningless
+		// input should make functionalities work
+		obj := &api.CatalogService{}
+		var event model.Event
+		for _, f := range m.instanceHandlers {
+			if err := f(obj, event); err != nil {
```
ijsnellf: `go f(obj, event)` so that we don't block on notifications?

platform/consul/conversion.go
GregHanson: @kyessenov The k8s convert function is `convertProtocol(name string, proto v1.Protocol)` which does conversion based on protocol and port name.  Consul doesn't have a concept of protocol in its service instance definition or a port name so I am not sure it maps well to use the same function

platform/consul/controller.go
```diff
@@ -0,0 +1,171 @@
...
+		glog.V(2).Infof("parseHostname(%s) => error %v", hostname, err)
+		return nil
+	}
+
+	endpoints := c.getCatalogService(name, nil)
+
+	instances := []*model.ServiceInstance{}
+	for _, endpoint := range endpoints {
+		instance := convertInstance(endpoint)
+		if tags.HasSubsetOf(instance.Tags) && portMatch(instance, ports) {
```
ijsnellf: still not using a map for the port names to speedup lookup here

platform/consul/controller.go
```diff
@@ -0,0 +1,171 @@
...
+	c.monitor.AppendServiceHandler(func(instances []*api.CatalogService, event model.Event) error {
+		f(convertService(instances), event)
+		return nil
+	})
+	return nil
+}
+
+// AppendInstanceHandler implements a service catalog operation
+func (c *Controller) AppendInstanceHandler(f func(*model.ServiceInstance, model.Event)) error {
+	c.monitor.AppendInstanceHandler(func(instance *api.CatalogService, event model.Event) error {
```
ijsnellf: is `instance` supposed to be called `service` here?

platform/consul/conversion.go
```diff
@@ -0,0 +1,160 @@
...
+	name, addr, external := "", "", ""
+
+	ports := make(map[int]*model.Port)
+	for _, endpoint := range endpoints {
+		name = endpoint.ServiceName
+
+		port := convertPort(endpoint.ServicePort, endpoint.NodeMeta[protocolTagName])
+
+		if svcPort, exists := ports[port.Port]; exists && svcPort.Protocol != port.Protocol {
+			// TODO Warn in this case?
```
ijsnellf: remove TODO

platform/consul/monitor.go
```diff
@@ -0,0 +1,168 @@
...
+	m.run(stop)
+}
+
+func (m *consulMonitor) run(stop <-chan struct{}) {
+	ticker := time.NewTicker(m.period)
+	for {
+		select {
+		case <-stop:
+			ticker.Stop()
+			return //?
```
ijsnellf: get rid of `//?`

platform/consul/monitor.go
```diff
@@ -0,0 +1,168 @@
...
+			m.updateServiceRecord()
+			m.updateInstanceRecord()
+		}
+	}
+}
+
+func (m *consulMonitor) updateServiceRecord() {
+	svcs, _, err := m.discovery.Catalog().Services(nil)
+	if err != nil {
+		glog.Warningf("Error:%s in fetching service result", err)
```
ijsnellf: nit: `glog.Warningf("Could not fetch services: %v", err)`

platform/consul/monitor.go
```diff
@@ -0,0 +1,168 @@
...
+			}(f)
+		}
+		m.serviceCachedRecord = newRecord
+	}
+}
+
+func (m *consulMonitor) updateInstanceRecord() {
+	svcs, _, err := m.discovery.Catalog().Services(nil)
+	if err != nil {
+		glog.Warningf("Error:%s in fetching instance result", err)
```
ijsnellf: same as above

platform/consul/monitor.go
```diff
@@ -0,0 +1,168 @@
...
+		return
+	}
+
+	instances := []*api.CatalogService{}
```
ijsnellf: nit: the style in the rest of the pilot code is to use `make`

ijsnellf:
few small nits, but otherwise seems pretty good

platform/consul/controller.go
GregHanson: consul has no formal `Service` object and `CatalogService` is the representation of a service instance.  I went with `instance` to avoid confusion with model.Service and model.ServiceInstance

PR #892, opened 2017-06-20
upload coverage report to cloud storage

yutongz:


yutongz:
https://github.com/istio/test-infra/issues/221

WORKSPACE
```diff
@@ -307,6 +307,24 @@ new_go_repository(
     importpath = "github.com/golang/sync",
 )
 
+new_go_repository(
```
kyessenov: we don't use these repos, do we?

yutongz: They are dependencies for test-infra/toolbox/pkg_check.

PR #897, opened 2017-06-26
fix match rule discarded from ingress routes

### LINKED ISSUE #895: Istio route rules `match` criteria not respected for ingress traffic
@kalantar commented on [Wed Jun 21 2017](https://github.com/istio/issues/issues/38)

I created a service that selects among 2 deployments (different versions).
I created two istio route-rules.
The first defaults all traffic to v1.
The second sends requests that have a particular header to v2.
I then deployed the sleep sample so that I could test these rules from inside the cluster. I found that both rules worked as expected.
I then created an ingress so that I could initiate requests from outside the cluster.
When I tested it by sending requests from outside the cluster, I found that the default rule was applied to every request, including those that match the second rule.
Note that I am doing this in a non default namespace. However, this does not seem to be the issue, see also: https://groups.google.com/forum/#!topic/istio-users/yQEogcx2hJM

Here are the rules: 

    $ more *rule*
    ::::::::::::::
    default-route-rule.yml
    ::::::::::::::
    type: route-rule
    name: default-route
    namespace: user-kalantar
    spec:
      destination: echo.user-kalantar.svc.cluster.local
      precedence: 1
      route:
      - tags:
          version: one
        weight: 100
    ::::::::::::::
    user-route-rule.yml
    ::::::::::::::
    type: route-rule
    name: echo-user-rule
    namespace: user-kalantar
    spec:
      destination: echo.user-kalantar.svc.cluster.local
      precedence: 2
      match:
        httpHeaders:
          x-user:
            exact: test
      route:
      - tags:
          version: two

The ingress is: 

    kind: Ingress
    metadata:
      name: echo-ingress
      namespace: user-kalantar
      annotations:
        kubernetes.io/ingress.class: istio
    spec:
      backend:
        serviceName: echo
        servicePort: 80

---

@ijsnellf commented on [Wed Jun 21 2017](https://github.com/istio/issues/issues/38#issuecomment-310163784)

@kyessenov, could you take a look at this? You are better acquainted with ingress route rules and their limitations.

---

@frankbu commented on [Wed Jun 21 2017](https://github.com/istio/issues/issues/38#issuecomment-310167790)

To reproduce:
```
$ kubectl apply -f <(istioctl kube-inject -f samples/apps/helloworld/helloworld.yaml)

$ cat <<EOF | istioctl create
type: route-rule
name: hello-default
spec:
  destination: helloworld.default.svc.cluster.local
  route:
  - tags:
      version: v1
EOF

$ cat <<EOF | istioctl create
type: route-rule
name: hello-canary
spec:
  destination: helloworld.default.svc.cluster.local
  precedence: 2
  match:
    httpHeaders:
      foo:
        exact: bar
  route:
  - tags:
      version: v2
EOF

$ export HELLOWORLD_URL=$(kubectl get po -l istio=ingress -o 'jsonpath={.items[0].status.hostIP}'):$(kubectl get svc istio-ingress -o 'jsonpath={.spec.ports[0].nodePort}')

$ curl -H "foo: bar" http://$HELLOWORLD_URL/hello # always returns v1
Hello version: v1, instance: helloworld-v1-1286076397-2nrdg
```

frankbu:
This fixes #895

User issue: https://github.com/istio/issues/issues/38

This fix is made and tested against the release-0.1 branch. Corresponding fix needs to be made in master and the question is do we want to release this fix in a new alpha driver (0.1.x) or are we going to only fix it in 0.2.x? If we decide to put out another 0.1.x, we should also provide some kind of interim fix for the rbac issue that several people are complaining about (https://github.com/istio/istio/pull/333)

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -180,11 +180,11 @@ func (i *IstioRegistry) RouteRulesBySource(namespace string, instances []*Servic
 	rules := make([]*routeRuleConfig, 0)
 	for key, rule := range i.RouteRules(namespace) {
 		// validate that rule match predicate applies to source service instances
-		if rule.Match != nil {
+		if rule.Match != nil && rule.Match.Source != "" {
 			found := false
 			for _, instance := range instances {
 				// must match the source field if it is set
-				if rule.Match.Source != "" && rule.Match.Source != instance.Service.Hostname {
+				if rule.Match.Source != instance.Service.Hostname {
 					continue
 				}
 				// must match the tags field - the rule tags are a subset of the instance tags


model/config.go
```diff
@@ -180,11 +180,11 @@ func (i *IstioRegistry) RouteRulesBySource(namespace string, instances []*Servic
 	rules := make([]*routeRuleConfig, 0)
 	for key, rule := range i.RouteRules(namespace) {
 		// validate that rule match predicate applies to source service instances
-		if rule.Match != nil {
+		if rule.Match != nil && rule.Match.Source != "" {
 			found := false
```
rshriram: @frankbu can you add a comment on top of this if block to explain the rationale? Ingress routes won't have a source field right?

frankbu: @rshriram: this is just a simple coding bug. Any rule without a Match.Source field means "any source", so the function is supposed to return that rule, regardless of who's asking. But the bug is because Ingress calls this function with an empty instance list, and in that case it was doing the wrong thing with (i.e., discarding) the non-source-specific rules.

PR #901, opened 2017-06-29
fix match rule discarded from ingress routes (#897)

frankbu:
cherry-pick bug fix from release-0.1 branch


kyessenov:
I think linters got smarter and post some warnings unrelated to your change. Can you also please fix those to make presubmit pass?

PR #906, opened 2017-06-30
Correct the sds data structure

### LINKED ISSUE #898: Our JSON definition of load balancing weight is incorrect
I don't believe load balancing weights are implemented in Istio, but the data structure in _discovery.go_ does not match Envoy.

Istio will serialize the weight as a peer to the IP address (following https://github.com/istio/pilot/blob/master/proxy/envoy/discovery.go#L156 ) but the Envoy documentation says the weight goes under _tags_.

https://lyft.github.io/envoy/docs/configuration/cluster_manager/sds_api.html#config-cluster-manager-sds-api

liamawhite:
Resolves #898

### DIFF
--- a/proxy/envoy/discovery.go
+++ b/proxy/envoy/discovery.go
@@ -151,6 +151,12 @@ type hosts struct {
 type host struct {
 	Address string `json:"ip_address"`
 	Port    int    `json:"port"`
+	Tags    *tags  `json:"tags,omitempty"`
+}
+
+type tags struct {
+	AZ     string `json:"az,omitempty"`
+	Canary bool   `json:"canary,omitempty"`
 
 	// Weight is an integer in the range [1, 100] or empty
 	Weight int `json:"load_balancing_weight,omitempty"`
--- a/model/config.go
+++ b/model/config.go
@@ -15,7 +15,6 @@
 package model
 
 import (
-	"fmt"
 	"sort"
 
 	"github.com/golang/glog"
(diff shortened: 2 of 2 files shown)


proxy/envoy/discovery.go
```diff
@@ -151,6 +151,12 @@ type hosts struct {
 type host struct {
 	Address string `json:"ip_address"`
 	Port    int    `json:"port"`
+	Tags    tags   `json:"tags,omitempty"`
```
esnible: Perhaps we want *tags instead of tags?  I played around with this last week (trying to write my own, non-Istio SDS) and the _omitempty_ didn't do what I wanted unless tags was a *.

liamawhite: I don't think pointers should effect the behaviour of omitempty?

esnible: I wrote this: ```type tags struct {
	Weight int `json:"load_balancing_weight,omitempty"`
}

type host1 struct {
	Address string `json:"ip_address"`
	Port    int    `json:"port"`
	Tags    *tags  `json:"tags,omitempty"`
}

type host2 struct {
	Address string `json:"ip_address"`
	Port    int    `json:"port"`
	Tags    tags  `json:"tags,omitempty"`
}

func main() {
	var h1 host1
	var h2 host2
	
	sh1, _ := json.Marshal(h1)
	sh2, _ := json.Marshal(h2)
	
	fmt.Printf("h1 marshals as %v\n", string(sh1)) 
	fmt.Printf("h2 marshals as %v\n", string(sh2)) 
}``` The output was ```h1 marshals as {"ip_address":"","port":0}
h2 marshals as {"ip_address":"","port":0,"tags":{}}```

liamawhite: I see now. Thats because an empty pointer is nil but an empty tag is tag{}.

Probably makes sense to go with the pointer.

rshriram:
Ping?

PR #921, opened 2017-07-10
added priviledged=true to the init container and to all tests

bjartek:
See https://github.com/istio/issues/issues/34 for discussion on issue. 
Adding priviledged=true to the init container removes the error that me and @debianmaster got when we tried to run the init container with selinux enabled.

### DIFF
--- a/platform/kube/inject/inject.go
+++ b/platform/kube/inject/inject.go
@@ -130,8 +130,8 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
 		"securityContext": map[string]interface{}{
 			"capabilities": map[string]interface{}{
 				"add":        []string{"NET_ADMIN"},
-				"privileged": true,
 			},
+		  "privileged": true,
 		},
 	})
 
--- a/platform/kube/inject/inject.go
+++ b/platform/kube/inject/inject.go
@@ -129,7 +129,8 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
 		"imagePullPolicy": "Always",
 		"securityContext": map[string]interface{}{
 			"capabilities": map[string]interface{}{
-				"add": []string{"NET_ADMIN"},
+				"add":        []string{"NET_ADMIN"},
+				"privileged": true,
 			},
 		},
 	})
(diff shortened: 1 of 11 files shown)


bjartek:
I signed it!

bjartek:
There was an error in my testfiles, I am pretty new to bazel how do i run the tests locally. I tried bazel test in the  inject directory but it did not report any failure.

bjartek:
Sorry for the mess. I did not read the testing docs and therefore did not test it properly before i pushed. I am sorry for wasting time. All teste pass locally now using `bazel test ...`

I will squash when it is OK

bjartek:
I can try to turn it off and we can see. Since i have added NET_ADMIN to the priviledged scc in Openshift, mye guess is that it will work. 

I guess taking time to find out exactly what permission is needed on Openshift with SELinux would be the way to go. Then we can add a rbac rule for Openshift similar to what it is for kubernetes to the main repo.

bjartek:
I just tested with removing the NET_CAP_ADMIN capability from the init container and it runs fine without it. Do you want me to test when I remove the NET_CAP_ADMIN capability from the priviledged scc?

rshriram:
On Tue, Jul 11, 2017 at 7:08 AM Bjarte S. Karlsen <notifications@github.com>
wrote:

> I just tested with removing the NET_CAP_ADMIN capability from the init
> container and it runs fine without it. Do you want me to test when I remove
> the NET_CAP_ADMIN capability from the priviledged scc?
>
>
> I don't know what a priviledged scc means w.r.t. Openshift. But if you
feel that its more restrictive, yet accomplishes the task, that would be
great.


-- 

~shriram

ayj:
@sebastienvas, whats the current prow/jenkins incantation to run to kick start the presubmit and smoke tests?

> I guess taking time to find out exactly what permission is needed on Openshift with SELinux would be the way to go. Then we can add a rbac rule for Openshift similar to what it is for kubernetes to the main repo.

Agreed. In the meantime, its probably okay to have CAP_ADMIN and `privileged=true`. https://docs.google.com/document/d/1-5rLhv6MbB27lL8l2CWD6kR-eMa13_PPwOScVVZq8bo/edit may need to revisit this to only inject the minimal permissions for iptables manipulation. cc @andraxylia

rshriram:
closing this as another PR (based on this) has been merged. Thanks a lot @bjartek for your contribution. Please keep em coming :).

PR #922, opened 2017-07-11
added priviledged=true to the init container and to all tests

ayj:
Commit https://github.com/istio/pilot/pull/921 on behalf of @bjartek.   

Add `privileged=true` to init-containers to unblock deployments with SELINUX (e.g. OpenShift). As @bjartek suggested in his original PR, this is functionally sufficient, but longer term someone needs to sort through what exact permissions are needed with SELinux and add the necessary RBAC rules to OpenShift similar to what what it is for Kubernetes in the main repo.

Fixes istio/issues#34

### DIFF
--- a/platform/kube/inject/inject.go
+++ b/platform/kube/inject/inject.go
@@ -130,8 +130,8 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
 		"securityContext": map[string]interface{}{
 			"capabilities": map[string]interface{}{
 				"add":        []string{"NET_ADMIN"},
-				"privileged": true,
 			},
+		  "privileged": true,
 		},
 	})
 
--- a/platform/kube/inject/inject.go
+++ b/platform/kube/inject/inject.go
@@ -129,7 +129,8 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
 		"imagePullPolicy": "Always",
 		"securityContext": map[string]interface{}{
 			"capabilities": map[string]interface{}{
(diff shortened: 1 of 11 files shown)


kyessenov:
The mystery is still why SELinux requires more capabilities than NET_ADMIN for iptables.
Let's get this in, and solve the mystery once we involve more SELinux users/folks.

platform/kube/inject/inject.go
```diff
@@ -131,6 +131,7 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
 			"capabilities": map[string]interface{}{
 				"add": []string{"NET_ADMIN"},
 			},
+			"privileged": true,
```
rshriram: Quick feedback. We know that stuff works in Kube without this. And we know that for OpenShift, we need this. Rather than elevating privileges for all cases, can we constrain this to just OpenShift only? This need not be done in this PR, as I don't want to add onto @bjartek 's commits. Lets get this in and may be figure out an istioctl option for this?

ldemailly:
next time let's make sure to add the credit in the commit message itself
(and hopefully soon we don't have to do this 'takeover' business)

bjartek:
Let me know if you need more help with istio/Openshift :)

ldemailly:
so it turns out this is required after all ? should we revert this ?

https://groups.google.com/d/msg/istio-users/8gI2DdnvOag/Ds_6oG-bBgAJ

bjartek:
The reason I did this patch is not that post. There was issues when SELinux is enabled. Setting priviledged=true fixed it.

|bjarte

PR #929, opened 2017-07-14
Add kube support for availability zones

liamawhite:
Solves #737

### DIFF
--- a/platform/kube/cache.go
+++ b/platform/kube/cache.go
@@ -1,0 +1,73 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kube
+
+import (
+	"istio.io/pilot/model"
+	"k8s.io/client-go/pkg/api/v1"
+)
+
+// PodCache is an eventually consistent pod cache
+type PodCache struct {
+	cacheHandler
+
+	// keys maintains stable pod IP to name key mapping
+	// this allows us to retrieve the latest status by pod IP
+	keys map[string]string
+}
+
+func newPodCache(ch cacheHandler) *PodCache {
+	out := &PodCache{
+		cacheHandler: ch,
+		keys:         make(map[string]string),
+	}
+
+	ch.handler.Append(func(obj interface{}, ev model.Event) error {
+		pod := *obj.(*v1.Pod)
+		ip := pod.Status.PodIP
+		if len(ip) > 0 {
+			switch ev {
+			case model.EventAdd, model.EventUpdate:
+				out.keys[ip] = KeyFunc(pod.Name, pod.Namespace)
+			case model.EventDelete:
+				delete(out.keys, ip)
+			}
+		}
+		return nil
+	})
+	return out
+}
+
+// getPodByIp returns the pod or nil if pod not found or an error occurred
+func (pc *PodCache) getPodByIP(addr string) (*v1.Pod, bool) {
+	key, exists := pc.keys[addr]
+	if !exists {
+		return nil, false
+	}
+	item, exists, err := pc.informer.GetStore().GetByKey(key)
+	if !exists || err != nil {
+		return nil, false
+	}
+	return item.(*v1.Pod), true
+}
+
+// tagsByIP returns pod tags or nil if pod not found or an error occurred
+func (pc *PodCache) tagsByIP(addr string) (model.Tags, bool) {
+	pod, exists := pc.getPodByIP(addr)
+	if !exists {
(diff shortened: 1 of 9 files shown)


platform/kube/cache.go
```diff
@@ -0,0 +1,138 @@
...
+	ch.handler.Append(func(obj interface{}, ev model.Event) error {
+		node := *obj.(*v1.Node)
+		var ip string
+		for _, a := range node.Status.Addresses {
+			if a.Type == "InternalIP" {
+				ip = a.Address
+				break
+			}
+		}
+		if len(ip) > 0 {
```
rkpagadala: @rshriram the code is correct need > 0.

platform/kube/cache.go
```diff
@@ -0,0 +1,138 @@
...
+				delete(out.keys, ip)
+			}
+		}
+		return nil
+	})
+	return out
+}
+
+// getNodeByIP returns the node or nil if pod not found or an error occurred
+func (nc *NodeCache) getNodeByIP(addr string) (*v1.Node, bool) {
```
kyessenov: That's a common pattern in go. It's not expensive to return multiple values.

platform/kube/cache.go
```diff
@@ -0,0 +1,138 @@
...
+	item, exists, err := nc.informer.GetStore().GetByKey(key)
+	if !exists || err != nil {
+		return nil, false
+	}
+	return item.(*v1.Node), true
+}
+
+// getAZByIP returns the az and true if found
+// empty string and false if the node couldn't be found or an error occurred
+func (nc *NodeCache) getAZByIP(addr string) (string, bool) {
```
kyessenov: It lets us avoid assigning special meaning to empty string. Common pattern.

platform/kube/controller.go
```diff
@@ -33,6 +33,9 @@ import (
 	"istio.io/pilot/model"
 )
 
+// NodeAZLabel is the well-known label for kubernetes node AZs
```
kyessenov: Yes, please correct the doc.

platform/kube/cache.go
liamawhite: Yeah, that and I wanted to keep it in a similar structure to podcache. Also exists/bool often makes code more readable.

platform/kube/controller.go
liamawhite: What's the difference? I went with labels based on the definitions of the two in the docs.

platform/kube/cache.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, false
+	}
+	return convertTags(pod.ObjectMeta), true
+}
+
+// NodeCache is an eventually consistent node cache
+type NodeCache struct {
+	cacheHandler
+
+	// keys maintains stable node internal IP to name key mapping
```
liamawhite: I'm not 100% sure of the differences when in kube? Presumably its internal/external to the cluster?

platform/kube/controller.go
liamawhite:
```
metadata:
  annotations:
    volumes.kubernetes.io/controller-managed-attach-detach: "true"
  creationTimestamp: 2017-07-19T19:39:27Z
  labels:
    arch: amd64
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/os: linux
    failure-domain.beta.kubernetes.io/region: us-south
    failure-domain.beta.kubernetes.io/zone: hou02
``` 
Are we sure its not a label? 😕

liamawhite:
@rshriram @kyessenov Just added some more tests, should be ready to for full review. Implementation has changed slightly due to the ability to use node names as pointed out by @esnible, see code comments for link to kube docs that describe the relevant behaviour.

As discussed on istio-dev group this is turned on by default but a further issue will be created to enable users to set the `upstream.zone_routing.min_cluster_size` of envoy as documented [here](https://lyft.github.io/envoy/docs/configuration/cluster_manager/cluster_runtime.html#zone-aware-load-balancing). This needs a separate issue as its not a trivial piece of work.

liamawhite:
Need to write a few more tests in `proxy/envoy`...

platform/kube/controller.go
rshriram: Fine, lets stick with label. :)

liamawhite:
Not sure how code coverage has decreased, I wrote a whole bunch of new tests for code that was already there.

rshriram:
@istio/pilot-hackers - please do the Go code review.

platform/kube/controller.go
```diff
@@ -33,6 +33,9 @@ import (
...
+const NodeAZLabel = "failure-domain.beta.kubernetes.io/zone"
+
```
rshriram: Don't you want to process the region as well?

rshriram: Essentially you can flatten region and zone into one string since Envoy does not have support for regions

liamawhite: Done!

platform/kube/controller.go
```diff
@@ -33,6 +33,13 @@ import (
 	"istio.io/pilot/model"
 )
 
+const (
+	// NodeRegionLabel is the well-known label for kubernetes node region
+	NodeRegionLabel = "failure-domain.beta.kubernetes.io/region"
+	// NodeZoneLabel is the well-known label for kubernetes node zone
+	NodeZoneLabel = "failure-domain.beta.kubernetes.io/zone"
```
kyessenov: @rshriram the examples show that the zone includes the region.

PR #930, opened 2017-07-14
Publish coverage at postsubmit as well

yutongz:
As we did before at presubmit, now we want like to collect data from postsubmit. Theoretically, they should be the same, but we would like to figure it out. And if so, we will only keep one of them.


yutongz:
@kyessenov  PTAL

yutongz:
build pilot/e2e-smoketest

PR #933, opened 2017-07-18
Enable RBAC in pilot e2e

andraxylia:
This is needed so that e2e tests can be run on an RBAC enabled cluster.

### DIFF
--- a/test/integration/testdata/rbac-beta.yaml.tmpl
+++ b/test/integration/testdata/rbac-beta.yaml.tmpl
@@ -1,0 +1,102 @@
+# Permissions and roles for istio
+# To debug: start the cluster with -vmodule=rbac,3 to enable verbose logging on RBAC DENY
+# Also helps to enable logging on apiserver 'wrap' to see the URLs.
+# Each RBAC deny needs to be mapped into a rule for the role.
+# If using minikube, start with '--extra-config=apiserver.Authorization.Mode=RBAC'
+kind: ClusterRole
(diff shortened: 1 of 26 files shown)


test/integration/testdata/rbac-beta.yaml.tmpl
```diff
@@ -0,0 +1,109 @@
+# Permissions and roles for istio
+# To debug: start the cluster with -vmodule=rbac,3 to enable verbose logging on RBAC DENY
+# Also helps to enable logging on apiserver 'wrap' to see the URLs.
+# Each RBAC deny needs to be mapped into a rule for the role.
+# If using minikube, start with '--extra-config=apiserver.Authorization.Mode=RBAC'
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: istio-pilot
```
rshriram: Would you mind adding some comments on what this is for and what the other istio-pilot cluster roles are for?

andraxylia: Permissions are very tightly coupled with the code, and subject to change often. Any time pilot CRUD k8s resources, it needs to have associated permissions. The resources and verbs affected by RBAC are self described from the file. Why this is needed is everywhere in the code.

rshriram: I meant clearly describe what each role does. This is helpful for new folks who are grokking at the code.

PR #940, opened 2017-07-19
prow: add postsubmit job

nlandolfi:
Introduce the postsubmit job for prow.

Steps in post submit:
- Install prerequisites
- Go build
- Bazel test
- Code coverage

BN: can be run locally, but hard to test code coverage reporting unless in actual CI environment.


nlandolfi:
Corresponding test-infra PR: https://github.com/istio/test-infra/pull/287

kyessenov:
Why does codecov report about CI failures?

nlandolfi:
I think it might be related to the e2e test failing...is that flaky? Seeing as this change only adds a file I'm not sure why that would fail.

nlandolfi:
I think codecov automatically detects the Jenkins environment for the codecov being run by Jenkins presubmit. Then codecov checks that the builds pass...and it noticed that e2e test failed in this instance. Seems to me that we should disable codecov checking CI status because we manage that ourselves...

but the [documentation section on it is empty](https://github.com/codecov/support/wiki/Codecov-Yaml#how-to-disable-checking-ci-status)...:/

nlandolfi:
@istio-testing test this

nlandolfi:
Just re-ran `e2e-smoketest`, would be nice to see it pass, since adding a file in `prow/` directory shouldn't affect it. After that, @kyessenov , we should be good to merge.

nlandolfi:
oh, and just updated branch.

Issue #945, opened 2017-07-20
mount mesh config (from ConfigMap) as a file in sidecars


rshriram:
With the introduction of LDS support ( #934 ), we no longer need to query K8S api from the proxy agent for configuring Envoy. The only dependency is the mesh configuration, that is provided as a config map today.

Need to change our deployment configs (in istio/istio) and the tests to mount the config map as a file, and change the proxy agent code to read the mesh config from the file. This would greatly reduce the size of the proxy agent binary, plus make it more portable across non k8s environments.

ldemailly:
cc @costinm

PR #950, opened 2017-07-21
Eliminate k8s dependencies from proxy agent

kyessenov:
Proxy agent will have 0 dependencies on k8s after this PR.
As a by-product, the agent binary size is smaller, going from 40MB to 10MB.

Eliminates two uses of k8s API:
1. Fetching mesh config from config maps. This PR changes kube-inject to insert a volume mount and makes the agent read from file instead.
2. Fetching ingress secrets. This PR changes discovery service to send secrets over the wire directly.


andraxylia:
I see some changes in the certificates, can you elaborate on this?

kyessenov:
jenkins rebuild e2e/smoke-test

kyessenov:
jenkins rebuild e2e/smoke-test

cmd/agent/main.go
```diff
@@ -0,0 +1,127 @@
...
+
+			glog.V(2).Infof("version %s", version.Line())
+
+			// receive mesh configuration
+			mesh, err := cmd.ReadMeshConfig(meshconfig)
+			if err != nil {
+				return multierror.Prefix(err, "failed to read mesh configuration.")
+			}
+
+			glog.V(2).Infof("mesh configuration %s", spew.Sdump(mesh))
```
ldemailly: might want to put this under
```
if glog.V(2) {
  glog.V(2).Infof("mesh configuration %s", spew.Sdump(mesh))
}
```
as Sdump seems likely to be expensive
(on the flip side if this is just executed once, probably not a big deal)

cmd/cmd.go
```diff
@@ -17,45 +17,30 @@ package cmd
...
-	ConfigMapKey = "mesh"
-
-	// DefaultConfigMapName is the default config map name that holds the mesh configuration.
-	DefaultConfigMapName = "istio"
-)
-
-// GetMeshConfig fetches configuration from a config map
-func GetMeshConfig(kube kubernetes.Interface, namespace, name string) (*proxyconfig.ProxyMeshConfig, error) {
-	config, err := kube.CoreV1().ConfigMaps(namespace).Get(name, v1.GetOptions{})
+// ReadMeshConfig gets mesh configuration from a config file
```
ldemailly: How will the file get there ?

edit: looks like it's at inject time ? can you update your PR description to say a bit more about the change ? TIA

cmd/agent/main.go
kyessenov: Only called once, seems unnecessary.

costinm: For people not familiar with spew - a small comment on what it does, and why an extra dep is needed
( would be good to keep agent small ).

cmd/agent/main.go
```diff
@@ -0,0 +1,127 @@
...
+
+			var role proxy.Role = sidecar
+			if len(args) > 0 {
+				switch args[0] {
```
costinm: For standalone, it may be needed to combine the 3 roles (i.e. the sidecar to act as egress/ingress, in addition to its normal service-to-service)

cmd/agent/main.go
```diff
@@ -0,0 +1,127 @@
...
+			cmd.WaitSignal(stop)
+			<-stop
+			cancel()
+			return nil
+		},
+	}
+)
+
+func init() {
+	proxyCmd.PersistentFlags().StringVar(&meshconfig, "meshconfig", "/etc/istio/mesh",
```
costinm: For raw VM, I was planning to create a 'istio' user, and keep all configs in /home/istio.
This avoids the need for root - it's not common to have /etc/ files owned by regular users. 
Also easier to mount volumes/etc.

cmd/agent/main.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	proxyCmd = &cobra.Command{
+		Use:   "proxy",
+		Short: "Envoy proxy agent",
+		RunE: func(c *cobra.Command, args []string) error {
+			// set values from environment variables
+			if sidecar.IPAddress == "" {
+				sidecar.IPAddress = os.Getenv("POD_IP")
+			}
```
costinm: TODO: if still empty, query the node interfaces.
For pod_name: default to hostname.
For namespace: default to domainname.

cmd/agent/main.go
```diff
@@ -0,0 +1,127 @@
...
+		},
+	}
+)
+
+func init() {
+	proxyCmd.PersistentFlags().StringVar(&meshconfig, "meshconfig", "/etc/istio/mesh",
+		fmt.Sprintf("File name for Istio mesh configuration"))
+	proxyCmd.PersistentFlags().StringVar(&sidecar.IPAddress, "ipAddress", "",
+		"Sidecar IP address. If not provided uses ${POD_IP} environment variable.")
+	proxyCmd.PersistentFlags().StringVar(&sidecar.InstanceName, "podName", "",
```
costinm: Please avoid the term 'pod' in new code :-)

In particular for flags or visible things.

cmd/cmd.go
costinm: Config map is great - it's also how we can customize mixer config today, and works perfectly for raw VMs (where the orchestration can inject files).

However - need to make sure we update the docker/etc, so this change doesn't break stuff.

docker/BUILD
```diff
@@ -71,7 +82,7 @@ debug_docker_build(
 )
 
 debug_docker_build(
-    entrypoint = ["/usr/local/bin/pilot"],
+    entrypoint = ["/usr/local/bin/agent"],
```
costinm: Can we use istio-agent ? In raw VM (and in docker as well ) the agent name is to vague, there are many other agents.

platform/kube/inject/testdata/auth.cert-dir.yaml.injected
```diff
@@ -49,10 +49,16 @@ spec:
         securityContext:
           runAsUser: 1337
         volumeMounts:
+        - mountPath: /etc/istio
```
costinm: Any chance we could combine the istio config and the certs in the same volume ? And under /home/istio ?

cmd/agent/main.go
rshriram: That enters an entirely different territory. Egress and ingress have completely different config files from sidecars. I would not pile that complexity here until later. Much more easy to run 3 envoy processes in one VM in the simplest case, as a starting point.

cmd/agent/main.go
rshriram: This is already configurable. And its also read only. So, it should work easily on VMs as well. I wouldnt hard code any /home path as a default. btw, if I was creating an istio user, I would lock it down, and probably not give it a shell/home directory.

cmd/agent/main.go
rshriram: Not sure if is wise to encode a common logic like "if no ip address, get Localhost IP". This isn't portable in VMs with multiple interfaces. Either we have a separate main.go for VMs or we expand this such that we provide the interface as an argument. Secondly, we rely on the sidecar having a single IP address. In cases where a VM has multiple interfaces (and hence multiple IP addresses), we need to plumb this all the way to the discovery server. This will quickly get ugly. Lets side step the exact alternatives for now, and simply add a TODO stating that we need to handle scenarios where the host has multiple interfaces.

cmd/agent/main.go
```diff
@@ -0,0 +1,127 @@
...
+		Short: "Envoy proxy agent",
+		RunE: func(c *cobra.Command, args []string) error {
+			// set values from environment variables
+			if sidecar.IPAddress == "" {
+				sidecar.IPAddress = os.Getenv("POD_IP")
+			}
+			if sidecar.InstanceName == "" {
+				sidecar.InstanceName = os.Getenv("POD_NAME")
+			}
+			if sidecar.InstanceNamespace == "" {
```
rshriram: as @costinm says, may be rename this to DNS domain? and Name to hostname

docker/BUILD
```diff
@@ -5,6 +5,17 @@ load("//docker:debug_docker.bzl", "debug_docker_build")
 # Use "manual" target tag to skip rules in the wildcard expansion
 
 pkg_tar(
+    name = "agent_tar",
+    extension = "tar.gz",
+    files = [
+        "//cmd/agent",
```
rshriram: istio-envoy-agent

docker/BUILD
rshriram: istio-envoy-agent

platform/kube/inject/inject.go
```diff
@@ -152,19 +186,34 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
...
-	ports, err := healthPorts(t)
-	if err != nil {
-		return err
-	}
-	for _, port := range ports {
-		args = append(args, "--passthrough", strconv.Itoa(port))
-	}
+	_, _ = healthPorts(t)
+	/*
+		if err != nil {
```
rshriram: Please add a TODO, that we are effectively losing health checks in K8S and we need to surface this in service model.

platform/kube/inject/testdata/auth.cert-dir.yaml.injected
rshriram: once again, I think /home/istio is a bad idea. Expecting people to create a /home for a system user is not ideal. If anything, people would prefer /var/run or /var/lib/ . And its best to keep this configurable.. If we have debian packaging, then its easy to automate installation of istio user, and create /var/lib/istio path, that is world readable and writable only by istio.

cmd/agent/main.go
costinm: 3 processes would be cleaner, indeed. It's also not clear we'll even need this, but I suspect
at least egress might be better if it's local/close to the VMs. We can worry about it later, I didn't realize the configs are not 'additive'

cmd/agent/main.go
costinm: My concern is that we have a bit of a mess in /etc - certs in some place, mixer in other, etc.
Some of the configs (certs) will not be static, and I don't like requiring root.
I know it's configurable - but the defaults should match the config we ship by default, and I'm hoping we can organize it a bit better.

AFAIK mysql, postgress and few others have 'home', and may be easier to control. 
/var/ is also good, but we'll also end up spread (unless we do /var/lib/istio for example ).

We can discuss options on the mailing list, I'm ok with the PR as is.

docker/BUILD
costinm: +1

platform/kube/inject/testdata/auth.cert-dir.yaml.injected
costinm: +1 on /var/lib/istio.  How about /opt/istio for the binaries ?

cmd/agent/main.go
costinm: Having a "POD_IP" on a raw VM is extremely unlikely - while having a single interface is more common. Granted, we hope the interface will also support IPv6 add so a bunch of IPs. However this is a problem with POD_IP too. 

I think we could support the most common case by default ( single interface), minimizing user
manual config for it - and for advanced case we'll have to let admin handle it. 

But can be separate PR.

cmd/agent/main.go
kyessenov: Let's first define what VM sidecar identity is before going further. This is mostly a requirement from Mixer (it needs some UID to populate missing attributes, pilot just cares about IP addresses). I'll rename "pod" to instance throughout, and then will leave it at that.

cmd/agent/main.go
kyessenov: Name and namespace are required by Mixer. Pilot doesn't care, so please coordinate with Mixer first before changing this. We are using the flag 
 for POD_IP due to uncertainty with multiple network interfaces. It would be same in k8s environment as reading POD_IP so I'm ambivalent which way to go.

cmd/agent/main.go
kyessenov: spew prints golang structs in readable form. Readable logs far out-weight the slightly larger binary.

cmd/agent/main.go
kyessenov: Combining ingress/egress/sidecar into proxy config is possible, but far from a minor change. Certainly not in this PR.

cmd/agent/main.go
kyessenov: Home is entirely unnecessary in docker environment where everything is under virtualized root. We should prioritize docker defaults IMHO, not VM defaults. These are all flags, so are trivial to change in a start-up scripts. I think systemd these days also virtualizes config files so the exact path is just a preference.

cmd/agent/main.go
kyessenov: Yes, I'll clean up all pod references.

cmd/cmd.go
kyessenov: Can you elaborate what you mean by update docker? Docker image won't function standalone.

docker/BUILD
kyessenov: Yes, will do.

platform/kube/inject/testdata/auth.cert-dir.yaml.injected
kyessenov: These are all virtualized paths. Can you first define the whole list of paths that are needed for istio-on-vm, and we'll align them later? For k8s/docker, these paths carry no meaning.

cmd/agent/main.go
ldemailly: (just) curious how different it is from %+v ?

platform/kube/inject/testdata/auth.cert-dir.yaml.injected
ldemailly: if the certs and config are written and updated differently it's better to have different paths

cmd/agent/main.go
rshriram: We could have the egress and sidecar as one element eliminating the need for a separate egress element. However, it requires the mixer to process outbound traffic and not just inbound traffic.

cmd/agent/main.go
rshriram: Yes. And if we do end up going to a common config root, my preference would be /var/lib/istio (mysql and Postgres are databases and they store their database files in /home if configured to. Iirc they default to /var/lib/mysql as well.)

sebastienvas:
build e2e/smoke-test

sebastienvas:
build pilot/e2e-smoketest

sebastienvas:
build pilot/e2e-smoketest

sebastienvas:
Jenkins job pilot/presubmit passed

costinm:
To clarify: I actually meant '$HOME', not /home - i.e. the actual dir can be set when the istio user is created to /var/istio or something else. I would assume distros will create their own packages at some point, with different conventions.

docker/BUILD
andraxylia: istio-agent, it's meant to work with other proxies as well.

docker/BUILD
andraxylia: istio-agent

For other proxies, we do not want to change all build files and docs.

platform/kube/inject/inject.go
andraxylia: Please open issue in addition to TODO and add it to the corresponding epic.

docker/BUILD
kyessenov: Ok, Bazel doesn't allow renaming binaries. Is this really important to assign a long name?

docker/BUILD
kyessenov: Bazel + go requires that cmd/agent be the binary name. cmd/istio-envoy-agent is rather long name for a binary.

platform/kube/inject/testdata/auth.cert-dir.yaml.injected
kyessenov: Isn't "etc" for configuration files? Why would we not put configuration into "etc"?

docker/BUILD
ldemailly: well you can rename the directory to istio-agent (or istio-envoy-agent)
with tab completion long names aren't really a big deal, and we don't expect users to manually run those binaries anyway

kyessenov: pilot-agent seems to be unique enough. istio-agent can be confused with CA node agent.

kyessenov:
PTAL:
- renamed agent and discovery to pilot-agent and pilot-discovery
- use `/etc/istio/` for `/config` mesh config, `/ingress-certs` certs, and `/envoy` configs
- use instance ID instead of pod name and namespace in the main file

kyessenov:
Fixes #945

kyessenov:
jenkins rebuild pilot/e2e-smoketest

PR #951, opened 2017-07-22
Change discovery to listen to all namespaces

andraxylia:
**Release note**:

```release-note
Multi-namespace support
```


platform/kube/controller.go
```diff
@@ -354,8 +354,16 @@ func generateServiceAccountID(sa string, ns string, domain string) string {
 // AppendServiceHandler implements a service catalog operation
 func (c *Controller) AppendServiceHandler(f func(*model.Service, model.Event)) error {
 	c.services.handler.Append(func(obj interface{}, event model.Event) error {
-		if svc := convertService(*obj.(*v1.Service), c.domainSuffix); svc != nil {
-			f(svc, event)
+		svc := *obj.(*v1.Service)
+
+		if svc.Namespace == "kube-system" || svc.Namespace == "istio-system" {
```
kyessenov: please no hard coding of special strings.

platform/kube/controller.go
```diff
@@ -73,26 +73,26 @@ func NewController(client kubernetes.Interface, mesh *proxyconfig.ProxyMeshConfi
 
 	out.services = out.createInformer(&v1.Service{}, options.ResyncPeriod,
 		func(opts meta_v1.ListOptions) (runtime.Object, error) {
-			return client.CoreV1().Services(options.Namespace).List(opts)
+			return client.CoreV1().Services(meta_v1.NamespaceAll).List(opts)
```
kyessenov: we need a test for this.

adapter/config/ingress/status.go
```diff
@@ -68,14 +71,21 @@ func NewStatusSyncer(mesh *proxyconfig.ProxyMeshConfig, client kubernetes.Interf
 	if mesh.IngressService != "" {
 		publishService = fmt.Sprintf("%v/%v", options.Namespace, mesh.IngressService)
 	}
+	glog.Infof("INGRESS STATUS publishService ", publishService)
```
rshriram: nit. isn't this glog.V(x) .. ?

bin/gen.sh
```diff
@@ -0,0 +1,2 @@
+cp bazel-genfiles/model/mock_config_gen_test.go model/
+cp -a bazel-genfiles/external/com_github_googleapis_googleapis/google/rpc/code.pb.go bazel-genfiles/external/com_github_googleapis_googleapis/google/rpc/error_details.pb.go bazel-genfiles/external/com_github_googleapis_googleapis/google/rpc/status.pb.go ../../github.com/googleapis/google/rpc
```
rshriram: This is your local dev environment file right? can you please skip?

bin/push-debug-images
```diff
@@ -0,0 +1,65 @@
...
+fi
+
+pushd docker
+  for image in app proxy proxy_init proxy_debug pilot; do
+    docker build -f "Dockerfile.${image}" -t "$hub/$image:$tag" .
+    docker push "$hub/$image:$tag"
+  done
+popd
+
+echo Pushed images to $hub with tag $tag
```
rshriram: Also remove this file please.

cmd/pilot-discovery/Gopkg.toml
```diff
@@ -0,0 +1,46 @@
...
+  branch = "master"
+  name = "github.com/spf13/cobra"
+
+[[constraint]]
+  branch = "costin-build"
+  name = "istio.io/api"
+
+[[constraint]]
+  name = "istio.io/pilot"
+  version = "0.2.0-pre"
```
rshriram: These 4 Gopkg.* files should be removed. We can add them in separate PR if need be.

model/mock_config_gen_test.go
```diff
@@ -0,0 +1,279 @@
+// Automatically generated by MockGen. DO NOT EDIT!
```
rshriram: And this file also goes right?

platform/kube/controller.go
```diff
@@ -399,8 +401,16 @@ func (c *Controller) GetIstioServiceAccounts(hostname string, ports []string) []
 // AppendServiceHandler implements a service catalog operation
 func (c *Controller) AppendServiceHandler(f func(*model.Service, model.Event)) error {
 	c.services.handler.Append(func(obj interface{}, event model.Event) error {
-		if svc := convertService(*obj.(*v1.Service), c.domainSuffix); svc != nil {
-			f(svc, event)
+		svc := *obj.(*v1.Service)
+
+		if svc.Namespace == meta_v1.NamespaceSystem {
```
rshriram: A comment here would be helpful

proxy/envoy/discovery.go
```diff
@@ -396,41 +396,44 @@ func (ds *DiscoveryService) ListEndpoints(request *restful.Request, response *re
...
 	}
 	writeResponse(response, out)
 }
 
-func (ds *DiscoveryService) parseRole(request *restful.Request) (proxy.Node, error) {
-	if sc := request.PathParameter(ServiceCluster); sc != ds.Mesh.IstioServiceCluster {
-		return proxy.Node{}, fmt.Errorf("unexpected %s %q", ServiceCluster, sc)
+func (ds *DiscoveryService) parseDiscoveryRequest(request *restful.Request) (string, string, proxy.Node, error) {
+	cluster := request.PathParameter(ServiceCluster)
+	if cluster != ds.Mesh.IstioServiceCluster {
```
rshriram: Please add a comment to clarify why this special case is needed.

proxy/envoy/discovery.go
```diff
@@ -396,41 +396,44 @@ func (ds *DiscoveryService) ListEndpoints(request *restful.Request, response *re
...
 	if !cached {
-		role, err := ds.parseRole(request)
+		cluster, node, role, err := ds.parseDiscoveryRequest(request)
 		if err != nil {
-			errorResponse(response, http.StatusNotFound, err.Error())
+			errorResponse(response, http.StatusNotFound, "CDS " + err.Error())
 			return
 		}
+		glog.V(3).Infof("CDS Discovery request to ListClusters for service_cluster %s, service_node %s, role %s",
+			cluster, node, role.Type)
```
rshriram: Remove? Even if accidentally turned on, this will flood the system with logs, at a 100ms refresh frequency

proxy/envoy/discovery.go
```diff
@@ -443,16 +446,19 @@ func (ds *DiscoveryService) ListListeners(request *restful.Request, response *re
...
-		role, err := ds.parseRole(request)
+		cluster, node, role, err := ds.parseDiscoveryRequest(request)
 		if err != nil {
-			errorResponse(response, http.StatusNotFound, err.Error())
+			errorResponse(response, http.StatusNotFound, "LDS " + err.Error())
 			return
 		}
+		glog.V(3).Infof("LDS Discovery request to ListListeners for service_cluster %s, service_node %s, role %s",
+			cluster, node, role.Type)
+
```
rshriram: also remove

proxy/envoy/discovery.go
```diff
@@ -467,30 +473,32 @@ func (ds *DiscoveryService) ListRoutes(request *restful.Request, response *restf
...
 	if !cached {
-		role, err := ds.parseRole(request)
+		cluster, node, role, err := ds.parseDiscoveryRequest(request)
 		if err != nil {
-			errorResponse(response, http.StatusNotFound, err.Error())
+			errorResponse(response, http.StatusNotFound, "RDS " + err.Error())
 			return
 		}
+		glog.V(3).Infof("RDS Discovery request to ListRoutes for service_cluster %s, service_node %s, role %s",
+			cluster, node, role.Type)
```
rshriram: also remove

proxy/envoy/discovery.go
```diff
@@ -501,12 +509,17 @@ func (ds *DiscoveryService) ListRoutes(request *restful.Request, response *restf
...
+
 	if err != nil {
-		errorResponse(response, http.StatusNotFound, err.Error())
+		errorResponse(response, http.StatusNotFound, "ListSecrets " + err.Error())
 		return
 	}
 
+	glog.V(3).Infof("ListSecrets Discovery request to ListSecret for service_cluster %s, service_node %s, role %s",
+		cluster, node, role.Type)
+
```
rshriram: also remove

proxy/envoy/discovery.go
```diff
@@ -515,20 +528,21 @@ func (ds *DiscoveryService) ListSecret(request *restful.Request, response *restf
 	_, secret := buildIngressRoutes(ds.Mesh, ds, ds)
 
 	if secret == "" {
+		glog.V(3).Infof("Secret is not set")
```
rshriram: definitely remove please.

proxy/envoy/ingress.go
```diff
@@ -45,6 +45,7 @@ func buildIngressRoutes(mesh *proxyconfig.ProxyMeshConfig,
 	discovery model.ServiceDiscovery,
 	config model.IstioConfigStore) (HTTPRouteConfigs, string) {
 	ingressRules := config.IngressRules()
+	glog.V(3).Infof("buildIngressRoute from %d rules ", len(ingressRules))
```
rshriram: remove?

proxy/envoy/ingress.go
```diff
@@ -55,7 +56,13 @@ func buildIngressRoutes(mesh *proxyconfig.ProxyMeshConfig,
 	rules := config.RouteRulesBySource(nil)
 
 	for _, rule := range ingressRules {
+		glog.V(3).Infof("Process rule %s , destination %s", rule.Name, rule.Destination)
 		routes, tls, err := buildIngressRoute(mesh, rule, discovery, rules)
+		if tls != "" {
+			glog.V(3).Infof("TLS is set for rule %s , destination %s", rule.Name, rule.Destination)
+		} else {
+			glog.V(3).Infof("TLS is empty for rule %s , destination %s", rule.Name, rule.Destination)
+		}
```
rshriram: remove debug stuff ?

test/grpcecho/echo.pb.go
```diff
@@ -0,0 +1,160 @@
+// Code generated by protoc-gen-go.
```
rshriram: do we check in generated pb files?

test/integration/driver.go
```diff
@@ -86,6 +88,8 @@ func init() {
 
 	// If specified, only run one test
 	flag.StringVar(&testType, "testtype", "", "Select test to run (default is all tests)")
+
+	flag.IntVar(&params.DlvPort, "dlvport", 0, "Dlv port for debugging")
```
rshriram: DlvPort ? may be use full form? or just call it debugport.

test/integration/http.go
```diff
@@ -59,6 +59,11 @@ func (r *http) makeRequests() error {
 	funcs := make(map[string]func() status)
 	for _, src := range testPods {
 		for _, dst := range testPods {
+			if src == "t" && dst == "t" {
+				// this is flaky in minikube
+				continue
+			}
+
```
rshriram: I guess this should be removed?

test/integration/infra.go
```diff
@@ -55,11 +56,13 @@ type infra struct {
 	Ingress bool
 	Egress  bool
 	Zipkin  bool
+	DlvPort int
```
rshriram: DebugPort

Gopkg.lock
```diff
@@ -183,6 +207,12 @@
   revision = "2f5df55504ebc322e4d52d34df6a1f5b503bf26d"
 
 [[projects]]
+  branch = "master"
```
kyessenov: Can we remove this from the PR? If it's necessary to update, please stacks the PRs.

test/integration/ingress.go
```diff
@@ -94,24 +95,24 @@ func (t *ingress) run() error {
 
 	funcs := make(map[string]func() status)
 	funcs["Ingress status IP"] = t.checkIngressStatus
-	funcs["Route rule for /c"] = t.checkRouteRule
+//	funcs["Route rule for /c"] = t.checkRouteRule
```
rshriram: Why is this commented out?

test/integration/tcp.go
```diff
@@ -44,6 +44,10 @@ func (t *tcp) run() error {
 	funcs := make(map[string]func() status)
 	for _, src := range testPods {
 		for _, dst := range testPods {
+			if src == "t" && dst == "t" {
+				// this is flaky in minikube
+				continue
+			}
```
rshriram: remove

test/integration/testdata/egress-proxy.yaml.tmpl
```diff
@@ -8,7 +8,7 @@ spec:
   - name: http
     port: 80
   selector:
-    app: istio-egress
```
rshriram: lets just retain these names as is.. Even though its in istio namespace. k8s does the same.. kube-dns, kube-apiserver, etc.

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -1,14 +1,53 @@
 # pilot service for discovery
+
+{{if .DlvPort}}
```
rshriram: what is this for and what does this do? Some comments in file would be helpful

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -32,10 +71,16 @@ spec:
       containers:
       - name: discovery
         image: {{.Hub}}/pilot:{{.Tag}}
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: Always
```
rshriram: why is this being changed?

adapter/config/ingress/controller.go
```diff
@@ -62,10 +62,10 @@ func NewController(client kubernetes.Interface, mesh *proxyconfig.ProxyMeshConfi
 	informer := cache.NewSharedIndexInformer(
 		&cache.ListWatch{
 			ListFunc: func(opts meta_v1.ListOptions) (runtime.Object, error) {
-				return client.ExtensionsV1beta1().Ingresses(options.Namespace).List(opts)
+				return client.ExtensionsV1beta1().Ingresses(meta_v1.NamespaceAll).List(opts)
```
kyessenov: can you please make this customizable? It's a matter of passing through the namespace to controllers through the code.

adapter/config/ingress/status.go
```diff
@@ -28,6 +29,8 @@ import (
 
 	proxyconfig "istio.io/api/proxy/v1/config"
 	"istio.io/pilot/platform/kube"
+	"github.com/golang/glog"
+	betaext "k8s.io/api/extensions/v1beta1"
```
kyessenov: please sort the imports with istio.io in a separate group.

adapter/config/ingress/status.go
```diff
@@ -17,6 +17,7 @@ package ingress
 import (
 	"fmt"
 
+	v1 "k8s.io/api/core/v1"
```
kyessenov: unnecessary import name `v1`

adapter/config/ingress/status.go
kyessenov: missing log level

adapter/config/ingress/status.go
```diff
@@ -68,14 +71,21 @@ func NewStatusSyncer(mesh *proxyconfig.ProxyMeshConfig, client kubernetes.Interf
...
+	}
+
 	sync := status.NewStatusSyncer(status.Config{
 		Client:              client,
 		IngressLister:       store.IngressLister{Store: informer.GetStore()},
 		ElectionID:          ingressElectionID, // TODO: configurable?
 		PublishService:      publishService,
 		DefaultIngressClass: defaultIngressClass,
 		IngressClass:        ingressClass,
+		CustomIngressStatus: customIngressStatus,
```
kyessenov: what does this do?

bin/gen.sh
```diff
@@ -0,0 +1,2 @@
+cp bazel-genfiles/model/mock_config_gen_test.go model/
```
kyessenov: separate this into another PR.

test/util/kubernetes.go
```diff
@@ -71,40 +71,43 @@ func GetPods(cl kubernetes.Interface, ns string) []string {
...
+		for n := 0; ; n++ {
+			list, err := cl.CoreV1().Pods(ns).List(meta_v1.ListOptions{})
+			if err != nil {
+				return pods, err
+			}
+			items = list.Items
+			ready := true
+
+			for _, pod := range items {
+				if pod.Status.Phase != "Running" {
```
rshriram: I seem to recall adding something along the lines of checking every container's status in the pod, as opposed to just the pod status (because k8s says pod is running even if only one container is running).

bin/push-debug-images
kyessenov: What's the purpose of this script as opposed to improving push-docker?

cmd/istioctl/main.go
```diff
@@ -301,7 +301,7 @@ func init() {
 	}
 	rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "c", defaultKubeconfig,
 		"Kubernetes configuration file")
-	rootCmd.PersistentFlags().StringVarP(&istioSystem, "namespace", "n", v1.NamespaceDefault,
+	rootCmd.PersistentFlags().StringVarP(&istioSystem, "namespace", "n", kube.IstioNamespace,
```
kyessenov: I think the meaning of this flag changed with CRDs to be the place where CRDs are stored. this has effect on rule match conditions.

model/mock_config_gen_test.go
kyessenov: do not commit this file since it should be generated by bazel.

platform/kube/controller.go
```diff
@@ -38,6 +38,8 @@ const (
 	NodeRegionLabel = "failure-domain.beta.kubernetes.io/region"
 	// NodeZoneLabel is the well-known label for kubernetes node zone
 	NodeZoneLabel = "failure-domain.beta.kubernetes.io/zone"
+	// NamespaceDefault means the object is in the default namespace which is applied when not specified by clients
```
kyessenov: comment should say `IstioNamespace ...`

platform/kube/controller.go
kyessenov: Please don't hardcode namespace all unless absolutely necessary.

platform/kube/controller.go
kyessenov: There's a constant defined somewhere that lists the special namespaces such as kube system and kube public. Are we sure we want to exclude those? This might prevent applying routing rules to kubernetes API server, for example.

platform/kube/controller_test.go
```diff
@@ -102,10 +102,15 @@ func TestServices(t *testing.T) {
 	eventually(func() bool {
 		out := sds.Services()
 		glog.Info("Services: %#v", out)
-		return len(out) == 1 &&
-			out[0].Hostname == hostname &&
-			len(out[0].Ports) == 1 &&
-			out[0].Ports[0].Protocol == model.ProtocolHTTP
+
+		for _, item := range out {
```
kyessenov: does not check for out = 1 anymore? Probably related to the fact that test listens on all namespaces, which it shouldn't.

proxy/envoy/discovery.go
```diff
@@ -396,41 +396,44 @@ func (ds *DiscoveryService) ListEndpoints(request *restful.Request, response *re
 		}
 		var err error
 		if out, err = json.MarshalIndent(hosts{Hosts: hostArray}, " ", " "); err != nil {
-			errorResponse(response, http.StatusInternalServerError, err.Error())
+			errorResponse(response, http.StatusInternalServerError, "SDS " + err.Error())
```
kyessenov: it's now called EDS.

proxy/envoy/discovery.go
```diff
@@ -515,20 +528,21 @@ func (ds *DiscoveryService) ListSecret(request *restful.Request, response *restf
...
 		errorResponse(response, http.StatusNotFound,
-			fmt.Sprintf("Failed to read the secret: %s", err))
+			fmt.Sprintf("ListSecrets Failed to read the secret: %s", err))
 		return
 	}
 
 	out, err := json.Marshal(tls)
 	if err != nil {
-		errorResponse(response, http.StatusInternalServerError, err.Error())
+		errorResponse(response, http.StatusInternalServerError, "ListSecrets " + err.Error())
```
kyessenov: These changes do not contribute to the goal of this PR. Move it to another PR.

test/grpcecho/echo.pb.go
kyessenov: No. Remove these files.

test/integration/access.go
```diff
@@ -74,10 +74,17 @@ func (a *accessLogs) check(infra *infra) error {
 
 				pod := infra.apps[app][0]
 				container := inject.ProxyContainerName
-				if app == "mixer" {
+				ns := infra.Namespace
+				switch app {
+				case "mixer":
 					container = "mixer"
+					ns = infra.IstioNamespace
```
kyessenov: How is this test going to pass for multiple people if they all use same namespace?

test/integration/driver.go
```diff
@@ -74,8 +74,10 @@ func init() {
 		"CA Docker image")
 	flag.StringVar(&params.MixerImage, "mixer", mixerImage,
 		"Mixer Docker image")
+	flag.StringVar(&params.IstioNamespace, "ns", "",
+		"Namespace in which to install Istio components (empty to create/delete temporary one)")
```
kyessenov: ah, ok, so it's a flag.

test/integration/http.go
```diff
@@ -59,6 +59,11 @@ func (r *http) makeRequests() error {
 	funcs := make(map[string]func() status)
 	for _, src := range testPods {
 		for _, dst := range testPods {
+			if src == "t" && dst == "t" {
```
kyessenov: but it's not on GKE. Please don't reduce test coverage.

test/integration/infra.go
```diff
@@ -225,8 +250,10 @@ func (infra *infra) clientRequest(app, url string, count int, extra string) resp
 	}
 
 	pod := infra.apps[app][0]
-	request, err := util.Shell(fmt.Sprintf("kubectl exec %s --kubeconfig %s -n %s -c app -- client -url %s -count %d %s",
-		pod, kubeconfig, infra.Namespace, url, count, extra))
+	cmd := fmt.Sprintf("kubectl exec %s --kubeconfig %s -n %s -c app -- client -url %s -count %d %s",
+		pod, kubeconfig, infra.Namespace, url, count, extra)
+	glog.Info(cmd)
+	request, err := util.Shell(cmd)
```
kyessenov: util shell already prints the command, so the extra log is not needed.

test/integration/ingress.go
```diff
@@ -154,7 +155,7 @@ func (t *ingress) run() error {
 
 // checkRouteRule verifies that version splitting is applied to ingress paths
 func (t *ingress) checkRouteRule() status {
-	url := fmt.Sprintf("http://%s/c", ingressServiceName)
+	url := fmt.Sprintf("http://%s.%s/c", ingressServiceName, t.IstioNamespace)
```
kyessenov: remove the namespace since both pods are in the same namespace

test/integration/testdata/ca.yaml.tmpl
```diff
@@ -20,9 +20,3 @@ spec:
       - name: istio-ca-container
         image: {{.CaImage}}
         imagePullPolicy: IfNotPresent
-        env:
```
kyessenov: why?

test/integration/testdata/config.yaml.tmpl
```diff
@@ -4,12 +4,13 @@ metadata:
   name: istio
 data:
   mesh: |-
-    discoveryAddress: istio-pilot:8080
+    discoveryAddress: istio-pilot.{{.IstioNamespace}}.svc.cluster.local:8080
+    egressProxyAddress: istio-egress.{{.IstioNamespace}}.svc.cluster.local:80
 {{if .Mixer}}
-    mixerAddress: istio-mixer:9091
+    mixerAddress: istio-mixer.{{.IstioNamespace}}.svc.cluster.local:9091
```
kyessenov: no need to put svc cluster local

test/integration/testdata/egress-proxy.yaml.tmpl
```diff
@@ -8,7 +8,7 @@ spec:
...
+    app: egress
```
kyessenov: also this seems unnecessary and is inconsistent with installation files

test/integration/testdata/pilot.yaml.tmpl
kyessenov: why?

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -50,8 +95,17 @@ spec:
...
           mountPath: /etc/istio/config
+{{if .DlvPort}}
+        - mountPath: "/data/delve"
+          name: dlv
+{{end}}
       volumes:
       - name: config-volume
         configMap:
           name: istio
+{{if .DlvPort}}
```
kyessenov: I feel like DLV changes should be in a separate PR as well.

test/integration/testdata/rbac-beta.yaml.tmpl
```diff
@@ -12,7 +12,10 @@ rules:
   resources: ["istiokinds"]
   verbs: ["get", "list", "watch"]
 - apiGroups: ["apiextensions.k8s.io"]
-  resources: ["customresourcedefinitions"]
+  resources: ["*"]
```
kyessenov: is there a reason to change these?

platform/kube/controller.go
andraxylia: It's not hardcoded.

Gopkg.lock
andraxylia: This file will likely disappear from the change after I merge with master again, it is now here because at the time Costin's PR was not merged.

adapter/config/ingress/status.go
andraxylia: This is one time at startup, and I think it is pretty important. The status IP of the ingress is calculated based on this.

andraxylia: The files related to the go build will disappear after the merge.

The other logs I found important during troubleshooting, and I think we should leave them in. I generally used V(3).

adapter/config/ingress/status.go
```diff
@@ -68,14 +71,21 @@ func NewStatusSyncer(mesh *proxyconfig.ProxyMeshConfig, client kubernetes.Interf
...
```
andraxylia: Yes, it is related to upgrading the ingress.
This could be done separately, but giving that we did not enforce this rule in the past, and we approved reviews containing refactoring + fixes for unrelated features, I will claim the same privilege and save the extra effort.

bin/gen.sh
andraxylia: Go build stuff, already in master.

bin/push-debug-images
andraxylia: This is useful for mac osx development, it allows building docker images from mac osx.

cmd/pilot-discovery/Gopkg.toml
andraxylia: Will disappear after the merge.

model/mock_config_gen_test.go
andraxylia: yup.

proxy/envoy/discovery.go
andraxylia: It's V3, and it's needed. I can change it to higher level, but I will not remove it.

proxy/envoy/discovery.go
andraxylia: The Ingress handling of secret caused me a lot of pain.
We should have better logs in general.

test/integration/driver.go
andraxylia: Sure, I can change it to debugport.

test/integration/http.go
andraxylia: It slows down minikube tests run a lot and do not provide anything useful. There is no proxy involved in t--> t communication.

I am all for deterministic tests.

test/integration/ingress.go
andraxylia: Will put back.

test/integration/testdata/rbac-beta.yaml.tmpl
andraxylia: You asked me to make this change in istio/istio to be future proof, made the same change in pilot.

test/integration/testdata/pilot.yaml.tmpl
andraxylia: Ack, not needed.

test/integration/testdata/config.yaml.tmpl
andraxylia: ok.

test/integration/testdata/egress-proxy.yaml.tmpl
andraxylia: Should be consistent with istio-ingress.

test/integration/testdata/ca.yaml.tmpl
andraxylia: So that CA listens to all namespaces.

test/integration/testdata/rbac-beta.yaml.tmpl
kyessenov: Well, this is the future while istio/istio one is the past :) I'd keep this as precise as possible but since it's just too painful to coordinate with istio/istio, leave the released ones more open.

test/integration/testdata/pilot.yaml.tmpl
kyessenov: There's quite a big template addition from what I can tell. I expect some documentation how to make use of these, too.

test/integration/testdata/ca.yaml.tmpl
kyessenov: Wouldn't that break when I run two tests at the same time?

proxy/envoy/discovery.go
```diff
@@ -467,30 +473,32 @@ func (ds *DiscoveryService) ListRoutes(request *restful.Request, response *restf
...
+				fmt.Sprintf("RDS Unexpected %s %q", RouteConfigName, routeConfigName))
 			return
 		}
 
 		httpRouteConfigs := buildRDSRoutes(ds.Mesh, role, ds, ds)
 		routeConfig, ok := httpRouteConfigs[port]
 		if !ok {
 			errorResponse(response, http.StatusNotFound,
-				fmt.Sprintf("Missing route config for port %d", port))
+				fmt.Sprintf("rds Missing route config for port %d", port))
```
kyessenov: Please fix the error. I don't think this needs to be here, too.

test/integration/ingress.go
andraxylia: But this is to the ingress service, which is in a different namespace.

test/integration/infra.go
andraxylia: ok

test/integration/http.go
andraxylia: We are not testing anything related to istio here, so there is zero benefit.

test/integration/access.go
andraxylia: IstioNamespace is generated.

platform/kube/controller.go
andraxylia: If we ever need to apply routing rules to kube api server, we will re-visit this. For now, there is no such requirement.

platform/kube/controller.go
andraxylia: ack.

platform/kube/controller.go
kyessenov: Once it's cluster-wide, this becomes a requirement. It should be possible to use kubectl from within istio-enabled pods.

test/integration/http.go
kyessenov: It could help with debugging #978...

proxy/envoy/discovery.go
andraxylia: It's an extra check.

test/integration/testdata/egress-proxy.yaml.tmpl
andraxylia: I made the change to have consistent labeling scheme with istio-ingress.

test/integration/testdata/pilot.yaml.tmpl
andraxylia: I will write a separate doc in a separate PR. It is helpful for remote debugging of an existing pilot container. Briefly, it allows to open an extra port to which a remote debugger can attach. It also mounts a volume containing dlv.

test/integration/testdata/pilot.yaml.tmpl
andraxylia: reverted

test/util/kubernetes.go
andraxylia: Right, that would be a good change to have

adapter/config/ingress/status.go
andraxylia: It's used below

adapter/config/ingress/status.go
andraxylia: ok

adapter/config/ingress/status.go
andraxylia: This works around a bug in k8s where there is no null check if that field is null and it causes the watcher to crash.

bin/push-debug-images
andraxylia: this is only for go build, while push-docker uses bazel.

test/integration/testdata/egress-proxy.yaml.tmpl
ldemailly: then we should change ingress to be istio-ingress?

test/integration/zipkin.go
```diff
@@ -86,7 +86,8 @@ func (t *zipkin) verifyTraces() error {
 		for _, id := range t.traces {
 			response := t.infra.clientRequest(
 				"t",
-				fmt.Sprintf("http://zipkin:9411/api/v1/traces?annotationQuery=guid:x-request-id=/%s", id),
+				fmt.Sprintf("http://zipkin.%s:9411/api/v1/traces?annotationQuery=guid:x-request-id=/%s",
+					t.IstioNamespace, id),
```
ldemailly: what if the user wants to reuse their pre existing zipkin in some other namespace ?

platform/kube/controller_test.go
andraxylia: Yes, controllers listen on all namespaces.

proxy/envoy/discovery.go
andraxylia: ok

test/integration/testdata/pilot.yaml.tmpl
andraxylia: will do proper doc

test/integration/zipkin.go
andraxylia: They should point to it in the configMap.

{{if .Zipkin}}
    zipkinAddress: zipkin.{{.IstioNamespace}}:9411
{{end}}

ldemailly: oh nvm this is just a test file sorry

bin/push-debug-images
rshriram: atleast rename this file please..

proxy/envoy/discovery.go
rshriram: Yes, but for what? I meant some explanation of what this is excluding.. for other contributors to understand the code.

proxy/envoy/discovery.go
rshriram: Sure. Then change to higher level please. Like 6 or something.

proxy/envoy/discovery.go
rshriram: I agree. May be increase log level to 6 or something? Generally, everything in the *DS path loops needs to have high log level.. So that it is easier to debug in stages..

test/integration/http.go
rshriram: I sort of agree with @andraxylia here. What is it that this test is accomplishing @kyessenov ?

bin/push-debug-images
andraxylia: gobuild-push-docker?

test/integration/http.go
kyessenov: It tests that istio does not mess up things that don't have proxy. This might be obvious but since we do more and more invasive things (like cluster-wide), it's good to verify that.

proxy/envoy/discovery.go
andraxylia: I will add a comment, but remark this is not my code, the check was already there. I merely changed the code to return all the fields and display proper error message.

proxy/envoy/discovery.go
andraxylia: Changed to 5.

test/integration/http.go
andraxylia: ok, will put it back if it helps anybody.

adapter/config/ingress/controller.go
andraxylia: As discussed, there could be issues if multiple ingress controllers are deployed, like when multiple people run tests in the same time, and they all claim all ingress resources. This will not be the case in production, but it may affect our testing.

I will pass a flag from the test to limit the ingress to a single namespace.

bin/push-debug-images
kyessenov: Please don't duplicate shell scripts. Just add a flag to the original, or even better fix bazel build for mac osx. Again, this has nothing to do with this PR.

test/integration/testdata/egress-proxy.yaml.tmpl
andraxylia: It does not matter beyond consistency, it's just a label we added.

bin/gen.sh
rshriram: this doesn't seem to be in master.

andraxylia:
I actually removed the t-->t skip, but I will put it back. It fails consistently in minikube, and our scripts keep retrying for 100 times. Being able to run e2e tests in minikube is more important than catching some random issue.

ayj:
> It fails consistently in minikube, and our scripts keep retrying for 100 times. Being able to run e2e tests in minikube is more important than catching some random issue.

Apologies if this was answered elsewhere but does bin/e2e.sh fail consistently without these changes? "Catching some random issue" seems fairly important considering the interest in [ProxyBypass](https://docs.google.com/document/d/1fovubmdW-cjFjemjY-R-hVdct7LxOAkvnhBM-h1UhmA/edit#)

andraxylia:
The tcp call from t-->t fails consistently for me in minikube with kubernetes-version=v1.7.0.

test/util/kubernetes.go
andraxylia: The container check got in when I merged.

rshriram:
```release-note
Control multiple namespaces with single Istio install
```

proxy/envoy/discovery.go
```diff
@@ -396,41 +396,45 @@ func (ds *DiscoveryService) ListEndpoints(request *restful.Request, response *re
...
 }
 
-func (ds *DiscoveryService) parseRole(request *restful.Request) (proxy.Node, error) {
-	if sc := request.PathParameter(ServiceCluster); sc != ds.Mesh.IstioServiceCluster {
-		return proxy.Node{}, fmt.Errorf("unexpected %s %q", ServiceCluster, sc)
+func (ds *DiscoveryService) parseDiscoveryRequest(request *restful.Request) (string, string, proxy.Node, error) {
+	cluster := request.PathParameter(ServiceCluster)
+	// request has to match the IstioServiceCluster (default is "istio-proxy")
+	if cluster != ds.Mesh.IstioServiceCluster {
+		return cluster, "", proxy.Node{}, fmt.Errorf("unexpected %s %q", ServiceCluster, cluster)
```
rshriram: while you are at this, can you get rid of this check completely? Why do we need to check if the service cluster is istio-proxy? In VM case, this is most likely going to fail, if the user customizes their envoy startup. (cc @costinm )
This is an artificial requirement we imposed on ourselves that everything has to be called istio-proxy.

proxy/envoy/discovery.go
```diff
@@ -443,16 +447,18 @@ func (ds *DiscoveryService) ListListeners(request *restful.Request, response *re
...
 	if !cached {
-		role, err := ds.parseRole(request)
+		cluster, node, role, err := ds.parseDiscoveryRequest(request)
 		if err != nil {
-			errorResponse(response, http.StatusNotFound, err.Error())
+			errorResponse(response, http.StatusNotFound, "LDS "+err.Error())
 			return
 		}
+		glog.V(5).Infof("LDS Discovery request to ListListeners for service_cluster %s, service_node %s, role %s",
+			cluster, node, role.Type)
```
rshriram: In light of this, its all the more nicer to get rid of that check above.

proxy/envoy/watcher.go
```diff
@@ -159,11 +159,13 @@ func (w *watcher) UpdateIngressSecret(ctx context.Context) error {
 		return multierror.Prefix(err, "failed to read request body")
 	}
 	if len(tlsData) == 0 {
+		glog.Errorf("failed to get data, zero data")
```
rshriram: nit: failed to get TLS data.

proxy/envoy/watcher.go
```diff
@@ -159,11 +159,13 @@ func (w *watcher) UpdateIngressSecret(ctx context.Context) error {
...
 		return nil
 	}
 
 	var tls model.TLSSecret
 	if err = json.Unmarshal(tlsData, &tls); err != nil {
+		glog.Errorf("failed to unmarshal")
```
rshriram: FAiled to unmarshal TLS secret

proxy/envoy/discovery.go
andraxylia: The check is to verify if the request matches what is configured in pilot, it does not match against "istio-proxy".

test/integration/infra.go
```diff
@@ -78,27 +81,44 @@ func (infra *infra) setup() error {
...
 			return err
 		}
 		return nil
 	}
-	if err := deploy("rbac-beta.yaml.tmpl"); err != nil {
+	if err := deploy("rbac-beta.yaml.tmpl", infra.IstioNamespace); err != nil {
 		return err
 	}
-	if err := deploy("config.yaml.tmpl"); err != nil {
+	if err := deploy("config.yaml.tmpl", infra.Namespace); err != nil {
```
rshriram: shouldn't this be IstioNamespace as well?

test/integration/ingress.go
```diff
@@ -92,24 +92,24 @@ func (t *ingress) run() error {
 
 	funcs := make(map[string]func() status)
 	funcs["Ingress status IP"] = t.checkIngressStatus
-	funcs["Route rule for /c"] = t.checkRouteRule
+	//funcs["Route rule for /c"] = t.checkRouteRule
```
rshriram: why commented out?

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -1,14 +1,53 @@
 # pilot service for discovery
+
+{{if .DebugPort}}
+kind: PersistentVolume
+apiVersion: v1
+metadata:
+  name: dlv
```
rshriram: dbg

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -1,14 +1,53 @@
...
+    type: local
+spec:
+  storageClassName: manual
+  persistentVolumeReclaimPolicy: Delete
+  capacity:
+    storage: 10Gi
+  accessModes:
+    - ReadWriteOnce
+  hostPath:
+    path: "/data/delve"
```
rshriram: typo.. May be change to /data/debug

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -1,14 +1,53 @@
...
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: dlv-claim
```
rshriram: s/dlv/dbg

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -1,14 +1,53 @@
...
 spec:
+{{if .DebugPort}}
+  type: LoadBalancer
+{{end}}
   ports:
   - port: 8080
     name: http-discovery
+{{if .DebugPort}}
+  - port: {{.DebugPort}}
+    name: tcp-dlv
```
rshriram: dbg

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -35,9 +74,15 @@ spec:
...
         image: {{.Hub}}/pilot:{{.Tag}}
         imagePullPolicy: IfNotPresent
-        args: ["discovery", "-v", "{{.Verbosity}}"]
+        args: ["discovery", "-v", "{{.Verbosity}}", "-a", "{{.Namespace}}"]
         ports:
         - containerPort: 8080
+{{if .DebugPort}}
+        - containerPort: {{.DebugPort}}
+        volumeMounts:
+          - mountPath: "/data/delve"
```
rshriram: delve->debug

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -35,9 +74,15 @@ spec:
...
+            name: dlv
```
rshriram: dbg

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -52,8 +97,17 @@ spec:
         volumeMounts:
         - name: config-volume
           mountPath: /etc/istio/config
+{{if .DebugPort}}
+        - mountPath: "/data/delve"
```
rshriram: sme as above

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -52,8 +97,17 @@ spec:
...
+          name: dlv
+{{end}}
       volumes:
       - name: config-volume
         configMap:
           name: istio
+{{if .DebugPort}}
+      - name: dlv
+        persistentVolumeClaim:
+          claimName: dlv-claim
```
rshriram: same as above

rshriram:
few more nits and then this is ready to go.

test/integration/tcp.go
ijsnellf: I usually end up putting similar code in when I am developing. Can we leave it in?

rshriram:
```release-note
Support for multiple namespaces
```

cmd/istioctl/main.go
```diff
@@ -288,6 +303,10 @@ func init() {
 	}
 	rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "c", defaultKubeconfig,
 		"Kubernetes configuration file")
+
+	rootCmd.PersistentFlags().StringVarP(&istioNamespace, "istioNamespace", "ns", kube.IstioNamespace,
```
ldemailly: ins or in ? (to be consistent with -n being the app namespace)

ldemailly:
you need `make fmt` (and also please change the flag for istionamespace)

cmd/istioctl/main.go
andraxylia: changed to -i.

test/integration/infra.go
andraxylia: The config map must be deployed in the app namespaces, as well as in the istio namespace, istio services read it as well, or we may have apps deployed in the istio namespace.

test/integration/tcp.go
andraxylia: Left it in.

adapter/config/crd/controller.go
```diff
@@ -47,6 +47,10 @@ type cacheHandler struct {
 // NewController creates a new Kubernetes controller for CRDs
 // Use "" for namespace to listen for all namespace changes
 func NewController(client *Client, options kube.ControllerOptions) model.ConfigStoreCache {
+
+	glog.V(2).Infof("CRD controller running in namespace %s, watching app namespaces %s",
+		options.Namespace, options.AppNamespace)
```
ldemailly: maybe you should remove Namespace entirely and only have AppNamespace and IstioNamespace ?

cmd/istioctl/main.go
```diff
@@ -319,7 +319,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "c", defaultKubeconfig,
 		"Kubernetes configuration file")
 
-	rootCmd.PersistentFlags().StringVarP(&istioNamespace, "istioNamespace", "ns", kube.IstioNamespace,
+	rootCmd.PersistentFlags().StringVarP(&istioNamespace, "istioNamespace", "i", kube.IstioNamespace,
```
ldemailly: thanks, though we may want to keep -i for something more likely to be changed by our users (like for some -include of some sort... afaik only test scripts would ever change the istionamespace ?)
you don't like `-ins` ?

adapter/config/crd/controller.go
andraxylia: namespace is a nice name to indicate where the pod is deployed and where container is running, so I will not rename.

test/integration/testdata/ca.yaml.tmpl
andraxylia: no

platform/kube/controller.go
```diff
@@ -37,12 +37,15 @@ const (
...
 	NodeZoneLabel = "failure-domain.beta.kubernetes.io/zone"
+	// IstioNamespace used by default for Istio cluster-wide installation
+	IstioNamespace = "istio-system"
 )
 
 // ControllerOptions stores the configurable attributes of a Controller.
 type ControllerOptions struct {
 	// Namespace to restrict controller to (empty to disable restriction)
 	Namespace    string
+	AppNamespace string
```
ldemailly: please document what that is exactly and how it relates to 'Namespace' and what changed in the meaning of 'Namespace' if anything
I thought (from the code elsewhere) Namespace was IstioNamespace but it seems to be (sometimes?) watched namespace ? for instance what does empty Namespace and non empty AppNamespace do, vs empty,empty and non empty, non empty...

adapter/config/crd/controller.go
ldemailly: see my other question/request about commenting more which is which in controller.go, also not sure your explanation (where the pod is deployed) matches the comment above "Use "" for namespace to listen for all namespace changes"

kyessenov: Why does controller need to know which namespace it runs in? I should be able to run it out of cluster.

cmd/istioctl/main.go
kyessenov: I suggest not to use short flags at all, except when it's obvious what they mean.

test/integration/infra.go
kyessenov: Config map is picked up at startup by proxy agents. Can kubernetes mount a configmap from another namespace?

kyessenov:
Can the DLV changes be stashed into another PR?

platform/kube/controller.go
andraxylia: Empty "" means all k8s namespaces, it is a k8s constant.

andraxylia:
@rshriram @kyessenov PTAL, sanity is passing. I can address minor comments in a subsequent PR.

sebastienvas:
I signed it

cmd/istioctl/main.go
rshriram: +1. This is not intuitive. --istio-namespace is better.

adapter/config/crd/controller.go
andraxylia: Indeed, it does not need to know, but in the overall code I had seen the pattern of using downward API and passing POD_NAMESPACE, which then gets translated into options.Namespace, for all controllers.

platform/kube/controller.go
andraxylia: k8s.io/apimachinery/pkg/apis/meta/v1.NamespaceAll is ""

cmd/pilot-discovery/main.go
```diff
@@ -183,6 +183,8 @@ func init() {
 		fmt.Sprintf("File name for Istio mesh configuration"))
 	discoveryCmd.PersistentFlags().StringVarP(&flags.controllerOptions.Namespace, "namespace", "n", "",
 		"Select a namespace for the controller loop. If not set, uses ${POD_NAMESPACE} environment variable")
+	discoveryCmd.PersistentFlags().StringVarP(&flags.controllerOptions.AppNamespace, "app namespace", "a", "",
+		"Restrict the applications namespace that controller manages, do n")
```
kyessenov: typo: `do n`?

model/config.go
```diff
@@ -43,6 +43,9 @@ type ConfigMeta struct {
 	// (security domains, fault domains, organizational domains)
 	Namespace string `json:"namespace,omitempty"`
 
+	// Namespace where istio control plane is installed
+	IstioNamespace string `json:"istioNamespace,omitempty"`
```
kyessenov: Is this really necessary as a general config metadata? This struct is mostly a copy of k8s ObjectMeta and this is deviating from that.

platform/kube/controller.go
kyessenov: Again, please explain in the comment what this is and how it is different from `Namespace`

platform/kube/controller.go
kyessenov: shoudn't this be options.Namespace?

platform/kube/controller.go
```diff
@@ -416,8 +419,17 @@ func (c *Controller) GetIstioServiceAccounts(hostname string, ports []string) []
 // AppendServiceHandler implements a service catalog operation
 func (c *Controller) AppendServiceHandler(f func(*model.Service, model.Event)) error {
 	c.services.handler.Append(func(obj interface{}, event model.Event) error {
-		if svc := convertService(*obj.(*v1.Service), c.domainSuffix); svc != nil {
-			f(svc, event)
+		svc := *obj.(*v1.Service)
+
+		// Do not handle "kube-system" services
+		if svc.Namespace == meta_v1.NamespaceSystem {
```
kyessenov: Please place a TODO to make this configurable. Eventually, istio will run in kube-system and will be subject to our control plane.

proxy/envoy/watcher.go
```diff
@@ -159,11 +159,13 @@ func (w *watcher) UpdateIngressSecret(ctx context.Context) error {
 		return multierror.Prefix(err, "failed to read request body")
 	}
 	if len(tlsData) == 0 {
+		glog.Errorf("failed to get TLS data, zero data")
 		return nil
 	}
 
 	var tls model.TLSSecret
 	if err = json.Unmarshal(tlsData, &tls); err != nil {
+		glog.Errorf("failed to unmarshal TLS secret")
```
kyessenov: nit: this might be very noisy in the test logs

andraxylia: Indeed, I get plenty of these fails, I will see if I can set it to a verbosity level. I wonder why we get them, do they indicate a real issue?

andraxylia:
Will address the AppNamespaces vs IstioNamespace comments in a subsequent PR.

PR #956, opened 2017-07-25
[WIP] Platform Adapter for VMs

kimikowang:
This is the platform adapter for Istio on VMs. 

Now Amalgam8 registry is used to support the service discovery functionality. But this is only a work-around solution only to prove the concept of this design. Consul integration(work by Greg) is the final target for this adapter. 

The memory config store is running in discovery container to test the performability of routing rules in this adapter. For test purpose, the route rule is pre-hardcoded in the config store. 

Currently demo is available for bookinfo example on docker platform. Instructions to run this demo can be found in branch platform_vms_test.


adapter/config/memory/monitor.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+const ()
+
+type Configs []model.Config
+
+// Handler specifies a function to apply on a Config for a given event type
+type Handler func(model.Config, model.Event)
+
+type Monitor interface {
```
kyessenov: I'm thinking of adding this in-memory controller into master. It seems incomplete as it doesn't properly compare poll result with existing cache. Are you interested in separating this controller out and finishing it, or this is purely for prototyping?

GregHanson:
Work related to PR moved to #1000

PR #966, opened 2017-07-27
k8s client-go update

kyessenov:


kyessenov:
Potentially. It's a strict super-set of 1.6.

adapter/config/tpr/client.go
```diff
@@ -187,7 +188,7 @@ func (cl *Client) RegisterResources() error {
 		for _, kind := range kinds {
 			list := &ConfigList{}
 			err := cl.dynamic.Get().
-				Namespace(api.NamespaceAll).
+				Namespace(v1.NamespaceAll).
```
andraxylia: Can we create an alias for this modules? We cannot change the code every time there is a new version.

kyessenov: k8s is in constant state of flux, so not much we can do about their code migrations.

platform/kube/inject/testdata/hello-config-map-name.yaml.injected
```diff
@@ -61,6 +61,7 @@ spec:
         name: istio-config
       - emptyDir:
           medium: Memory
+          sizeLimit: "0"
```
kyessenov: @ayj any ideas how to deal with this?
I filed a bug https://github.com/kubernetes/client-go/issues/257 but that might take a long time to get fixed.

ayj: Some temporary/hacky solutions that come to mind:

1. custom YAML/JSON marshaler
2. textual removal post-encoding
3. programmatic removal
   * convert typed resource to generic `map[string]interface{}` (https://github.com/fatih/structs maybe?)
   * remove offending key(s)
   * marshal generic map to YAML

WORKSPACE
```diff
@@ -16,11 +16,17 @@ go_repositories()
 
 go_repository(
     name = "com_google_cloud_go",
-    commit = "57377bad3486b37af17b47230a61603794c798ae",
+    commit = "3b1ae45394a234c385be014e9a488f2bb6eef821",
```
yutongz: Do you want this specific SHA, how old is this commit? Maybe try something newer (https://code.googlesource.com/gocloud/) and toolbox/pkg_check will work just fine.

yutongz: Also, you need to update google.golang.org/api (https://code.googlesource.com/google-api-go-client/)

kyessenov:
I borrowed dependencies from k8s. Not sure why they are pinned to an older
version.

adapter/config/tpr/config.go
```diff
@@ -45,6 +46,14 @@ func (e *Config) GetObjectMeta() meta_v1.Object {
 	return &e.Metadata
 }
 
+// DeepCopyObject implements a mandatory interface
+func (e *Config) DeepCopyObject() runtime.Object {
+	if e == nil {
+		return nil
+	}
+	return e
```
odeke-em: IMHO DeepCopyObject for something that just directly returns the reference might be misleading. Perhaps we should call it Reference()?

adapter/config/tpr/config.go
```diff
@@ -55,6 +64,14 @@ func (el *ConfigList) GetListMeta() meta_v1.List {
 	return &el.Metadata
 }
 
+// DeepCopyObject implements a mandatory interface
+func (el *ConfigList) DeepCopyObject() runtime.Object {
+	if el == nil {
+		return nil
+	}
+	return el
```
odeke-em: Ditto

adapter/config/tpr/config.go
kyessenov: This is a required interface from k8s. Implementation is borrowed from apiextensions example.

platform/kube/inject/testdata/auth.cert-dir.yaml.injected
```diff
@@ -64,6 +64,7 @@ spec:
         name: istio-config
       - emptyDir:
           medium: Memory
+          sizeLimit: "0"
```
cmluciano: emptyDir does not seem to like this even though the API ref lists it as an option. I'm trying to determine if the Memory setting cannot be used in union with sizeLimit.

kyessenov: Just filed issue #975 (see details)

kyessenov:
jenkins rebuild pilot/presubmit

kyessenov:
@sebastienvas force merging, only default routing failed (for unclear reasons)

PR #967, opened 2017-07-28
CustomResourceDefinition implementation (nee TPR)

kyessenov:
Copy of TPR implementation for CRD. Only works on k8s 1.7+.

Passed the tests once but CRDs appear to be flaky (e.g. deleting them once seems to botch something on the kube master). (figured this out below, TPRs shadow CRDs)

cc @geeknoid @ayj @rshriram


platform/kube/inject/testdata/multi-init.yaml.injected
```diff
@@ -61,6 +61,7 @@ spec:
         name: istio-config
       - emptyDir:
           medium: Memory
+          sizeLimit: "0"
```
andraxylia: what does this do?

kyessenov:
Flakiness is due to overlap with TPR. Using a different API group fixes the issue.

cmd/istioctl/main.go
```diff
@@ -301,7 +301,7 @@ func init() {
 	}
 	rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "c", defaultKubeconfig,
 		"Kubernetes configuration file")
-	rootCmd.PersistentFlags().StringVarP(&istioSystem, "namespace", "n", api.NamespaceDefault,
+	rootCmd.PersistentFlags().StringVarP(&istioSystem, "namespace", "n", v1.NamespaceDefault,
```
andraxylia: better to keep api, v1 will change, and the default namespace will still be there.

platform/kube/inject/testdata/multi-init.yaml.injected
kyessenov: artifact of k8s 1.7 (backwards incompatible)

cmd/istioctl/main.go
kyessenov: there's no longer api, only v1

andraxylia: you can alias the module name, so we do not keep changing the code every time the version changes.

andraxylia:
Are CRD actually used in this change?

model/conversion.go
```diff
@@ -112,11 +114,15 @@ func ApplyYAML(yml string, pb proto.Message) error {
 
 // FromJSONMap converts from a generic map to a proto message using canonical JSON encoding
 // JSON encoding is specified here: https://developers.google.com/protocol-buffers/docs/proto3#json
-func (ps *ProtoSchema) FromJSONMap(data map[string]interface{}) (proto.Message, error) {
-	// Marshal to json bytes
-	str, err := json.Marshal(data)
+func (ps *ProtoSchema) FromJSONMap(data interface{}) (proto.Message, error) {
+	// Marshal to YAML bytes
```
ayj: why?

kyessenov: why indeed. The new CRD serializer seems to put `map[interface{}]interface{}` fields instead of `map[string]interface{}` which breaks JSON serializer but works with YAML serializer.

ayj: Hmm, good to know -- thanks for the clarification.

PR #970, opened 2017-07-29
Update pre and post submit to skip pkg_check build issue

sebastienvas:


yutongz:
Thanks Seb, will fix it later and turn it on.

Jenkinsfile
```diff
@@ -135,7 +135,7 @@ def postsubmit(gitUtils, bazel, utils) {
       bazel.test('//...')
       sh('bin/init.sh')
       sh('bin/codecov.sh | tee codecov.report')
-      sh('bin/toolbox/presubmit/pkg_coverage.sh')
+      sh('bazel run @com_github_istio_test_infra/toolbox/pkg_check || echo "Failed to run pkg_check"')
```
yutongz: I understand the concern that don't want code coverage to block pr, but do you guys think we should at least expose error and exit at post-submit. Otherwise, no one will notice the error at all.

sebastienvas:
Sorry Yutong I forgot that this was also checking the code coverage. I updated to skip only in case of build issue as this what Kuat issue was.

PR #976, opened 2017-07-31
Enable istioctl build by Jenkins

sebastienvas:
- Update bin/push-docker to take multiple tags
- Re-enable istioctl builds as well as stable artifacts creation for release.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -96,6 +96,10 @@ def stablePresubmit(gitUtils, bazel, utils) {
     bazel.updateBazelRc()
     utils.initTestingCluster()
     sh('ln -s ~/.kube/config platform/kube/')
+    stage('Build istioctl') {
+      def remotePath = gitUtils.artifactsPath('istioctl')
+      sh("bin/upload-istioctl -p ${remotePath}")
+    }
     stage('Integration Tests') {
       timeout(60) {
         sh("bin/e2e.sh -count 10 -logs=false -tag ${env.GIT_SHA}")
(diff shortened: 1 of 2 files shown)


sebastienvas:
@kyessenov For some reason the pilot e2e test is flaky from running in Prow https://k8s-gubernator.appspot.com/build/istio-prow/pull/istio_pilot/976/pilot-presubmit/157/. The only difference with Jenkins is the cluster we run from is using 1.6.4 instead of 1.6.0

kyessenov:
Is there something running the cluster that's intefering?

sebastienvas:
Maybe @nlandolfi. What should we look for  ?

kyessenov:
I can't reproduce it. Seems to work fine on my cluster.

kyessenov:
can this be closed now that jenkins is deprecated in pilot?

sebastienvas:
This is still needed for the release and would rather keep while disabling
jobs in Jenkins for a little while.

nlandolfi:
What needs to be done to move this to prow? Can I help?

kyessenov:
Why do we need a separate job if post-submit can do this?

sebastienvas:
Because this job is called from the release to just rebuild the artifacts. https://github.com/istio/istio/blob/master/release/README.md#rebuild-artifacts-to-include-the-tags

sebastienvas:
@kyessenov, @rvkubiak is waiting for this to run manually through the release.

PR #987, opened 2017-08-03
Allow agent to use a custom config file

costinm:
This is important for testing and customization - the new config is mostly static, agent only needs to watch for cert updates for restarts. 

The config may redirect the logs (/dev/stdout works for docker, but not for VM), may disable the admin,etc.


cmd/pilot-agent/main.go
```diff
@@ -111,9 +115,13 @@ func init() {
 		"Sidecar proxy unique ID. If not provided uses ${POD_NAME}.${POD_NAMESPACE} environment variables")
 	proxyCmd.PersistentFlags().StringVar(&sidecar.Domain, "domain", "cluster.local",
 		"DNS domain suffix")
+	proxyCmd.PersistentFlags().StringVar(&customConfig, "envoyconfig", "",
+		"Custom envoy config")
 
 	cmd.AddFlags(rootCmd)
 
+	rootCmd.Flags().AddGoFlagSet(flag.CommandLine)
```
kyessenov: Use `cmd.AddFlags` function for this.

cmd/pilot-agent/main.go
```diff
@@ -63,8 +66,10 @@ var (
 			glog.V(2).Infof("sidecar %#v", sidecar)
 			glog.V(2).Infof("mesh configuration %#v", mesh)
 
-			if err = os.MkdirAll(configpath, 0700); err != nil {
-				return multierror.Prefix(err, "failed to create directory for proxy configuration")
+			if len(customConfig) == 0 {
+				if err = os.MkdirAll(configpath, 0700); err != nil {
```
kyessenov: configpath is used to store ingress certs irrespective of customConfig

costinm:
We will have a meshconfig - however it's still useful to allow user (and tests) to have his own config.
Otherwise we'll end up with meshconfig duplicating all the options from envoy.json.

proxy/envoy/watcher.go
```diff
@@ -194,18 +193,24 @@ func envoyArgs(fname string, epoch int, mesh *proxyconfig.ProxyMeshConfig, node
...
 			if !ok {
 				return fmt.Errorf("Unexpected config type: %#v", config)
 			}
 
-			// attempt to write file
-			fname := configFile(configpath, epoch)
-			if err := envoyConfig.WriteFile(fname); err != nil {
-				return err
+			var fname string
+			if len(customConfig) > 0 {
```
kyessenov: I had to stop and think for awhile whether this has implication on the certificate revocation workflow. The agent inserts the SHA of certs into the generated config. With a custom config, the SHA cannot be inserted. Seems like it's still not a problem since you trap it just before the binary call. Please document this since it's non-trivial.

cmd/pilot-agent/main.go
```diff
@@ -111,9 +115,13 @@ func init() {
 		"Sidecar proxy unique ID. If not provided uses ${POD_NAME}.${POD_NAMESPACE} environment variables")
 	proxyCmd.PersistentFlags().StringVar(&sidecar.Domain, "domain", "cluster.local",
 		"DNS domain suffix")
+	proxyCmd.PersistentFlags().StringVar(&customConfig, "envoyconfig", "",
+		"Custom envoy config")
```
kyessenov: Document that this effectively disables meshConfig.

cmd/pilot-agent/main.go
costinm: Will update - but that part of the code MUST be fixed, we can't fetch ingress TLS private keys over http with no authentication.

cmd/pilot-agent/main.go
costinm: Not sure what you mean - I want to have all the standard go flags available, no reason to not support them - AddGoFlagSet is on Flags()

kyessenov: We force some values for glog like writing to stderr to prevent writing files in docker containers.

cmd/pilot-agent/main.go
costinm: Done

costinm:
Yes, I saw that - and indeed if user attempts to set that particular flag
explicitly, it'll likely fail.
Not sure if this is the proper way - did you try filling a bug to glog ? Or
maybe we can
use a logger that doesn't require this kind of hacks and is friendly to
docker containers...

I can add a comment with the go flags that won't work due to pilot forcing
- but we should
still accept the flags defined in other packages besides glog.

cmd/pilot-agent/main.go
ayj: _Force_ is probably too strong a word. `cmd.AddFlags` changes the default behavior to write to stderr instead of defaulting to a tmp file. The user should still be able to write to file with `--logtostderr=false`.

costinm:
PTAL

cmd/pilot-agent/main.go
kyessenov: I still don't understand why you need this. Change `cmd.Flags` to fit your needs, please.

costinm: I'm not sure what you mean - AddGoFlagSet means that any go flag defined using standard go will work, as expected. There is no reason to have to add them manually - or disable them.

costinm:
PTAL.

tools/deb/description-discovery
```diff
@@ -0,0 +1 @@
+Istio Pilot server provides services discovery for the http://istio.io project.
```
kyessenov: What's this stuff?

rshriram: debian packaging stuff. need to provide desc of package

cmd/pilot-agent/main.go
```diff
@@ -29,9 +29,10 @@ import (
 )
 
 var (
-	configpath string
-	meshconfig string
-	role       proxy.Node
+	configpath   string
+	meshconfig   string
+	role         proxy.Node
+	customConfig string
```
rshriram: rather than adding another *config string, how about a simple boolean that says "generateConfig" ? You can reuse the configpath variable and populate it with the path to envoy config. Later in code, if generateConfig is true, generate config, else reuse config.

costinm:
Sorry, my mistake - used the wrong branch to merge, need to find how to
revert. Too many workspaces...

costinm:
No longer needed.

PR #993, opened 2017-08-04
Shouldn't have hardcoded 0.1 (or any 2 digit) tags

ldemailly:
0.1.x releases have that dangerous setting of using 0.1 instead of 0.1.x for proxy and init images

fixing this (with the assumption that the correct -tag is passed for test/dev per @sebastienvas )

### DIFF
--- a/cmd/istioctl/inject.go
+++ b/cmd/istioctl/inject.go
@@ -105,9 +105,6 @@ kubectl get deployment -o yaml | istioctl kube-inject -f - | kubectl apply -f -
 			if versionStr == "" {
 				versionStr = version.Line()
 			}
-			if tag == "" {
-				tag = versionStr
-			}
 
 			client, err := kube.CreateInterface(kubeconfig)
 			if err != nil {
@@ -140,7 +137,7 @@ func init() {
 	rootCmd.AddCommand(injectCmd)
 
 	injectCmd.PersistentFlags().StringVar(&hub, "hub", "docker.io/istio", "Docker hub")
-	injectCmd.PersistentFlags().StringVar(&tag, "tag", "", "Docker tag, defaults to same as version")
+	injectCmd.PersistentFlags().StringVar(&tag, "tag", version.Info.Version, "Docker tag")
 
 	injectCmd.PersistentFlags().StringVarP(&inFilename, "filename", "f",
 		"", "Input Kubernetes resource filename")
--- a/cmd/istioctl/inject.go
+++ b/cmd/istioctl/inject.go
@@ -105,6 +105,9 @@ kubectl get deployment -o yaml | istioctl kube-inject -f - | kubectl apply -f -
 			if versionStr == "" {
 				versionStr = version.Line()
 			}
+			if tag == "" {
+				tag = versionStr
+			}
 
 			client, err := kube.CreateInterface(kubeconfig)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/istioctl/inject.go
```diff
@@ -105,6 +105,9 @@ kubectl get deployment -o yaml | istioctl kube-inject -f - | kubectl apply -f -
 			if versionStr == "" {
 				versionStr = version.Line()
 			}
+			if tag == "" {
+				tag = versionStr
```
kyessenov: version.Line() is actually a long line with golang version, SHA, and no release version.

ldemailly: which one is what gets printed (for releases, short version) by "istioctl version" ?

kyessenov: We don't have a buildstamp for the release version 0.2.0. Best that can be done is to use SHA, which is version.Info.GitRevision or version.Info.Version. Take a look at bin/get_workspace_status script.

cmd/istioctl/inject.go
```diff
@@ -137,7 +140,7 @@ func init() {
 	rootCmd.AddCommand(injectCmd)
 
 	injectCmd.PersistentFlags().StringVar(&hub, "hub", "docker.io/istio", "Docker hub")
-	injectCmd.PersistentFlags().StringVar(&tag, "tag", "0.1", "Docker tag")
+	injectCmd.PersistentFlags().StringVar(&tag, "tag", "", "Docker tag, defaults to same as version")
```
ayj: You can use the default version SHA directly, e.g.

```golang
injectCmd.PersistentFlags().StringVar(&tag, "tag", version.Info.GitRevision, "Docker tag, defaults to same as version")
```

Note that this particular SHA method won't always work if developer creates a new GIT commit, rebuilds istioctl, but doesn't push a new  proxy/init images with corresponding SHA.

ldemailly: in that case they'd need to use -tag previoustag
that's ok no ?

ldemailly:
@kyessenov and @ayj thanks for looking at this !
So it seems like, from bin/get-workspace_status
```
if [[ -n "${RELEASE_TAG}" ]]; then
  VERSION="${RELEASE_TAG}"
elif [[ -n ${ISTIO_VERSION} ]]; then
  VERSION="${ISTIO_VERSION}-${BUILD_GIT_REVISION}"
fi
echo buildAppVersion    "${VERSION}"
```
that VERSION would be the right value, assuming we do use that in docker push ?

so using `version.Info.buildAppVersion` is the right thing ?

kyessenov:
I actually don't know who sets `RELEASE_TAG` or `ISTIO_VERSION`. I suggest we default to pilot SHA by default instead of "unknown", and then we can use `version.Info.Version` as default tag.

sebastienvas:
ISTIO_VERSION is set in jenkins https://github.com/istio/pilot/blob/master/Jenkinsfile#L24  and RELEASE_TAG  here https://github.com/istio/pilot/blob/master/bin/get_workspace_status#L45. This reminds me that Prow should probably be setting it.

ldemailly:
@kyessenov 
>  I suggest we use pilot SHA as default instead of "unknown"
done

ldemailly:
before my last update the generated istioctl had:
```
$ ./istioctl-linux version
Version: unknown
GitRevision: df9d35a
GitBranch: test
User: bootstrap@pilot-presubmit-213
GolangVersion: go1.8.3
```
which I didn't quite expect for an automated build (shouldn't it have the branch or sha even without my change) - we'll need to verify it does say 0.2.0 when we make 0.2.0

Issue #989, opened 2017-08-04
Add an option to select docker pull strategy (Always/IfNotPresent) to kube-inject


kyessenov:

saumoh:
Seems like there is a way to do this the k8s way. As a ref:
https://kubernetes.io/docs/admin/admission-controllers/#alwayspullimages

cmluciano:
Should we keep the default as Always?

PR #999, opened 2017-08-07
Add imagePullPolicy flag to istioctl and kube platform

cmluciano:
Signed-off-by: Christopher M. Luciano <cmluciano@us.ibm.com>

Fixes: #989

### DIFF
--- a/cmd/istioctl/inject.go
+++ b/cmd/istioctl/inject.go
@@ -35,6 +35,7 @@ var (
 	versionStr      string // override build version
 	enableCoreDump  bool
 	meshConfig      string
+	imagePullPolicy string
 	includeIPRanges string
 
 	inFilename  string
@@ -126,6 +127,7 @@ kubectl get deployment -o yaml | istioctl kube-inject -f - | kubectl apply -f -
 				EnableCoreDump:    enableCoreDump,
 				Mesh:              mesh,
 				MeshConfigMapName: meshConfig,
+				ImagePullPolicy:   imagePullPolicy,
 				IncludeIPRanges:   includeIPRanges,
 			}
 			return inject.IntoResourceFile(params, reader, writer)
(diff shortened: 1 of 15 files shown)


cmluciano:
Notes for reviewers:

- Does a string value with a default make sense or would you rather have a bool option like --imagePullAlways?
- Does the default of Always make sense given proxy is using IfNotPresent?
- Based on feedback I will update the e2e tests to use IfNotPresent?

platform/kube/inject/inject.go
```diff
@@ -143,13 +144,20 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
 		initArgs = append(initArgs, "-i", p.IncludeIPRanges)
 	}
 
+	var pullPolicy v1.PullPolicy
```
ayj: This should have a reasonable default or return an error for unknown values.

cmd/istioctl/inject.go
```diff
@@ -159,6 +161,8 @@ func init() {
 	injectCmd.PersistentFlags().BoolVar(&enableCoreDump, "coreDump",
 		true, "Enable/Disable core dumps in injected Envoy sidecar (--coreDump=true affects "+
 			"all pods in a node and should only be used the cluster admin)")
+	injectCmd.PersistentFlags().StringVar(&imagePullPolicy, "imagePullPolicy", "Always",
+		"Sets the container image pull policy to Always or PullIfNotPresent")
```
ayj: re: string vs. bool, the description suggests either/or behavior (e.g. bool) but its really a string value which is confusing. Maybe just update the description to include the list of valid options [Always,Never,IfNotPresent]?

platform/kube/inject/inject.go
```diff
@@ -143,13 +144,20 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
...
+	if p.ImagePullPolicy == "Always" {
```
ijsnellf: style nit: a `switch` would be more concise

cmluciano: ack

PR #1000, opened 2017-08-07
Make Consul/VMs option for pilot

kimikowang:
Make Istio work on non-kube environment but general VMs.
Service Discovery is supported by Consul integration.

Design principle is based on description here: https://github.com/istio/pilot/tree/platform_vms_test/bookinfo_istio


cmd/pilot-agent/main.go
```diff
@@ -111,6 +136,14 @@ func init() {
...
 	proxyCmd.PersistentFlags().StringVar(&sidecar.Domain, "domain", "cluster.local",
 		"DNS domain suffix")
+	proxyCmd.PersistentFlags().StringVar(&vmsArgs.config, "vmsconfig", "",
+		"VMs config file for sidecar")
+
+	proxyCmd.PersistentFlags().StringVar(&vmsArgs.serverURL, "serverURL", "",
+		"URL for the registry server")
+
+	proxyCmd.PersistentFlags().StringVar(&vmsArgs.config, "authToken", "",
+		"Authorization token used to access the registry server")
```
rshriram: Why do we need this for consul ? This is for amalgam8 registry right?

cmd/pilot-discovery/main.go
```diff
@@ -36,13 +36,31 @@ import (
 	"istio.io/pilot/tools/version"
 )
 
+// Adapter defines options for underlying platform
+type Adapter string
+
+const (
+	KubernetesAdapter Adapter = "Kubernetes"
+	VMsAdapter        Adapter = "VMs"
```
rshriram: rename to consul

cmd/pilot-discovery/main.go
```diff
@@ -36,13 +36,31 @@ import (
 	"istio.io/pilot/tools/version"
 )
 
+// Adapter defines options for underlying platform
```
rshriram: Rename to serviceregistry

cmd/pilot-discovery/main.go
```diff
@@ -36,13 +36,31 @@ import (
...
+type Adapter string
+
+const (
+	KubernetesAdapter Adapter = "Kubernetes"
```
rshriram: KubernetesRegistry and ConsulRegistry

cmd/pilot-discovery/main.go
```diff
@@ -36,13 +36,31 @@ import (
...
+	VMsAdapter        Adapter = "VMs"
+)
+
+// store the args related to VMs configuration
+type VMsArgs struct {
+	config    string
+	serverURL string
+	authToken string
+}
```
rshriram: What is this for?

cmd/pilot-discovery/main.go
```diff
@@ -149,6 +179,13 @@ func init() {
 	discoveryCmd.PersistentFlags().BoolVar(&flags.discoveryOptions.EnableCaching, "discovery_cache", true,
 		"Enable caching discovery service responses")
 
+	discoveryCmd.PersistentFlags().StringVar(&flags.vmsArgs.config, "vmsconfig", "",
+		"VMs Config file for discovery")
+	discoveryCmd.PersistentFlags().StringVar(&flags.vmsArgs.serverURL, "serverURL", "",
+		"URL for the registry server")
+
+	discoveryCmd.PersistentFlags().StringVar(&flags.vmsArgs.config, "authToken", "",
+		"Authorization token used to access the registry server")
```
rshriram: Not needed

cmd/pilot-agent/main.go
kimikowang: removed

cmd/pilot-agent/BUILD
```diff
@@ -7,6 +7,7 @@ go_library(
     visibility = ["//visibility:private"],
     deps = [
         "//cmd:go_default_library",
+        "//platform/consul:go_default_library",
```
costinm: Why does the agent need a dep to consul ? Shouldn't the discovery handle the integration instead ?
We just managed to remove the dep of agent to k8s, please let's not add another one.

cmd/pilot-agent/main.go
```diff
@@ -47,60 +56,87 @@ var (
...
-			}
-			if sidecar.ID == "" {
-				sidecar.ID = os.Getenv("POD_NAME") + "." + os.Getenv("POD_NAMESPACE")
-			}
-			// receive mesh configuration
-			mesh, err := cmd.ReadMeshConfig(meshconfig)
-			if err != nil {
-				return multierror.Prefix(err, "failed to read mesh configuration.")
-			}
+			if serviceregistry == proxy.KubernetesRegistry {
```
costinm: Why ? The env variables should work without any k8s registry - the name 'pod' and instance can be used without k8s :-)

GregHanson: @costinm please keep in mind that until istio's config server is integrated, the logic in discovery and agent's main.go where an adapter is chosen is completely subject to change.  Ideally, yes, the config pilot retrieves from the config server on start will have fields shared between service registry implementations

cmd/pilot-agent/BUILD
GregHanson: by k8s to you mean a dependency on the istio pilot/platform/kube package or a dependency on the external kubernetes github repo?  Either way, right now agent is just using a utility function in pilot's platform/consul package, so I could move the util function into another common package in pilot

kimikowang:
> Instead of picking a single adapter, why can't we have multiple active at the same time?
I have some confusions on this question.

If you are talking about getting information from multiple adapters at the same time, I think it's needless. Take pilot-agent as an example. It sets platform specific context values(e.g. ip address, domain, ID) for pilot. Thus activating more than one adapter would be useless.

Or do you mean, register one app to multiple adapters? For example, register the service running on k8s to Consul as well?
@rshriram

rshriram:
I meant the pilot discovery service. It can read info from multiple adapters

cmd/pilot-discovery/main.go
```diff
@@ -28,21 +28,32 @@ import (
...
+
 type args struct {
 	kubeconfig string
 	meshconfig string
 
 	// ingress sync mode is set to off by default
 	controllerOptions kube.ControllerOptions
 	discoveryOptions  envoy.DiscoveryServiceOptions
+
+	serviceregistry proxy.ServiceRegistry
```
rshriram: this should be platform.serviceregistry

proxy/context.go
```diff
@@ -45,13 +45,23 @@ type Environment struct {
...
+// ServiceRegistry defines underlying platform supporting service regisgtry
+type ServiceRegistry string
+
+const (
+	// KubernetesRegistry environment flag
+	KubernetesRegistry ServiceRegistry = "Kubernetes"
+	// ConsulRegistry environment flag
+	ConsulRegistry ServiceRegistry = "Consul"
+)
+
```
rshriram: move to platform or model

rshriram:
This looks pretty good. two small nits.. I hope you will add integration tests as well :).

linsun:
sorry for coming late into this, can we describe how the config is surfaced to the user?  and whether if the user can configure pilot to use both consul and k8s or just one of them.

kimikowang:
> This looks pretty good. two small nits.. I hope you will add integration tests as well :).

Sure, tests will be added later.

kimikowang:
> sorry for coming late into this, can we describe how the config is surfaced to the user? and whether if the user can configure pilot to use both consul and k8s or just one of them.

We are now testing k8s api-server running on docker platform. Later crd will be used to expose the APIs and thus user can use `istioctl` to manipulate on the configs.  Now in-memory config store is used for temporary purpose.

Currently user can only use one adapter. Activation of multiple adapters at the same time will be supported in future implementations.

cmd/pilot-discovery/main.go
```diff
@@ -58,79 +70,113 @@ var (
...
+
+				environment.ServiceDiscovery = kubeController
+				environment.ServiceAccounts = kubeController
+				environment.IstioConfigStore = model.MakeIstioStore(configController)
+				environment.SecretRegistry = kube.MakeSecretRegistry(client)
+				serviceController = kubeController
+				ingressSyncer := ingress.NewStatusSyncer(mesh, client, flags.controllerOptions)
+
+				go ingressSyncer.Run(stop)
+			} else if flags.serviceregistry == platform.ConsulRegistry {
```
ldemailly: this should be different functions rather than giant if/else

cmd/pilot-agent/main.go
```diff
@@ -68,6 +61,31 @@ var (
 			glog.V(2).Infof("version %s", version.Line())
 			glog.V(2).Infof("mesh configuration %#v", mesh)
 
+			// set values from registry platform
```
ldemailly: this would also be clearer as a switch and per platform init, specially the line 78-88

rshriram:
@yutongz why did bot merge if tests were failing ?

yutongz:
Which tests were failing? cla check?

GregHanson:
@yutongz 
https://k8s-gubernator.appspot.com/build/istio-prow/pull/istio_pilot/1000/pilot-e2e-smoketest/200/
The according to the job: `Result: Failure` but yet somehow the test also says `No Test Failures!`  and the test output shows unit test failures in mixer

costinm:
Late comment/request: it seems 'agent' is pulling platform/consul package just to get the IP address. Do we really need that ? It won't work well for multi-interface devices - and I'm not sure will work well even for single interface device if it has both IPv4 and IPv6 address.

Would it be possible in a separate PR to remove this and stay consistent with what we do for k8s, i.e. rely on CLI or env to set the desired interface ?

kyessenov:
This PR got merged too fast... Let's fix the test at least before merging.

rshriram:
FYI, it was automated merging. Blame the bots.. I was under the assumption that even if I approve, bot won't merge unless all tests pass. Just didn't take into account the fact that bugs in test itself can result in false positives :(

PR #1009, opened 2017-08-09
Push docker image to docker.io for release

sebastienvas:


sebastienvas:
@rvkubiak When moving to Container Builder we need to deal with authenticating to docker.io

bin/push-docker
```diff
@@ -24,13 +24,24 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-hub="gcr.io/istio-testing"
-tags=$(whoami)_$(date +%y%m%d_%h%m%s)
+
+function docker_push() {
+  im="${1}"
+  if [[ "${im}" =~ 'gcr.io' ]]; then
+    gcloud docker -- push "${im}"
```
kyessenov: this is not necessary. why does gcloud need special docker push?

sebastienvas: because you need gcloud authentication

kyessenov: `gcloud docker --authorize-only` is not enough?

sebastienvas:
ahah We ran out of global external ip, and this is why we failed to get an ingress. I still dont understand why this use an global ip, but nevermind.

sebastienvas:
@rshriram Are you referring to the yaml files in istio/istio ? testing artifacts are still going to point to gcr.io/istio-testing, the official releases (starting with the next one) will point to docker.io. Since 0.2.0 is not real release, I don t think there is a point a pushing those tags to docker.io

bin/push-docker
sebastienvas: I removed this. I think those accomplish the same thing. In any case you have to call gcloud for gcr.io images, so what s the issue ?

kyessenov: One less bash function and one less exception for gcloud :)

PR #1014, opened 2017-08-10
Ensure expected GOPATH setup in the init script

cmluciano:
Signed-off-by: Christopher M. Luciano <cmluciano@us.ibm.com>

### DIFF
--- a/bin/init.sh
+++ b/bin/init.sh
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -ex
 
+# Ensure expected GOPATH setup
+PDIR=`pwd`
+if [ $PDIR != "$GOPATH/src/istio.io/pilot" ]; then
+       echo "Pilot not found in GOPATH/src/istio.io/"
+       exit 1
+fi
+
 # Building and testing with Bazel
 bazel build //...
 


ldemailly:
that'll help indeed, thx

super nit: "Signed-off-by: Christopher M. Luciano..." maybe you want to update your PR template as it looks a bit odd

kyessenov:
That's GPG signature I think.

cmluciano:
Ya I originally just put this in the commit. I think if I only have 1 commit, this becomes the body of the PR.

ldemailly:
doesn't look like gpg to me :-) anyway, if you can use the template with ```release-note ... section that'd be great - or just say more details about the fix in that spot

bin/init.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -ex
 
+# Ensure expected GOPATH setup
+PDIR=`pwd`
+if [ $PDIR != "$GOPATH/src/istio.io/pilot" ]; then
```
ldemailly: I should have caught this but this is incorrect. GOPATH doesn't need to be set (defaults to ~/go https://golang.org/cmd/go/#hdr-GOPATH_environment_variable )

ldemailly: fixing in https://github.com/istio/pilot/pull/1181

PR #1016, opened 2017-08-11
Use customizable log files

kyessenov:


proxy/context.go
```diff
@@ -112,6 +112,8 @@ func DefaultMeshConfig() proxyconfig.ProxyMeshConfig {
 		DiscoveryAddress:   "istio-pilot:8080",
 		EgressProxyAddress: "istio-egress:80",
 
+		AccessLogFile:          "/dev/stdout",
+		ErrorLogFile:           "/dev/stderr",
```
costinm: I don't think we need to redirect stderr - it is handled correctly by systemd, and goes where expected (journal).
Also AFAIK stderr doesn't show up on envoy config, and envoy doesn't use it.

The access logs are the only problem.

proxy/envoy/watcher.go
```diff
@@ -232,10 +233,22 @@ func runEnvoy(mesh *proxyconfig.ProxyMeshConfig, node, configpath string) proxy.
 
 			glog.V(2).Infof("Envoy command: %v", args)
 
+			logFile, errLogFile := os.OpenFile(mesh.ErrorLogFile, os.O_RDWR|os.O_APPEND, 0666)
```
costinm: Well, that makes things worse - systemd will pass a pipe, so "/dev/stderr" can't be opened. Let's not mess stderr.

proxy/envoy/watcher.go
```diff
@@ -232,10 +233,22 @@ func runEnvoy(mesh *proxyconfig.ProxyMeshConfig, node, configpath string) proxy.
...
+			logWriter := bufio.NewWriter(logFile)
+			defer func() {
+				if err := logWriter.Flush(); err != nil {
+					glog.Warning(err)
+				}
+			}()
+
 			/* #nosec */
 			cmd := exec.Command(BinaryPath, args...)
-			cmd.Stdout = os.Stdout
```
costinm: We don't want both going to the same file ?

kyessenov:
not needed

rshriram:
Can we reopen this and make these optional config fields? People who deploy just ingress controller or even otherwise would like to have some standard place to obtain access logs as several monitoring systems scrap these files.

kyessenov:
No longer working

PR #1023, opened 2017-08-12
[wip] local development environment for pilot

kyessenov:
This let you spin up local proxy, agent, discovery, and mixer, and issue requests as if through the client-side proxy.

E.g.:
```
go run test/local/main.go -logtostderr -envoy bazel-genfiles/proxy/envoy/envoy -v 2
curl localhost:8000  --header 'Host: dev.local'
```

This should also work on mac (without mixer) if you use darwin build of Envoy.

### DIFF
--- a/bin/push-docker
+++ b/bin/push-docker
@@ -51,7 +51,7 @@ done
 bin=$(bazel info bazel-bin)
 bazel build --output_groups=static //cmd/... //test/...
 \cp -f  "${bin}/test/client/client.static" docker/client
-\cp -f  "${bin}/test/server/server.static" docker/server
+\cp -f  "${bin}/test/server/main/main.static" docker/server
 \cp -f  "${bin}/cmd/pilot-agent/pilot-agent.static" docker/pilot-agent
 \cp -f  "${bin}/cmd/pilot-discovery/pilot-discovery.static" docker/pilot-discovery
 
--- a/proxy/envoy/discovery.go
+++ b/proxy/envoy/discovery.go
@@ -36,7 +36,8 @@ import (
 // DiscoveryService publishes services, clusters, and routes for all proxies
 type DiscoveryService struct {
 	proxy.Environment
-	server *http.Server
+	server     *http.Server
+	useVirtual bool
 
 	// TODO Profile and optimize cache eviction policy to avoid
 	// flushing the entire cache when any route, service, or endpoint
@@ -194,6 +195,7 @@ type DiscoveryServiceOptions struct {
 	Port            int
 	EnableProfiling bool
 	EnableCaching   bool
+	EnableLocal     bool
 }
 
 // NewDiscoveryService creates an Envoy discovery service on a given port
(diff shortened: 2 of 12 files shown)


kyessenov:
Confirmed that it also works on mac with upstream envoy.

bin/push-docker
```diff
@@ -51,7 +51,7 @@ done
 bin=$(bazel info bazel-bin)
 bazel build --output_groups=static //cmd/... //test/...
 \cp -f  "${bin}/test/client/client.static" docker/client
-\cp -f  "${bin}/test/server/server.static" docker/server
+\cp -f  "${bin}/test/server/main/main.static" docker/server
```
rshriram: has this changed in our repos? or is this your local copy?

kyessenov:
Yeah, it's using non-transparent mode in this setup. It gets more complicated with server-side which is why this is WIP.

bin/push-docker
kyessenov: This change is just in this PR.

proxy/envoy/discovery.go
```diff
@@ -194,13 +195,15 @@ type DiscoveryServiceOptions struct {
...
 	EnableCaching   bool
+	EnableLocal     bool
 }
 
 // NewDiscoveryService creates an Envoy discovery service on a given port
 func NewDiscoveryService(ctl model.Controller, configCache model.ConfigStoreCache,
 	environment proxy.Environment, o DiscoveryServiceOptions) (*DiscoveryService, error) {
 	out := &DiscoveryService{
 		Environment: environment,
+		useVirtual:  !o.EnableLocal,
```
andraxylia: I would rename the useVirtual flag to enableLocal and use the same naming everywhere.

If I understand correctly, this distinguishes the case where iptable redirection is used, so I am not sure if the word "virtual" is the most appropriate.

proxy/envoy/discovery.go
```diff
@@ -219,13 +222,16 @@ func NewDiscoveryService(ctl model.Controller, configCache model.ConfigStoreCach
...
+
+	if ctl != nil {
+		serviceHandler := func(s *model.Service, e model.Event) { out.clearCache() }
+		if err := ctl.AppendServiceHandler(serviceHandler); err != nil {
+			return nil, err
+		}
+		instanceHandler := func(s *model.ServiceInstance, e model.Event) { out.clearCache() }
+		if err := ctl.AppendInstanceHandler(instanceHandler); err != nil {
+			return nil, err
+		}
```
andraxylia: Is ctl == nil an error case? If yes, check it in the beginning and return an error.

If not, is it worth logging it.

test/local/main.go
```diff
@@ -0,0 +1,172 @@
...
+	discoveryPort int
+	configpath    string
+	binarypath    string
+	service       *model.Service
+)
+
+func init() {
+	flag.IntVar(&baseHTTPPort, http, 9000, "Base HTTP/1.1 port")
+	flag.IntVar(&baseGRPCPort, grpc, 9100, "Base gRPC port")
+	flag.IntVar(&n, "n", 10, "Number of instances")
```
andraxylia: Qualify what instances is this about. For example:
"Number of proxy instances" or "Envoy instance" or "app instances"

test/local/main.go
```diff
@@ -0,0 +1,172 @@
...
+func (_ registry) Instances(hostname string, ports []string, tagsList model.TagsList) []*model.ServiceInstance {
+	if hostname != host {
+		return nil
+	}
+
+	out := make([]*model.ServiceInstance, 0)
+
+	for _, port := range ports {
+		var base int
+		if port == http {
```
andraxylia: Use a switch instead of if/else.

linsun:
Very nice.  Does it make sense to have the instruction added to here?  https://github.com/istio/pilot/blob/master/doc/testing.md

kyessenov:
It's not ready and not high priority for me. I'm putting it out since it's useful for debugging.

rshriram:
Closing temporarily to reduce load on bot.

PR #1024, opened 2017-08-14
Enable health checks in kubernetes

### LINKED ISSUE #953: Support k8s custom pod health checks
Change #934 disabled passthrough ports for health checks. We need to re-introduce the support back.

rshriram:
Fixes #953


model/service.go
```diff
@@ -201,6 +201,14 @@ type ServiceDiscovery interface {
 
 	// HostInstances lists service instances for a given set of IPv4 addresses.
 	HostInstances(addrs map[string]bool) []*ServiceInstance
+
+	// ManagementPorts lists set of management ports associated with an IPv4 address.
```
kyessenov: Might be worth explaining what purpose these ports serve in istio context: the proxy should allow connections through this port for transparent proxy injection.

platform/kube/conversion.go
```diff
@@ -18,10 +18,14 @@ import (
 	"fmt"
 	"strings"
```
kyessenov: remove whitespace

platform/kube/conversion.go
```diff
@@ -124,3 +128,104 @@ func convertProtocol(name string, proto v1.Protocol) model.Protocol {
...
+					}
+					// Deduplicate along the way. We don't differentiate between HTTP vs TCP mgmt ports
+					if set[p.Name] == nil {
+						set[p.Name] = p
+					}
+				}
+			}
+		}
+
+		if container.ReadinessProbe != nil {
```
kyessenov: please de-duplicate the code (HTTPGetAction in Probe is same for liveness and readiness)

model/service.go
rshriram: Its already there in the explanation.

platform/kube/conversion_test.go
```diff
@@ -201,3 +202,97 @@ func TestInvalidExternalServiceConversion(t *testing.T) {
...
+
+			podSpec.Containers[0].LivenessProbe.Handler = handler1
+			podSpec.Containers[0].ReadinessProbe.Handler = handler2
+
+			mgmtPorts, err := convertProbesToPorts(podSpec)
+			if err != nil {
+				t.Errorf("Failed to convert Probes to Ports: %v", err)
+			}
+
+			if len(mgmtPorts) != len(expected) {
```
kyessenov: nit: could use reflect.DeepEqual here

proxy/envoy/config.go
```diff
@@ -162,6 +164,12 @@ func buildListeners(env proxy.Environment, sidecar proxy.Sidecar) (Listeners, Cl
 		insertMixerFilter(listeners, instances, sidecar)
 	}
 
+	// Append management listeners after injecting mixer filter to normal listeners
+	// because management port listeners cannot use mixer or auth (see comment
+	// under buildMgmtPortListeners)
+	listeners = append(listeners, mgmtListeners...)
```
kyessenov: The previous iteration would de-duplicate management and inbound listeners. Can we at least skip management listeners that are also inbound listeners? This used to cause segfaults in Envoy before LDS.

proxy/envoy/config.go
```diff
@@ -432,46 +488,13 @@ func buildInboundListeners(instances []*model.ServiceInstance,
...
-
-		case model.ProtocolTCP, model.ProtocolHTTPS:
-			listeners = append(listeners, buildTCPListener(&TCPRouteConfig{
-				Routes: []*TCPRoute{buildTCPRoute(cluster, []string{endpoint.Address})},
-			}, endpoint.Address, endpoint.Port))
 
-		default:
-			glog.Warningf("Unsupported inbound protocol %v for port %#v", protocol, servicePort)
+		listener, cluster := buildInboundConfig(endpoint.Address, endpoint.Port, servicePort.Protocol, mesh)
+		if cluster != nil && listener != nil {
```
kyessenov: refactor to use errors instead of nils

proxy/envoy/config.go
```diff
@@ -481,3 +504,38 @@ func buildInboundListeners(instances []*model.ServiceInstance,
...
+// that the health check ports are distinct from the service ports.
+func buildMgmtPortListeners(managementPorts model.PortList,
+	managementIP string, mesh *proxyconfig.ProxyMeshConfig) (Listeners, Clusters) {
+	listeners := make(Listeners, 0, len(managementPorts))
+	clusters := make(Clusters, 0, len(managementPorts))
+
+	// assumes that inbound connections/requests are sent to the endpoint address
+	for _, mPort := range managementPorts {
+		listener, cluster := buildInboundConfig(managementIP, mPort.Port, mPort.Protocol, mesh)
+		if cluster != nil && listener != nil {
```
kyessenov: same, use errors instead

proxy/envoy/config.go
rshriram: Not sure I understand. the normalize() calls are still happening before returning the final set of listeners and clusters. I thought the normalize was responsible to deduplicate

kyessenov: normalize assumes that the duplicates are identical. otherwise, we get non-determinism which one gets chosen.

rshriram:
I added the health check port to the integration test YAML. If auth was applied, then the e2e test wouldn't pass right?

rshriram:
@kyessenov PTAL. Updated goldens

proxy/envoy/config.go
```diff
@@ -183,13 +156,29 @@ func buildClusters(env proxy.Environment, role proxy.Node) (clusters Clusters) {
...
+	for _, m := range mgmtListeners {
+		l := listeners.GetByAddress(m.Address)
+		if l != nil {
+			glog.Warningf("Omitting listener for management address %s (%s) due to collision with service listener %s (%s)",
+				m.Name, m.Address, l.Name, l.Address)
+			continue
+		}
+		listeners = append(listeners, m)
+	}
+	clusters = append(clusters, mgmtClusters...)
```
kyessenov: we shouldn't be adding clusters for non-existent listeners, they might override clusters for non-management listeners

kyessenov:
Are you sure goldens are right? Doesn't look like they match the code.

PR #1025, opened 2017-08-14
updating to match new debug image

ldemailly:
https://github.com/istio/proxy/commit/2a8426631283669ef1a433c5d1bb84642f042e81


kyessenov:
Actually, you also need to update this: https://github.com/istio/pilot/blob/master/WORKSPACE#L423

PR #1032, opened 2017-08-15
Add controller and eventing support for memory configstore 

kimikowang:
Currently the memory config adapter lacks eventing function.
This PR adds a monitor to work with the controller to allow ADD/UPDATE/DELETE events of configs to be detected and apply handler functions on these events.


adapter/config/memory/monitor.go
```diff
@@ -0,0 +1,95 @@
...
+	handlers := make(map[string][]Handler)
+
+	for _, typ := range store.ConfigDescriptor().Types() {
+		handlers[typ] = make([]Handler, 0)
+	}
+
+	return &configstoreMonitor{
+		store:    store,
+		handlers: handlers,
+		eventCh:  make(chan ConfigEvent),
```
kyessenov: might need some buffer here

kyessenov:
Great work! Let's validate this against the existing config system, and it looks OK to merge.

kimikowang:
> I wrote generic conformance tests for config registry here: https://github.com/istio/pilot/blob/master/adapter/config/crd/controller_test.go. Can you add a test to make sure in-memory config is identical to CRD config?

I added the three tests for the controller.
You may have some questions on the ControllerEvents test. I used the controller as the input value for `store` parameter in the mock.CheckCacheEvents function. In other words, the controller plays roles as both ConfigStore and ConfigStoreCache. Since in memory case, cache is always synced with memory. Do you think this makes sense? @kyessenov

rshriram:
merge with master @kimikowang

adapter/config/memory/monitor_test.go
```diff
@@ -0,0 +1,73 @@
...
+	})
+
+	stop := make(<-chan struct{})
+	go controller.Run(stop)
+
+	// Test Add Event
+	testEvent = model.EventAdd
+	if rev, err := controller.Create(testConfig); err != nil {
+		t.Error(err)
+		return
```
ijsnellf: instead of 
```go
t.Error(err)
return
```
just do
```go
t.Fail(err)
```

adapter/config/memory/monitor_test.go
```diff
@@ -0,0 +1,73 @@
...
+	} else {
+		testConfig.ResourceVersion = rev
+	}
+	<-done
+
+	// Test Update Event
+	testEvent = model.EventUpdate
+	if _, err := controller.Update(testConfig); err != nil {
+		t.Error(err)
+		return
```
ijsnellf: see above

adapter/config/memory/monitor_test.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	<-done
+
+	// Test Delete Event
+	testEvent = model.EventDelete
+	if err := controller.Delete(model.MockConfig.Type, testConfig.Name, TestNamespace); err != nil {
+		t.Error(err)
+		return
```
ijsnellf: see above

PR #1033, opened 2017-08-15
Enable TCP mixer check

kyessenov:
+ small change to forward attributes in ingress

### DIFF
--- a/proxy/envoy/mixer.go
+++ b/proxy/envoy/mixer.go
@@ -110,6 +110,5 @@ func mixerTCPConfig(role proxy.Node) *FilterMixerConfig {
 			AttrTargetIP:  role.IPAddress,
 			AttrTargetUID: "kubernetes://" + role.ID,
 		},
-		QuotaName: MixerRequestCount,
 	}
 }
--- a/proxy/envoy/mixer.go
+++ b/proxy/envoy/mixer.go
@@ -65,7 +65,7 @@ type FilterMixerConfig struct {
 	QuotaName string `json:"quota_name,omitempty"`
 }
 
-func (_ *FilterMixerConfig) isNetworkFilterConfig() {}
+func (*FilterMixerConfig) isNetworkFilterConfig() {}
 
 func buildMixerCluster(mesh *proxyconfig.ProxyMeshConfig) *Cluster {
 	mixerCluster := buildCluster(mesh.MixerAddress, MixerCluster, mesh.ConnectTimeout)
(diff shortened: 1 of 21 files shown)


proxy/envoy/mixer.go
```diff
@@ -103,3 +102,14 @@ func mixerHTTPRouteConfig(role proxy.Node) *FilterMixerConfig {
...
 }
+
+// Mixer TCP filter config for inbound requests
+func mixerTCPConfig(role proxy.Node) *FilterMixerConfig {
+	return &FilterMixerConfig{
+		MixerAttributes: map[string]string{
+			AttrTargetIP:  role.IPAddress,
+			AttrTargetUID: "kubernetes://" + role.ID,
+		},
+		QuotaName: MixerRequestCount,
```
qiwzhang: this will check quota?  do we always want that?

proxy/envoy/testdata/rds-ingress-ssl.json.golden
```diff
@@ -11,7 +11,7 @@
       "cluster": "out.81c187d71467b1608736c57bb0734f9ef9b68f7d",
       "opaque_config": {
        "mixer_control": "on",
-       "mixer_forward": "off"
+       "mixer_forward": "on"
```
qiwzhang: should we have cases where control and forward both are on?  I think we should not.

proxy/envoy/mixer.go
kyessenov: This was forced here due to mixer requirement. Can we drop this?

proxy/envoy/testdata/rds-ingress-ssl.json.golden
kyessenov: For ingress, we actually want both: first we check against the cluster ingress incoming request, then we forward the source attributes set to the ingress service to the backends.
Is there a problem with doing both?

qiwzhang: It may not work.  In the current code,  we set the special HTTP header first,  then extract its attributes and remove it later.   I need to switch the order around.

kyessenov: Filed a bug https://github.com/istio/proxy/issues/411. It should work as-is, it's just istio special attribute is leaking.

proxy/envoy/mixer.go
kyessenov: @douglas-reid does this even matter for TCP?

douglas-reid: I think we decided only to enforce rate limits on number of new connections for TCP (but even that seems like it could come later). I don't think we want `MixerRequestCount` as the quota checked though.

PR #1035, opened 2017-08-15
(Linters) Adding missing --update which should ...

ldemailly:
(Linters) Adding missing --update which should enhance the chance that
everyone sees the same/latest linter errors


ldemailly:
same as https://github.com/istio/mixer/pull/1081

kyessenov:
can we pin these linters to a SHA?

ldemailly:
> can we pin these linters to a SHA?
we should make an image or something like that across all repos but that's out of scope for this change. this change is to address the "doug sees linter errors than noone else sees"

Issue #335, opened 2017-03-15
Transparent proxy injection into existing K8s clusters


andraxylia:
- Initial initializer deployment and configuration is merged into istio/pilot. Additional enhancements are pending review (https://github.com/istio/pilot/pull/1104). 
- The feature is disabled by default in in e2e presubmit tests but can be enabled with `-use-initializer` flag to `bin/e2e.sh`. The feature will by enabled by default in istio/pilot after https://github.com/istio/pilot/pull/1104 is merged. The initializer is cluster scoped so any bugs could affect the entire test cluster. We should be prepared to revert to default-off and manually delete the initializer configuration. ClusterPool testing feature will help here (cc @sebastienvas). The feature can be enabled in istio/istio after istio/pilot switch over is complete.

TODO
- [x] Merge enhancement PR https://github.com/istio/pilot/issues/1104
- [x] Update istio/istio template files for initializer deployment/configuration 
- [x] Document injection method for istio.io https://github.com/istio/pilot/issues/1046

```release-note
Automated sidecar injection
By leveraging the alpha initializer feature provided by Kubernetes 1.7, envoy sidecars can now be automatically injected into application deployments when your cluster has the initializer enabled. This enables you to deploy microservices using kubectl, the exact same command that you normally use for deploying the microservices without Istio.
```

kyessenov:
Please reference the upstream issue from Kubernetes. We might get it for 1.7.

andraxylia:
Depends on kubernetes/community#132

andraxylia:
I opened this issue, hopefully we can enhance PodPreset and re-use functionality:
kubernetes/kubernetes#43874

ldemailly:
doc is https://docs.google.com/document/d/1-5rLhv6MbB27lL8l2CWD6kR-eMa13_PPwOScVVZq8bo/

kyessenov:
There are two aspects of proxy injection: inserting a sidecar and programming iptables. Rephrased the title to be about the first item.

ldemailly:
this is an epic so maybe it should cover both ?

andraxylia:
Yes, it's an epic and covers everything. You can open issues and group them under this epic.

andraxylia:
The title matches the roadmap item, let's leave it as it is.

ldemailly:
epic should not be closed yet

andraxylia:
This is done. P1 enhancements are possible, but they will need to be prioritized.

[...]

PR #1041, opened 2017-08-16
Implement Istio sidecar initializer

ayj:
This PR adds the Kubernetes initializer for transparent proxy injection. The initializer configuration is currently disabled to avoid breaking other test instances without the corresponding initializer deployment. This feature will be incrementally enabled after the initial set of changes are merged and all test instances have a default initializer running with policy=OFF.

Fixes https://github.com/istio/pilot/issues/335


platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	return true
+}
+
+func (i *Initializer) modifyResource(objectMeta *metav1.ObjectMeta, templateObjectMeta *metav1.ObjectMeta, spec *v1.PodSpec) error { // nolint: lll
+	switch i.options.Namespace {
+	case v1.NamespaceAll:
+		// skip kubernetes system namespace
+		if objectMeta.Namespace == "kube-system" {
```
kyessenov: can we define of list of these special namespaces somewhere in `platform/kube`? what about `kube-public`?

platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,456 @@
...
+		// initializing.
+		return nil
+	}
+
+	// Remove self from the list of pending Initializers while
+	// preserving ordering.
+	if pending := objectMeta.GetInitializers().Pending; len(pending) == 1 {
+		objectMeta.Initializers = nil
+	} else {
+		objectMeta.Initializers.Pending = append(pending[:0], pending[1:]...)
```
kyessenov: is `pending[:0]` just a nil? some other trick?

platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,456 @@
...
+
+	glog.Infof("Initializing replicaset: %s", in.Name)
+
+	o, err := runtime.NewScheme().DeepCopy(in)
+	if err != nil {
+		return err
+	}
+	out := o.(*v1beta1.ReplicaSet)
+
+	if err = i.modifyResource(&out.ObjectMeta, &out.Spec.Template.ObjectMeta, &out.Spec.Template.Spec); err != nil {
```
kyessenov: you could try code generation to get around this repetitive code problem (like what I had to do for CRDs)

platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,456 @@
...
+		return err
+	}
+	patchBytes, err := i.createTwoWayMergePatch(in, out, appsv1beta1.Deployment{})
+	if err != nil {
+		return err
+	}
+	_, err = i.clientset.AppsV1beta1().Deployments(in.Namespace).
+		Patch(in.Name, types.StrategicMergePatchType, patchBytes)
```
kyessenov: can you explain the choice of Patch vs Put? Is it to avoid serializing concurrent changes to the config? Is the patch library smart enough to do the right thing here?

platform/kube/inject/inject.go
```diff
@@ -123,16 +170,62 @@ func GetMeshConfig(kube kubernetes.Interface, namespace, name string) (*proxycon
...
+	glog.V(2).Infof("Sidecar injection policy for %v/%v: namespace:%v resource:%v status:%q required:%v",
+		objectMeta.Namespace, objectMeta.Name, namespacePolicy, resourcePolicy, status, required)
+
+	if checkDeprecatedAlphaAnnotation {
+		if _, ok = objectMeta.Annotations[deprecatedIstioSidecarAnnotationSidecarKey]; ok {
+			return false
+		}
+	}
+
+	if !required {
```
kyessenov: this can be moved up

platform/kube/inject/inject.go
```diff
@@ -123,16 +170,62 @@ func GetMeshConfig(kube kubernetes.Interface, namespace, name string) (*proxycon
...
+		return false
+	}
+
+	// TODO - add version check for sidecar upgrade
+
+	return !ok
```
kyessenov: so the idea is that if the status annotation is present then there should be no injection?

docker/Dockerfile.sidecar_initializer
```diff
@@ -0,0 +1,3 @@
+FROM scratch
+ADD sidecar-initializer /usr/local/bin/
+ENTRYPOINT ["/usr/local/bin/sidecar-initializer"]
```
ayj: The current version only ever interacts with the k8s apiserver via patch/list/watch. It uses the in-cluster config similar to pilot discovery service.

cmd/sidecar-initializer/main.go
```diff
@@ -0,0 +1,111 @@
...
+			cmd.WaitSignal(stop)
+
+			return nil
+		},
+	}
+
+	root.PersistentFlags().StringVar(&flags.hub, "hub", "docker.io/istio", "Docker hub")
+	root.PersistentFlags().StringVar(&flags.tag, "tag", "0.1", "Docker tag")
+	root.PersistentFlags().StringVar(&flags.namespace, "namespace",
+		v1.NamespaceDefault, "Namespace managed by initializer")
```
ayj: I went back and forth on the default here ."all" is a reasonable default for cluster-wide installations of Istio. It causes problems for namespace isolated installs. For internal testing we override it anyway so maybe its reasonable for cluster-wide installs to leave this as "all" to avoid broken deployments.

cmd/sidecar-initializer/main.go
```diff
@@ -0,0 +1,111 @@
...
+	root.PersistentFlags().StringVar(&flags.policy, "policy",
+		string(inject.InjectionPolicyOff), "default injection policy")
+
+	root.PersistentFlags().StringVar(&flags.kubeconfig, "kubeconfig", "",
+		"Use a Kubernetes configuration file instead of in-cluster configuration")
+	root.PersistentFlags().StringVar(&flags.meshconfig, "meshConfig", "/etc/istio/config/mesh",
```
ayj: Discovery and Istioctl use "meshConfig".  The agent uses "meshconfig". I'm happy to make this consistent with either - let me know :)

cmd/sidecar-initializer/main.go
```diff
@@ -0,0 +1,111 @@
...
+		fmt.Sprintf("File name for Istio mesh configuration"))
+	root.PersistentFlags().DurationVar(&flags.resyncPeriod, "resync", time.Duration(0),
```
ayj: I don't know. I'll increase to 30 seconds just in case.

docker/Dockerfile.sidecar_initializer
kyessenov: Ok, then it's fine. Running out of cluster would not use docker anyways.

cmd/sidecar-initializer/main.go
kyessenov: Let's use "all". "default" is not really default in the k8s context.

cmd/sidecar-initializer/main.go
kyessenov: let's rename both to meshconfig to be consistent with kubeconfig.

cmd/sidecar-initializer/main.go
kyessenov: for cluster-wide, probably 5 minutes is enough time not to overwhelm the config server.

platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,456 @@
...
+		{
+			"pods",
+			i.clientset.CoreV1().RESTClient(),
+			&v1.Pod{},
+			i.initializePod,
+		},
+	}
+
+	for _, kind := range kinds {
+		kind := kind // capture the current value of `kind`
```
ayj: Yup. Surprisingly the linters don't complain about this (yet).

platform/kube/inject/initializer.go
ayj: This was carried over from keley's initializer tutorial. It's a type'd `nil` which avoids reallocating the underlying array (see https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating)

cmd/sidecar-initializer/main.go
ayj: Changed to `meshconfig` for the initializer command. I can change the others in a follow-up PR.

platform/kube/inject/initializer.go
ayj: Initializers should already be serialized (via ordered pending list). `Patch()` was carried over from Kelsey's initializer tutorial.  He switched from PUT to PATCH [here](https://github.com/kelseyhightower/kubernetes-initializer-tutorial/commit/43c68cee0a08c8ebdcc40697aa0ebf275b64c195#diff-dbaf6084dce27ca6b2f557721fd256b9).

platform/kube/inject/initializer.go
kyessenov: cool

platform/kube/inject/initializer.go
kyessenov: cool. I trust that Kelsey knows what's he doing with PATCH.

platform/kube/inject/inject.go
ayj: Correct

platform/kube/inject/initializer.go
ayj: Good point. I tried a few of different ways to reduce duplication, but wasn't super happy with any of the results. I gave up and focused on getting everything integrated and tested.

platform/kube/inject/initializer.go
rshriram: but do we need this trick? can we just do `for _, k := range kinds` so that code is cleaner and easier for others to follow?

cmd/sidecar-initializer/BUILD
```diff
@@ -0,0 +1,24 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
```
rshriram: missing copyright?

cmd/sidecar-initializer/main.go
```diff
@@ -0,0 +1,111 @@
...
+	}{}
+
+	root := &cobra.Command{
+		Use:   "sidecar-initializer",
+		Short: "Kubernetes initializer for Istio sidecar",
+		RunE: func(*cobra.Command, []string) error {
+			switch inject.InjectionPolicy(flags.policy) {
+			case inject.InjectionPolicyOff, inject.InjectionPolicyOptIn, inject.InjectionPolicyOptOut:
+			default:
+				return fmt.Errorf("unknown namespace injection policy: %v", flags.policy)
```
rshriram: nit. unknown injection policy. (omit the namespace)

cmd/sidecar-initializer/main.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/spf13/cobra"
+)
+
+func getRootCmd() *cobra.Command {
+	flags := struct {
+		kubeconfig   string
+		meshconfig   string
+		hub          string
+		tag          string
+		namespace    string
```
rshriram: Does this enable initializer for a given namespace? If so, won't we need one initializer per namespace? Is there a possibility to watch some config map or something to pick up namespaces enabled for Istio?

platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,389 @@
...
+)
+
+const (
+	initializerName = "sidecar.initializer.istio.io"
+	patchType       = types.StrategicMergePatchType
+)
+
+var ignoredNamespaces = []string{
+	"kube-system",
+	"kube-public",
```
rshriram: probably istio-system as well?

platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// InitializerOptions stores the configurable options for an initializer
+type InitializerOptions struct {
+	// TODO - pull from ConfigMap?
```
rshriram: 👍

platform/kube/inject/initializer.go
rshriram: what is this doing? is there a cleaner way to write this line?

platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,389 @@
...
+		)
+		i.controllers = append(i.controllers, controller)
+	}
+	return i
+}
+
+func (i *Initializer) hasIstioInitializerNext(object metav1.Object) bool {
+	glog.V(2).Infof("ObjectMeta initializer info %v/%v policy:%q status:%q %v",
+		object.GetNamespace(), object.GetName(),
+		object.GetAnnotations()[istioSidecarAnnotationPolicyKey],
```
rshriram: will i be able to skip sidcear using annotations?

platform/kube/inject/inject.go
```diff
@@ -126,16 +173,63 @@ func GetMeshConfig(kube kubernetes.Interface, namespace, name string) (*proxycon
...
+	switch namespacePolicy {
+	case InjectionPolicyOptIn:
+		if resourcePolicy == istioSidecarAnnotationPolicyValueForceOn {
+			required = true
+		}
+	case InjectionPolicyOptOut:
+		if resourcePolicy != istioSidecarAnnotationPolicyValueForceOff {
+			required = true
+		}
+	}
```
rshriram: nice!

platform/kube/inject/testdata/kube-system.yaml.injected
```diff
@@ -0,0 +1,24 @@
...
+metadata:
+  creationTimestamp: null
+  initializers:
+    pending:
+    - name: sidecar.initializer.istio.io
+  name: hello
+  namespace: kube-system
+spec:
+  replicas: 7
+  strategy: {}
```
gyliu513: Why include some empty values here? Can we remove this? Ditto for others.

I also noticed that some old files also include such kind of empty values, is it really useful?

platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,477 @@
...
+		{
+			"statefulsets",
+			i.clientset.AppsV1beta1().RESTClient(),
+			&appsv1beta1.StatefulSet{},
+			i.initializeStatefulSet,
+		},
+		{
+			"jobs",
+			i.clientset.BatchV1().RESTClient(),
+			&batchv1.Job{},
```
gyliu513: How about `CronJob`?

test/integration/testdata/initializer.yaml.tmpl
```diff
@@ -0,0 +1,44 @@
...
+      labels:
+        istio: sidecar-initializer
+      annotations:
+        policy.sidecar.istio.io: "policy.sidecar.istio.io/force-off"
+    spec:
+      serviceAccountName: istio-sidecar-initializer-service-account
+      containers:
+        - name: sidecar-initializer
+          image: {{.Hub}}/sidecar_initializer:{{.Tag}}
+          imagePullPolicy: Always
```
rshriram: +1. I think we fixed this in non initializer case?

platform/kube/inject/testdata/kube-system.yaml.injected
rshriram: I think it's go JSON thingy that generates these empty ones

gyliu513: Can we remove those empty line?

platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,389 @@
...
+	// of watched resources for initialization.
+	ResyncPeriod time.Duration
+
+	// Hub is the image registry used for the injected sidecar proxy
+	// and init images, e.g. docker.io/istio.
+	Hub string
+
+	// Tag is the docker image tag used for the injected sidecar proxy
+	// and init images.
+	Tag string
```
costinm: Would be good to also customize the image names ( or use a full string, as used in docker ). Istio design allows for 
multiple implementations of sidecars.

platform/kube/inject/testdata/daemonset.yaml.injected
```diff
@@ -0,0 +1,87 @@
...
+        track: stable
+    spec:
+      containers:
+      - image: fake.docker.io/google-samples/hello-go-gke:1.0
+        name: hello
+        ports:
+        - containerPort: 80
+          name: http
+        resources: {}
+      - args:
```
costinm: Feature request: allow user to provide a template config - they may want other -v, cpu/mem requirements, etc.
(same for initializer - but probably just the args need to be customized). 

If resource constraints are enabled, it's tricky to schedule jobs if you can't control the injected containers.

platform/kube/inject/initializer.go
ayj: Good idea @costinm. I created https://github.com/istio/pilot/issues/1061 to track that FR.

platform/kube/inject/initializer.go
ayj: Opened https://github.com/istio/pilot/issues/1062 to track this issue.

platform/kube/inject/initializer.go
ayj: Agreed. We went back and forth on these options with kube-inject and settled on hub/tag instead since users had to type it at the command line everyone. This is no longer an issue with initializer but I kept hub/tag for consistency with existing kube-inject. We could fix this as part of https://github.com/istio/pilot/issues/1062.

platform/kube/inject/initializer.go
ayj: Sure. We may need to remove for istio-on-istio depending on how the proxy is deployed.

platform/kube/inject/testdata/daemonset.yaml.injected
ayj: https://github.com/istio/pilot/issues/1062 tracks this issue.

platform/kube/inject/testdata/kube-system.yaml.injected
gyliu513: I see, thanks @ayj

platform/kube/inject/initializer.go
```diff
@@ -0,0 +1,477 @@
...
+
+	// Namespace is the Kubernetes namespace the initializer is
+	// responsible for managing. The initializer will manage injection
+	// for all namespaces if the value of v1.NamespaceAll is
+	// specified.
```
gyliu513: I just mean the word of `specified` can be moved to L65

platform/kube/inject/initializer.go
gyliu513: I see, but shall we file an issue to trace `CronJob` for now?

PR #1042, opened 2017-08-16
Use individual CRDs for config rather than one shared config

kyessenov:


adapter/config/crd/conversion.go
```diff
@@ -16,66 +16,55 @@ package crd
...
-			}
-			return model.Config{
-				Type:     schema.Type,
-				Key:      schema.Key(data),
-				Revision: item.ObjectMeta.ResourceVersion,
-				Content:  data,
-			}, nil
-		}
+// camelCaseToKabobCase converts "my-name" to "MyName"
+func kabobToCamel(s string) string {
```
ayj: `kabobCaseToCamelCase` for symmetry with `camelCaseToKabobCase`?

model/config.go
```diff
@@ -210,20 +215,11 @@ type IstioConfigStore interface {
...
-	IngressRuleProto = "istio.proxy.v1.config.IngressRule"
+	// IstioAPIGroup defines API group name for Istio configuration resources
+	IstioAPIGroup = "config.istio.io"
 
-	// DestinationPolicy defines the type for the destination policy configuration
-	DestinationPolicy = "destination-policy"
-	// DestinationPolicyProto message name
-	DestinationPolicyProto = "istio.proxy.v1.config.DestinationPolicy"
+	// IstioAPIVersion defines API group version
+	IstioAPIVersion = "v1alpha1"
```
ayj: Mixer and Pilot config are part of the API group. Do they also share the same API version, or are they independent?

adapter/config/crd/generate.sh
```diff
@@ -0,0 +1,78 @@
...
+EOF
+
+done
+
+cat <<EOF
+}
+EOF
+
+for crd in $CRDS; do
+  sed -e "1,20d;s/IstioKind/$crd/g" adapter/config/crd/config.go
```
ayj: A comment explaining what this does would be helpful. Does this modify `adapter/config/crd/config.go` or is the output going into types.go?

model/config.go
kyessenov: I remember seeing bug reports about using two versions for same CRD group. I changed it to v1alpha2 with the goal of coordinating the change with @mandarjog

adapter/config/crd/generate.sh
kyessenov: Adding a comment

adapter/config/crd/conversion.go
kyessenov: done

PR #1049, opened 2017-08-17
Introduce config metadata

kyessenov:
Step 2 of config update: define config metadata and config wrapper object.

Step 1 was switching to individual kinds for each config type.
Step 3 is to update config proto definitions to utilize the metadata.


kyessenov:
validated with https://github.com/istio/istio/pull/571

adapter/config/crd/controller.go
```diff
@@ -47,7 +46,8 @@ type cacheHandler struct {
 }
 
 // NewController creates a new Kubernetes controller for CRDs
-func NewController(client *Client, resyncPeriod time.Duration) model.ConfigStoreCache {
+// Use "" for namespace to listen for all namespace changes
```
ayj: nit: maybe add platform neutral equivalent to `v1.NamespaceAll`?

kyessenov:
ping. PR in istio broke all PRs in pilot until this gets merged.

adapter/config/crd/controller.go
```diff
@@ -233,18 +228,19 @@ func (c *controller) List(typ string) ([]model.Config, error) {
...
-					Key:      schema.Key(data),
-					Revision: item.GetObjectMeta().ResourceVersion,
-					Content:  data,
-				})
-			}
+		if !ok {
+			continue
+		}
+
+		if namespace != "" && namespace != item.GetObjectMeta().Namespace {
```
ayj: nit: s/""/v1.NamespaceAll

adapter/config/ingress/controller.go
```diff
@@ -145,69 +138,74 @@ func (c *controller) ConfigDescriptor() model.ConfigDescriptor {
...
-			ingressRules := convertIngress(*ingress, c.domainSuffix)
-			for key, rule := range ingressRules {
-				out = append(out, model.Config{
-					Type:     model.IngressRule.Type,
-					Key:      key,
-					Revision: ingress.GetResourceVersion(),
-					Content:  rule,
-				})
-			}
+		if namespace != "" && namespace != ingress.Namespace {
```
ayj: nit: s/""/v1.NamespaceAll

adapter/config/ingress/conversion.go
```diff
@@ -106,28 +106,37 @@ func createIngressRule(name, host, path, namespace, domainSuffix string,
...
+	}
 }
 
 // encodeIngressRuleName encodes an ingress rule name for a given ingress resource name,
 // as well as the position of the rule and path specified within it, counting from 1.
 // ruleNum == pathNum == 0 indicates the default backend specified for an ingress.
-func encodeIngressRuleName(ingressName, ingressNamespace string, ruleNum, pathNum int) string {
-	return fmt.Sprintf("%s.%s.%d.%d", ingressNamespace, ingressName, ruleNum, pathNum)
+func encodeIngressRuleName(ingressName string, ruleNum, pathNum int) string {
+	return fmt.Sprintf("%s-%d-%d", ingressName, ruleNum, pathNum)
```
ayj: Why the switch from `.` to `-` as a separator?

model/config.go
```diff
@@ -25,19 +27,32 @@ import (
...
-// key identifier that is unique per type, and the content represented as a
-// protobuf message.  The revision is optional, and if provided, identifies the
+// ConfigMeta is metadata attached to each configuration unit.
+// The revision is optional, and if provided, identifies the
 // last update operation on the object.
-type Config struct {
+type ConfigMeta struct {
 	// Type is a short configuration name that matches the content message type
-	Type string
+	// (e.g. "route-rule")
```
ayj: Are type names normalized to non-k8s specific names? e.g. route-rule instead of route-rule.config.istio.io?

adapter/config/ingress/conversion.go
kyessenov: k8s allows only - in names.

model/config.go
kyessenov: Yes, type is `route-rule`, API is `route-rules.config.istio.io/v1alpha2`.

adapter/config/ingress/controller.go
kyessenov: Ok

kyessenov:
let me merge this to unblock the others. will address namespace.All in a subsequent PR.

model/config.go
rshriram: can we move to v1beta1 ? claim beta release with alpha api version is odd.

model/config.go
```diff
@@ -335,55 +350,64 @@ func (i *istioConfigStore) RouteRulesBySource(instances []*ServiceInstance) []*p
...
 			out = append(out, rule)
 		}
 	}
 	return out
 }
 
 func (i *istioConfigStore) DestinationPolicy(destination string, tags Tags) *proxyconfig.DestinationVersionPolicy {
-	value, exists, _ := i.Get(DestinationPolicy.Type, destination)
+	names := strings.Split(destination, ".")
+	name, namespace := "", ""
```
rshriram: what if I have foo.lyft.com as my destination and its in istio global namespace?

andraxylia:
Why did  you force merge this when pre-submit does not pass?

kyessenov:
It passed on istio/istio against the SHA of this PR. This PR required coordinating changes across two repositories which is iffy each time.

model/config.go
kyessenov: we don't have a global namespace, CRDs are namespaces just like services themselves.

kyessenov: This is provisioning the "name" and "namespace" use from CRD like we discussed. It was already happening before under the hoods, this is just making it explicit.

model/config.go
kyessenov: CRDs may only support one version at a time and mixer already started using v1alpha2

rshriram: I chatted with @geeknoid. Let's please change everywhere to v1beta1

frankbu: Changed to v1beta1 in config design doc: https://docs.google.com/document/d/1fGZpgFWJZhNRlQoBlCW815aOFR-ewfxKhla0CcPRsBM/edit#

kyessenov: If we use v1beta1 then by k8s conventions, this API will provide future backwards compatibility, @geeknoid. As long as we're ok taking on the work to maintain it forever, i'm OK changing to v1beta1.

PR #1052, opened 2017-08-17
resources: clarify acronyms

vbatts:
sometimes it's nice to get the full name out there, with all these xDS

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>


vbatts:
I could do this in other places like the `RDS` struct, etc.

PR #1056, opened 2017-08-17
Init OWNERS and pr template

yutongz:
Add files for mungegithub. 

Here is a refer: https://github.com/istio/mixer/pull/1024

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,8 @@
+reviewers:
+  - ayj
+  - kyessenov
+  - rshriram
+  - sebastienvas
+approvers:
+  - kyessenov
+  - rshriram
--- a/OWNERS
+++ b/OWNERS
@@ -1,4 +1,5 @@
 reviewers:
+  - andraxylia
   - ayj
   - GregHanson
   - ijsnellf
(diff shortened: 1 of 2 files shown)


OWNERS
```diff
@@ -0,0 +1,8 @@
+reviewers:
```
ldemailly: list seems a bit too short / bottleneck-y

yutongz: I agree, and I think the best part of this OWNERS system is that you can specify people in a specific area, which means we can put more people in the package he/she is mainly working on. The root OWNERS might be more suitable for people taking charge of the whole repo like Kuat and rshriram.

yutongz:
@rshriram sure thing

OWNERS
rshriram: In that case, put @GregHanson for consul, @ijsnellf for eureka, @ayj for kube, in addition to me and Kuat. I am probably missing more folks but these should be a good coverage for starters and we can add people as time progresses.

yutongz: Sounds good.

ldemailly: I'd add andra too

yutongz:
Can anyone help me merge it and I will deploy mungegithub right way.

Issue #1029, opened 2017-08-14
debuggability : add --debug option to istioctl kube-inject (with default on for 0.2)


ldemailly:
Right now we do inject debug images but

a) they ship with readOnlyRoot fs so not much can be done with them
b) they ship without privileges - so for instance you can't run `iptables`
c) they have too much unnecessary stuff that could be installed only if needed by the person debugging / needing them (using `sudo proxy-redirection-clear && sudo apt-get install favortitepkg && sudo proxy-redirection-restore` if a & b are fixed)

the flag would turn on
```
diff --git a/platform/kube/inject/inject.go b/platform/kube/inject/inject.go
index b8d5da6..2d5ed1c 100644
--- a/platform/kube/inject/inject.go
+++ b/platform/kube/inject/inject.go
@@ -263,7 +263,7 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
 		})
 	}
 
-	readOnly := true
+	readOnly := false
 	sidecar := v1.Container{
 		Name:  ProxyContainerName,
 		Image: p.ProxyImage,
@@ -294,6 +294,8 @@ func injectIntoPodTemplateSpec(p *Params, t *v1.PodTemplateSpec) error {
 		SecurityContext: &v1.SecurityContext{
 			RunAsUser:              &p.SidecarProxyUID,
 			ReadOnlyRootFilesystem: &readOnly,
+			// TODO: Determine SELINUX options needed to remove privileged
+			Privileged: &privTrue,
 		},
 		VolumeMounts: volumeMounts,
 	}
```

also with --debug=false the non debug image would be injected 
(and we should test both)

cc @andraxylia

ldemailly:
going to do this now

ldemailly:
Fixed

PR #1057, opened 2017-08-17
Change coredump pattern to writeable directory

kyessenov:


ldemailly:
I appreciate the quick fix but I think it's wrong - we shouldn't use "/etc/istio/" to put cores

the solution should be something like https://github.com/istio/pilot/issues/1029

andraxylia:
We need core file regardless of whether debug is enabled or not. 

I tend to agree with /etc not being the right place to write cores. What is the reason for changing the directory location from /tmp to /etc?

kyessenov:
/etc/istio/proxy is a tmpfs mounted by k8s. It's a rather painful process to change installation files at the moment. this addresses the immediate need to debug segfaults.

ldemailly:
the reason (explained in #1029) is that the fs is currently hardcoded to readonly

I am not sure I agree that optimized builds should generate core dumps in prod (given that fully optimized builds won't have useful stack traces for instance)

but I do strongly disagree with making core dumps in /etc

if we want a quick hack just change readOnly := true to readOnly := false
there is no data demonstrating that has downside

in other words I don't think you should merge this

kyessenov:
This has nothing to do with raw VM setup. We can mount envoy configs into /tmp if that helps. But we should not be writing files inside the docker container.

ldemailly:
> This has nothing to do with raw VM setup.

?? where does it say raw vm anywhere here or in #1029 ?

kyessenov:
I was implying that the choice of directory is not as important as on a real machine. It's all VFS.

kyessenov:
We can change the location in another PR.

andraxylia:
This has nothing to do with raw VM.

Can we try with a persistent volume? It needs to be available even if the pod/container dies.

PR #1063, opened 2017-08-18
Enable config for release artifacts

sebastienvas:


bin/init.sh
```diff
@@ -9,7 +19,7 @@ if [ $PDIR != "$GOPATH/src/istio.io/pilot" ]; then
 fi
 
 # Building and testing with Bazel
-bazel build //...
+bazel build ${bazel_build_args[@]} //...
```
kyessenov: If you really need to customize bazel build, then I suggest remove this line from this script.

bin/push-docker
```diff
@@ -37,19 +37,21 @@ function docker_push() {
 hubs="gcr.io/istio-testing"
 local_tag=$(whoami)_$(date +%y%m%d_%h%m%s)
 tags="${local_tag}"
+bazel_build_args=(--output_groups=static)
 
 while [[ $# -gt 0 ]]; do
     case "$1" in
         -tag) tags="$2"; shift ;;
         -hub) hubs="$2"; shift ;;
+        -config) bazel_build_args+=(--config="$2"); shift ;;
```
kyessenov: can we call it bazelconfig or something more specific than "config"?

bin/upload-istioctl
```diff
@@ -7,19 +7,22 @@
 set -o errexit
 set -o nounset
 set -o pipefail
+set -ex
```
kyessenov: why? that's same as options above

tools/bazel.rc
```diff
@@ -2,5 +2,8 @@
 build --verbose_failures
 test --test_output=errors
 
+# Release Build
+build:release -c opt
```
kyessenov: does -c opt even have any effect on golang?

bin/upload-istioctl
```diff
@@ -7,19 +7,22 @@
...
+INIT_ARGS=()
 
 function cleanup {
   rm -rf "${BIN_DIR}"
 }
 
-while getopts :p:r arg; do
+while getopts :c:p:r arg; do
   case ${arg} in
+    c) INIT_ARGS+=(-config ${OPTARG});;
```
kyessenov: can we avoid mysterious short flags and use same flag as above?

tools/bazel.rc
sebastienvas: From @ldemailly it does !

bin/init.sh
sebastienvas: Removing and building where ?

bin/upload-istioctl
sebastienvas: I ll remove it, this was just for debugging to print the commands.

tools/bazel.rc
kyessenov: what changes in the go binary?

ldemailly: the 2x benefits I measured is with envoy (c++) but I expect debug to be -gcflags="-N -l" and opt to be without those and thus make a difference, not sure the bazel go rules do do that (we should check, but even if they don't today, having 2 or more config will be useful in the future)

kyessenov: yes, envoy is special. let's first check if there's any difference with rules_go. if not, then we can't even use bazel to accomplish optimized build.

ldemailly: I asked in https://github.com/bazelbuild/rules_go/issues/741 but we should use this config seb is added anyway, to change the behavior of kube-inject / to select the optimized sidecar images for instance

sebastienvas: I wish I had known that before starting.

bin/push-docker
sebastienvas: Done

bin/upload-istioctl
sebastienvas: Done

bin/upload-istioctl
sebastienvas: Done

bin/init.sh
sebastienvas: I am not sure what you meant here.

sebastienvas:
Closing this as this is does not have any impact on the artifacts.

ldemailly:
it will in next bazel: see https://github.com/bazelbuild/rules_go/pull/743

kyessenov:
keep the PR open until rules_go releases the support.

ldemailly:
or we could just get it in and it'll get benefits automatically once bazel side changes - whichever works for you

sebastienvas:
closing this as I just force pushed to this branch and lost the code for this :(

ldemailly:
the bazel change is in, can we at least keep this for early part of 0.3

sebastienvas:
we can but I lost the code, so I have to it all over again.

ldemailly:
code is here no ? https://patch-diff.githubusercontent.com/raw/istio/pilot/pull/1063.patch

ldemailly:
or use the reflog to recover the code ?
(but we can close this if you prefer - and just leave the issue)

sebastienvas:
The code that you pointed out refer to another PR. Tell me more about
reflog.

ldemailly:
you can do "git reflog" and see your local history and go back to any point in that history
using git reset using the id on the left

sebastienvas:
it's gone!

Issue #1072, opened 2017-08-21
`make lint` failed due to linters not installed for pilot


gyliu513:
We should enable `./bin/install-prereqs.sh` when `make setup` for pilot.

Checkout here https://groups.google.com/forum/#!topic/istio-dev/jlG0Y9PMgFc for detail.

PR #1075, opened 2017-08-21
Enabled setup install pre-requested go binaries.

gyliu513:
**Release note**:

```release-note
None
```
Fixed https://github.com/istio/pilot/issues/1072


gyliu513:
@ldemailly I think the order does not impact much, as the two shell scripts do not have dependency, comments?

PR #1084, opened 2017-08-21
Fix a typo

vadimeisenbergibm:
**Release note**:

```release-note
None
```


vadimeisenbergibm:
@kyessenov removing the type causes the build to fail. It seems the typo prevented the line to execute, once executed, it causes the build to fail. Closing the PR.

vadimeisenbergibm:
@kyessenov my bad, it is not a typo, but part of the sed command, sorry for confusion.

PR #1087, opened 2017-08-21
Use absolute name of PWD to match Bazel's behavior

rvkubiak:
Fixes breakage of upload-istioctl when the build directory is a symlink.

**Release note**:
```release-note
Fix failure on cross-compiled istioctl builds when path contains a symlink
```


yutongz:
@kyessenov @sebastienvas  It seems both you two are ok with this pr, so if you want merge-bot taking care of this, just say "/lgtm" and "/approve"

PR #1092, opened 2017-08-22
build should depend on kubeconfig.

gyliu513:
`make build` will be failed if do not run `make setup`. This PR enabled `make setup` first before `make build`.

**Release note**:

```release-note
None
```

/cc @rshriram

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,8 @@ fmt:
 	@bin/fmt.sh
 
 .PHONY: build
-build:	
+build: kubeconfig
+	@bin/install-prereqs.sh
 	@bazel build //...
 
 .PHONY: clean
@@ -52,5 +53,5 @@ racetest:
 	@bazel test --features=race //...
 
 kubeconfig:
-	@ln -s ~/.kube/config platform/kube/
+	@bin/kubeconfig.sh
 
--- a/bin/kubeconfig.sh
+++ b/bin/kubeconfig.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+if [ ! -f platform/kube/config ]; then
+  ln -s ~/.kube/config platform/kube/
+fi


kyessenov:
Hm, `bin/init.sh` depends on `bazel build //...` since it symlinks fetched depedencies to vendor. This works on prow, since prow caches all dependencies. Why is the reverse dependency needed?

rshriram:
@gyliu513 bin/init.sh already does a bazel build.. This is going to end up doing it twice.

gyliu513:
Thanks @rshriram and @kyessenov , updated the PR a bit, I think the `build` still depend on `kubeconfig` and `bin/install-prereqs.sh`.

Makefile
```diff
@@ -24,7 +24,8 @@ fmt:
 	@bin/fmt.sh
 
 .PHONY: build
-build:	
+build: kubeconfig
+	@bin/install-prereqs.sh
```
kyessenov: install-prereqs is using `go get` which doesn't cache well. This means the build will be slowed down by quite a bit.
How about this: take a look at prow presubmit scripts. You should see a setup section there. Let's create `setup` task in `make` that does similar set-upy things (you may even split out those scripts out of prow-presubmit). Then leave build be just the bazel build.

gyliu513: @kyessenov The presubmit is also installing prereq first and then bazel build ter/prow/pilot-presubmit.sh#L45-L47 , also in Makefile, we already have the `setup` task, but the `setup` also include `init.sh` which will invoke bazel build.

The reason that I want to add `kubeconfig` to here is because if developer do not run `make setup`, then there maybe some go tool chain lost and cause bazel build failed, comments?

kyessenov: I see. So there are the following steps in the order of dependency:
1. add kubeconfig (for testing)
2. bazel build/test (to fetch dependencies)
3. bin/init.sh (to vendorize dependencies)
4. install prereqs (gometalinter mostly)
5. bin/check.sh for linting (to run linters on vendorized go-built code)

As long as you don't add any back dependency, we're good. Feel free to choose where to make the cut or how to cache steps.

gyliu513: Perhaps updating the document may be a good choice, what do you think @kyessenov

kyessenov: I'm out for a few days, but yeah, will get to updating docs once back.

rshriram:
temporarily closing due to overload.. please reopen if needed @gyliu513 (sorry for the inconvenience).

Issue #1006, opened 2017-08-08
Expose websocket routes


rshriram:
Envoy has support for WebSockets now (https://github.com/lyft/envoy/pull/1274). However, routes that support WebSockets must be explicitly marked using the `use_websocket` option. 

This has two implications:
* For a HTTP port (say 80) on a service, that has several routes (plain http and websocket) (/foo, /bar and /websocket), the user needs to mark websocket enabled routes explicitly through route rules.
* For services that have websocket on a dedicated port (i.e., no http routes), the user could name the port as a websocket port (like http, grpc, etc.). This would mean that all routes on that port are websocket routes (preventing normal http connections from working.

WE could start with option 1 and then see if it makes sense to introduce a new port name/type (Websocket) as I feel that option 2 is redundant.
cc @kyessenov

kyessenov:
1) makes more sense for ingress while 2) is more in-line how we use services in-cluster. I think we might need both options, but for different proxies.

rshriram:
(1) also makes sense for in-cluster proxies, if there are multiple services on same port. Or, if a given service has HTTP routes and one or more websocket routes. (2) is more of a coarse grained solution.

rshriram:
fixed by #1094

PR #1094, opened 2017-08-22
Configure Websocket routes

rshriram:
Configures websocket routes in Envoy for routes marked for websocket upgrade

#1006 

**Release note**:

```release-note
Configure websocket routes in Envoy
```


proxy/envoy/config.go
```diff
@@ -503,6 +503,19 @@ func buildInboundListeners(mesh *proxyconfig.ProxyMeshConfig, sidecar proxy.Node
 				Routes:  []*HTTPRoute{route},
 			}
 
+			// Websocket enabled routes need to have an explicit use_websocket : true
+			// This setting needs to be enabled on Envoys at both sender and receiver end
+			if protocol == model.ProtocolHTTP {
+				// get all the route rules applicable to the instances
+				rules := config.RouteRulesByDestination(instances)
```
kyessenov: Hm, the target generally doesn't know where the request comes from, so can't truthfully infer the rules that carried the request to destination unless it's in the request metadata.

rshriram: That is fine.. All we need is to know the URIs that support websockets and setup special route entries in envoy with websocket enabled. We don't have to do source based routing or any form of matching. This is just a roundabout way to collect websocket enabled URLs. I was hoping that we would put in some caching in that config object such that we dont have to compute this again and again.

test/client/client.go
```diff
@@ -105,12 +108,59 @@ func makeHTTPRequest(client *http.Client) func(int) func() error {
...
+				log.Printf("[%d] Body=%s\n", i, msg)
+			}
+
+			conn, _, err := client.Dial(url, req)
+			if err != nil {
+				// timeout or bad handshake
+				return err
+			}
+			// nolint: errcheck
+			defer conn.Close()
```
ijsnellf: does `// nolint: errcheck` have to be on the same line?

proxy/envoy/config.go
```diff
@@ -518,19 +518,41 @@ func buildInboundListeners(mesh *proxyconfig.ProxyMeshConfig, sidecar proxy.Node
 		// services' kubeproxy to our specific endpoint IP.
 		switch protocol {
 		case model.ProtocolHTTP, model.ProtocolHTTP2, model.ProtocolGRPC:
-			route := buildDefaultRoute(cluster)
+			defaultroute := buildDefaultRoute(cluster)
```
ijsnellf: nit: golang style `defaultroute` -> `defaultRoute`

test/client/client.go
rshriram: err on.. ths worked..

test/grpcecho/echo.proto
```diff
@@ -16,14 +16,14 @@ syntax = "proto3";
 
 package grpecho;
 
-service EchoTestService {
-  rpc Echo(EchoRequest) returns (EchoResponse);
+service GrpcEchoTestService {
```
ldemailly: @rshriram why is this changed ?
is that what causes
https://k8s-gubernator.appspot.com/build/istio-prow/pull/istio_pilot/1102/pilot-presubmit/574/

ldemailly: filled #1112 as I don't understand how it seems to have passed in this pr and not passing anymore

ldemailly: @rshriram ping - why the change ?

PR #1096, opened 2017-08-22
Add release script to build Pilot artifacts

rvkubiak:
Add config to run Pilot release on GCB

**Release note**:
```release-note
Add release.sh script that runs tasks that currently exist as part of the Jenkins config.
Add a config file for running Pilot release builds on Google Container Builder.
```

### DIFF
--- a/bin/release.sh
+++ b/bin/release.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -x
+
+function usage() {
+  echo "$0 \
+    -i <gcloud project ID name> \
+    -t <tag name to apply to artifacts>"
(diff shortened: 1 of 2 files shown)


bin/release.sh
```diff
@@ -0,0 +1,42 @@
...
+}
+
+# Initialize variables
+PROJECT_ID=""
+TAG_NAME=""
+
+# Handle command line args
+while getopts i:t: arg ; do
+  case "${arg}" in
+    i) PROJECT_ID="${OPTARG}";;
```
sebastienvas: I don't think that passing the PROJECT_ID makes sense.

For docker we need to push to gcr.io/istio-io and docker.io/istio, for istioctl let's use istio-release bucket.

PR #1102, opened 2017-08-24
Change secure naming annotation signature.

myidpt:
```release-note
Change secure naming annotation signature.
```


platform/kube/conversion.go
```diff
@@ -35,11 +35,11 @@ const (
 
 	// KubeServiceAccountsOnVMAnnotation is to specify the K8s service accounts that are allowed to run
 	// this service on the VMs
-	KubeServiceAccountsOnVMAnnotation = "alpha.istio.io/serviceaccounts/kubernetes"
+	KubeServiceAccountsOnVMAnnotation = "alpha.istio.io/kubernetes-serviceaccounts"
 
 	// CanonicalServiceAccountsOnVMAnnotation is to specify the non-Kubernetes service accounts that
 	// are allowed to run this service on the VMs
-	CanonicalServiceAccountsOnVMAnnotation = "alpha.istio.io/serviceaccounts/canonical"
+	CanonicalServiceAccountsOnVMAnnotation = "alpha.istio.io/canonical-serviceaccounts"
```
ldemailly: can you elaborate on what makes this "canonical" ?

platform/kube/conversion.go
```diff
@@ -35,11 +35,11 @@ const (
 
 	// KubeServiceAccountsOnVMAnnotation is to specify the K8s service accounts that are allowed to run
 	// this service on the VMs
-	KubeServiceAccountsOnVMAnnotation = "alpha.istio.io/serviceaccounts/kubernetes"
+	KubeServiceAccountsOnVMAnnotation = "alpha.istio.io/kubernetes-serviceaccounts"
```
ldemailly: should be serviceaccount ? no s ? (there is only 1 value)

myidpt: Actually there can be multiple service accounts on VM to run the same service, which is the same as in K8s :)

platform/kube/conversion.go
myidpt: "Canonical" means the service account is directly used as the Istio identity - for example, for GCP service account "foo@gserviceaccount.com", the Istio identity is "spiffe://foo@gserviceaccount.com".
This is different from K8s service account, which Istio applies a specific scheme for its Istio identity:
"spiffe://domain/ns/\<namespace\>/sa/\<serviceaccount\>".

ldemailly: ah thx, so canonical means without k8s prefixes (but with spiffe://)

maybe I should add that option too then in istioctl register

platform/kube/conversion.go
ldemailly: what is the syntax for multiple accounts and what does it mean to have multiple accounts ?

myidpt: For example, you can have one VM running by foo@gserviceaccount.com, and one VM running by bar@gserviceaccount.com. Both VMs are the endpoints for service "datastore". In this case, the service needs to specify both service accounts in the service's annotation.

platform/kube/conversion.go
costinm: My understanding was that spiffe syntax is spiffee://domain/something - not spiffe://email ?

myidpt: Actually we use spiffe://email scheme for two reasons here.
1. This email is globally unique, user may want to directly use it as the SPIFFE ID.
2. spiffee://domain/foo@gserviceaccount.com is not a valid URI.
In the long term, however, in order to support the name constraints on CA, I think we may need to have a scheme to translate Email IDs to SPIFFE IDs with domain names (something like spiffe://domain/gserviceaccount.com/foo). This will be implemented when we have a complete native identity to mesh identity mapping feature in auth control plane, probably in V0.3. So in short, this is a temporary solution for us.

Issue #1103, opened 2017-08-25
Add HTTP endpoint to sidecar initializer


ayj:
Add an HTTP endpoint equivalent to the sidecar initializer. This provides additional choices in how operators chose to inject the sidecar into their deployments, e.g. integration with CI/CD systems where injected resources are version controlled.

Issue #1062, opened 2017-08-18
Use ConfigMap for sidecar initializer deployment configuration


ayj:
The sidecar initializer is configured through command line flags. It might be nice to pull the configuration and sidecar template from a k8s ConfigMap (e.g. [envoy-initializer.yaml](https://github.com/kelseyhightower/kubernetes-initializer-tutorial/blob/master/configmaps/envoy-initializer.yaml)). 

NOTE: The sidecar initializer and kube-inject currently share the same injection programmatic template. Adoption of [envoy-initializer.yaml](https://github.com/kelseyhightower/kubernetes-initializer-tutorial/blob/master/configmaps/envoy-initializer.yaml) style template should ensure these templates don't diverge. kube-inject could use the newer k8s ConfigMap but care should be taken to avoid breaking offline mode, e.g. kube-inject might need to take ConfigMap YAML as argument to retrieve template.

gyliu513:
I think that the injected configuration should be handled by pilot but not the configMap as the end user may update the configMap and caused the initializer does not work. @ayj

ayj:
I'm not sure how pushing injection configuration into Pilot would solve that problem. The end user could make the same misconfiguration errors there. Ultimately the injection configuration should be protected and I believe that is possible today with appropriate RBAC settings on the ConfigMap.

Implemented by https://github.com/istio/pilot/pull/1104

PR #1104, opened 2017-08-25
Additional sidecar injection enhancements

ayj:

* Add an http endpoint for sidecar injection. This provides additional
  choices in how operators chose to inject the sidecar into their
  deployments.

  Support for kube-inject using this API will be added in a follow-up
  PR. Here is an example using curl and the k8s apiserver proxy.

```bash
kubectl proxy &
export PATH=localhost:8001/api/v1/namespaces/istio-system/services/istio-sidecar-initializer:8083/proxy/inject
curl -X POST ${PATH} --data-binary "@deployment.yaml" localhost:9090/inject > deployment-injected.yaml
```
* Use a ConfigMap for the sidecar template parameters.

* Refactor the injection logic and use k8s scheme and runtime support
  along with reflection to generalize the per-resource injection
  path. Adding new supported kinds is now a one line addition to the
  list of supported kinds, e.g.

```golang
{v1beta1.SchemeGroupVersion, &v1beta1.DaemonSet{}, "daemonsets", "/apis"},
{v1beta1.SchemeGroupVersion, &v1beta1.Deployment{}, "deployments", "/apis"},
{appsv1beta1.SchemeGroupVersion, &appsv1beta1.StatefulSet{}, "statefulsets", "/apis"}
```

fixes https://github.com/istio/pilot/issues/1103, https://github.com/istio/pilot/issues/1062

The Initializer configuration is disabled until the feature is stable and all test instances cluster-wide are running the initializer controller. You can test this feature locally with the the e2e test using the `-use-initializer` flag.

```bash
bin/e2e.sh -use-initializer
``` 

The e2e templated YAML files can be used as an additional reference until this feature is integrated in istio/istio.

* Initialzier configuration - [test/integration/testdata/initializer-configmap.yaml.tmpl](https://github.com/ayj/pilot/blob/f681997188faaac468fb2e037368c65acdc9698b/test/integration/testdata/initializer-configmap.yaml.tmpl)
* Initializer deployment - [initializer.yaml.tmpl](https://github.com/ayj/pilot/blob/f681997188faaac468fb2e037368c65acdc9698b/test/integration/testdata/initializer.yaml.tmpl)

**Release note**:

```release-note
* Add an http endpoint for sidecar injection. Users can POST k8s resource files to
  the '/inject' endpoint. The response body contains the file with the Istio sidecar
  injected into supported resource similar to kube-inject. This provides additional 
  choices in how operators chose to inject the sidecar into their deployments.

* Use a ConfigMap for the sidecar template parameters.

* Generalize the per-resource injection. It should be easier to add new supported resource kinds in the future.
```


costinm:
Just a note: please make sure that any new use of k8s API have a clearly documented RBAC policy/requirements.  If the http endpoint is intended for automated systems / jobs - 
we can't grant them admin and leave admin credentials on random machines, so I assume
a service account needs to be created and granted some permissions ?

IMHO a CLI for injection - without deps to a live k8s and without credential requirements - 
would work better for CI/CD. AFAIK we only use k8s to load resource maps / mesh config -
which could be checked in.

ayj:
The HTTP endpoint returns the modified resource back to the user. It doesn't submit anything to the k8s apiserver therefore no new service account or RBAC privileges should be needed. The user is still responsible for creating/updating the resource.

gyliu513:
@ayj this is an important feature, can we add some release note?

proxy/BUILD
```diff
@@ -25,6 +25,7 @@ go_test(
 
 go_test(
     name = "go_default_xtest",
+    size = "small",
```
gyliu513: what is the size mean here?

platform/kube/inject/testdata/required.yaml.patch
```diff
@@ -0,0 +1 @@
...{"annotations":{"alpha.istio.io/sidecar":"injected(deprecated)","status.sidecar.istio.io":"injected-version-12345678"}},"spec":{"$setElementOrder/containers":[{"name":"hello"},{"name":"istio-proxy"}],"containers":[{"args":["proxy","-v","2"],"env":[{"name":"POD_NAME","valueFrom":{"fieldRef":{"fieldPath":"metadata.name"}}},{"name":"POD_NAMESPACE","valueFrom":{"fieldRef":{"fieldPath":"metadata.namespace"}}},{"name":"INSTANCE_IP","valueFrom":{"fieldRef":{"fieldPath":"status.podIP"}}}],"image":"docker.io/istio/proxy_debug:unittest","imagePullPolicy":"IfNotPresent","name":"istio-proxy","resources":{},"securityContext":{"readOnlyRootFilesystem":true,"runAsUser":1337},"volumeMounts":[{"mountPath":"/etc/istio/config","name":"istio-config","readOnly":true},{"mountPath":"/etc/istio/proxy","name":"istio-envoy"}]}],"initContainers":[{"args":["-p","15001","-u","1337"],"image":"docker.io/istio/proxy_init:unittest","imagePullPolicy":"IfNotPresent","name":"istio-init","resources":{},"securityContext":{"capabilities":{"add":["CAP_NET_ADMIN"]},"privileged":true}}],"volumes":[{"configMap":{"name":"istio"},"name":"istio-config"},{"emptyDir":{"medium":"Memory","sizeLimit":"0"},"name":"istio-envoy"}]}}}}
```
gyliu513: Can you adjust the format to JSON?

platform/kube/inject/testdata/first-initializer.yaml.patch
```diff
@@ -0,0 +1 @@
...{"annotations":{"alpha.istio.io/sidecar":"injected(deprecated)","status.sidecar.istio.io":"injected-version-12345678"}},"spec":{"$setElementOrder/containers":[{"name":"hello"},{"name":"istio-proxy"}],"containers":[{"args":["proxy","-v","2"],"env":[{"name":"POD_NAME","valueFrom":{"fieldRef":{"fieldPath":"metadata.name"}}},{"name":"POD_NAMESPACE","valueFrom":{"fieldRef":{"fieldPath":"metadata.namespace"}}},{"name":"INSTANCE_IP","valueFrom":{"fieldRef":{"fieldPath":"status.podIP"}}}],"image":"docker.io/istio/proxy_debug:unittest","imagePullPolicy":"IfNotPresent","name":"istio-proxy","resources":{},"securityContext":{"readOnlyRootFilesystem":true,"runAsUser":1337},"volumeMounts":[{"mountPath":"/etc/istio/config","name":"istio-config","readOnly":true},{"mountPath":"/etc/istio/proxy","name":"istio-envoy"}]}],"initContainers":[{"args":["-p","15001","-u","1337"],"image":"docker.io/istio/proxy_init:unittest","imagePullPolicy":"IfNotPresent","name":"istio-init","resources":{},"securityContext":{"capabilities":{"add":["CAP_NET_ADMIN"]},"privileged":true}}],"volumes":[{"configMap":{"name":"istio"},"name":"istio-config"},{"emptyDir":{"medium":"Memory","sizeLimit":"0"},"name":"istio-envoy"}]}}}}
```
gyliu513: Can you adjust the format to JSON?

platform/kube/inject/testdata/first-initializer.yaml
```diff
@@ -0,0 +1,19 @@
...
+    pending:
+    - name: sidecar.initializer.istio.io
+    - name: some.other.initializer
+spec:
+  replicas: 7
+  template:
+    spec:
+      containers:
+        - name: hello
+          image: "fake.docker.io/google-samples/hello-go-gke:1.0"
```
gyliu513: How about adding imagePullPolicy here?

platform/kube/inject/initializer.go
```diff
@@ -88,242 +62,202 @@ type InitializerOptions struct {
...
+		groupVersion schema.GroupVersion
+		obj          runtime.Object
+		resource     string
+		apiPath      string
+	}{
+		{v1.SchemeGroupVersion, &v1.ReplicationController{}, "replicationcontrollers", "/api"},
+
+		{v1beta1.SchemeGroupVersion, &v1beta1.Deployment{}, "deployments", "/apis"},
+		{v1beta1.SchemeGroupVersion, &v1beta1.DaemonSet{}, "daemonsets", "/apis"},
+		{v1beta1.SchemeGroupVersion, &v1beta1.ReplicaSet{}, "replicasets", "/apis"},
```
gyliu513: You added a new type `replicasets` here, but the file https://github.com/istio/pilot/blob/master/test/integration/testdata/initializer-config.yaml.tmpl was not updated

platform/kube/inject/initializer.go
```diff
@@ -88,242 +62,202 @@ type InitializerOptions struct {
...
 
-	// Remove self from the list of pending Initializers while
-	// preserving ordering.
-	if pending := objectMeta.GetInitializers().Pending; len(pending) == 1 {
-		objectMeta.Initializers = nil
-	} else {
-		objectMeta.Initializers.Pending = append(pending[:0], pending[1:]...)
-	}
+	glog.V(2).Infof("ObjectMeta initializer info %v/%v policy:%q status:%q %v",
+		obj.GetNamespace(), obj.GetName(),
```
gyliu513: How about adding resource kind here? Such as daemonset, statefulset etc?

gyliu513:
Another comment is you may want to run `make fmt` to adjust the go format.

gyliu513:
@ayj can you please append more detail for how to test your patch? It would be great if you can provide a deployment file for the initializer like https://github.com/kelseyhightower/kubernetes-initializer-tutorial/blob/master/deployments/envoy-initializer.yaml and people can do some test, thanks!

platform/kube/inject/initializer.go
ayj: The list here is only what the sidecar-initializer supports. The resource types in initializer-config.yaml.tmpl are what is actually configured. I removed the non-deployment types in the e2e test to make this more clear.

platform/kube/inject/testdata/first-initializer.yaml
ayj: I don't think it matters too much since this file is never posted to k8s apiserver.

proxy/BUILD
ayj: https://docs.bazel.build/versions/master/test-encyclopedia.html#role-of-the-test-runner

ayj:
@gyliu513, I updated the reviewer notes to include instructions on how to test and added some release notes - PTAL

costinm:
Doesn't read CRDs need permissions (RBAC, auth) as well, or are CRDs
readable by anyone ? Would be very worried if that's the case, can you
double check?

We already have this problem with pilot - where we bypass k8s security
policies by exposing without permissions resources that
would normally require auth/acl, need to be very careful.

gyliu513:
Thanks @ayj 

I was using the command to run the `e2e test` but failed as following, can you help?

```
root@pc002:~/go/src/istio.io/pilot# bin/e2e.sh -hub docker.io/gyliu -use-initializer
....
INFO: Running command line: bazel-bin/test/integration/integration --logtostderr -use-initializer -hub docker.io/gyliu -tag root_170828_Aug081503906453
F0828 03:55:16.842336   29424 driver.go:115] the server has asked for the client to provide credentials (post customresourcedefinitions.apiextensions.k8s.io)
ERROR: Non-zero return code '255' from command: Process exited with status 255
```

Also can you please show more detail steps for how to enable me test with following configurations? There is not too much document for the test so hope can get some help from you.

* Initialzier configuration - [test/integration/testdata/initializer-configmap.yaml.tmpl](https://github.com/ayj/pilot/blob/f681997188faaac468fb2e037368c65acdc9698b/test/integration/testdata/initializer-configmap.yaml.tmpl)
* Initializer deployment - [initializer.yaml.tmpl](https://github.com/ayj/pilot/blob/f681997188faaac468fb2e037368c65acdc9698b/test/integration/testdata/initializer.yaml.tmpl)

ayj:
The initializer doesn't use CRD -- maybe you meant ConfigMap for the sidecar template? Either way, reads to either can be protected. The HTTP endpoint can be disabled with `--port=0` if there are concerns of exposing the sidecar template.

cmd/sidecar-initializer/main.go
```diff
@@ -35,68 +32,60 @@ func getRootCmd() *cobra.Command {
...
-	rootCmd.PersistentFlags().StringVar(&flags.policy, "policy",
-		string(inject.InjectionPolicyOff), "default injection policy")
-
 	rootCmd.PersistentFlags().StringVar(&flags.kubeconfig, "kubeconfig", "",
 		"Use a Kubernetes configuration file instead of in-cluster configuration")
 	rootCmd.PersistentFlags().StringVar(&flags.meshconfig, "meshconfig", "/etc/istio/config/mesh",
-		fmt.Sprintf("File name for Istio mesh configuration"))
-	rootCmd.PersistentFlags().DurationVar(&flags.resyncPeriod, "resync", 6*time.Minute,
-		"Initializers resync interval")
+		"File name for Istio mesh configuration")
```
rshriram: out of curiosity, why do we need to specify mesh config in sidecar initializer?

platform/kube/inject/inject.go
```diff
@@ -90,8 +92,11 @@ const (
 // Defaults values for injecting istio proxy into kubernetes
 // resources.
 const (
-	DefaultSidecarProxyUID = int64(1337)
-	DefaultVerbosity       = 2
+	DefaultSidecarProxyUID   = int64(1337)
+	DefaultVerbosity         = 2
```
rshriram: Doesn't this translate to tons of logs from pilot?

platform/kube/inject/inject.go
```diff
@@ -408,72 +503,29 @@ func IntoResourceFile(p *Params, in io.Reader, out io.Writer) error {
...
-					return injectWithPolicyCheck(&o.ObjectMeta, &o.Spec.Template.ObjectMeta, &o.Spec.Template.Spec)
-				},
-			},
-			"StatefulSet": {
-				typ: &appsv1beta1.StatefulSet{},
-				inject: func(typ interface{}) error {
-					o := typ.(*appsv1beta1.StatefulSet)
-					return injectWithPolicyCheck(&o.ObjectMeta, &o.Spec.Template.ObjectMeta, &o.Spec.Template.Spec)
-				},
-			},
```
rshriram: dumb question.. where are these being moved to?

cmd/sidecar-initializer/main.go
ayj: `ProxyListenPort`, `AuthPolicy`, and `AuthCertsPath`. They're used in the sidecar template. We could dup these between inject and proxy-mesh ConfigMaps but that seemed more error prone at this stage than continuing to use use the Proxymesh directly.

platform/kube/inject/inject.go
ayj: Yes, but this is no different than the kube-inject method. If desired, it would be good to change the default behavior in a follow-up PR.

platform/kube/inject/inject.go
ayj: This code was generalized (see [lines 512-513](https://github.com/istio/pilot/pull/1104/files/dbdcebb82793fc91408bdba443d3eaf48a53d593#diff-3f71ca8a1c3836bc4a3360d83741f2fbR512) and [intoObject](https://github.com/istio/pilot/pull/1104/files#diff-3f71ca8a1c3836bc4a3360d83741f2fbR447)). This uses k8s `schema` and `runtime` support to manipulate k8s objects and the `reflect` package to extract the `ObjectMeta` and `PodTemplateSpec` sub-structures.

The table of supported GroupVersionKind's and corresponding objects is [here](https://github.com/istio/pilot/pull/1104/files#diff-bbaf00e98adfa4174f7d712a3869dff1R64).

ayj:
> Also can you please show more detail steps for how to enable me test with following configurations? There is not too much document for the test so hope can get some help from you.

The `-use-initializer` flag should be sufficient for pilot e2e presubmit tests. If you want to try this feature elsewhere you'll need the following:

0) Copy RBAC rules from rbac-beta.yaml.tmpl if RBAC is enabled
1) create the Initializer ConfigMap with sidecar template (see initializer-configmap.yaml.tmpl)
2) create the Initializer deployment (see initializer.yaml.tmpl)
3) create the InitializerConfiguration (see initializer-config.yaml.tmpl)

Do not enable #3 in a shared (test) cluster. All namespaces will be subject to initialization. This will block resource creation for all namespaces not managed by the initializer deployment (see namespace field in the ConfigMap).

cmd/sidecar-initializer/main.go
kyessenov: I prefer that we use a single source of truth for global settings, so this is sort of necessary. AuthPolicy will eventually become a destination policy. I also like that istioctl doesn't have to fetch it itself, but rather have this service do conversion.

platform/kube/inject/initializer.go
```diff
@@ -14,72 +14,40 @@
...
 
 const (
 	initializerName = "sidecar.initializer.istio.io"
-	patchType       = types.StrategicMergePatchType
 )
 
 var ignoredNamespaces = []string{
-	"kube-system",
-	"kube-public",
+	"kube-system", // v1.NamespaceSystem
```
kyessenov: maybe do the replacement?

test/integration/infra.go
```diff
@@ -112,17 +112,32 @@ func (infra *infra) setup() error {
...
+			Version:           "integration-test",
+			Mesh:              mesh,
+			MeshConfigMapName: "istio",
+		},
+	}
+
+	// NOTE: InitializerConfiguration is cluster-scoped and may be
+	// created and used by other tests in the same test cluster.
+	if infra.UseInitializer {
 		if err := deploy("initializer-config.yaml.tmpl", infra.IstioNamespace); err != nil {
```
kyessenov: can we serialize inject.Config to a template or post it with k8s API? is that valuable?

platform/kube/inject/initializer.go
ayj: Yup. `NamespaceSystem` wasn't defined in `v1` along with `NamespaceDefault` so I thought it was a dep issue. Turns out some of these also appear in `metav1`.

test/integration/infra.go
ayj: The sidecar parameters (inject.Config) are available in the template [here](https://github.com/istio/pilot/pull/1104/files/#diff-2d12068fb87af5b21a9140990090b41a). Or did you mean something else?

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: istio-inject
data:
  config: |-
    policy: {{.Policy}}
    namespace: {{.Namespace}}
    params:
      initImage: {{.Params.InitImage}}
      proxyImage: {{.Params.ProxyImage}}
      verbosity: {{.Params.Verbosity}}
      SidecarProxyUID: {{.Params.SidecarProxyUID}}
      Version: {{.Params.Version}}
      EnableCoreDump: {{.Params.EnableCoreDump}}
      MeshConfigMapName: {{.Params.MeshConfigMapName}}
      ImagePullPolicy: {{.Params.ImagePullPolicy}}
      IncludeIPRanges: {{.Params.IncludeIPRanges}}
```

gyliu513:
Hope we can get this merged today @ayj @rshriram @kyessenov

gyliu513:
I also filed an issue in kubernetes proposing to move `initializers` feature to Beta https://github.com/kubernetes/kubernetes/issues/51443

PR #1105, opened 2017-08-25
Integration test improvements for Minikube

ijsnellf:
Several changes I end up adding to the integration tests when I am doing testing in Minikube. If possible, I'd like to merge them.


test/integration/tcp.go
```diff
@@ -44,6 +44,10 @@ func (t *tcp) run() error {
 	funcs := make(map[string]func() status)
 	for _, src := range testPods {
 		for _, dst := range testPods {
+			if src == dst {
+				continue // minikube bug prevents this scenario from working
```
rshriram: May be qualify a bit more? Add a -minikube flag to bazel.

test/util/kubernetes.go
```diff
@@ -81,7 +81,7 @@ func GetAppPods(cl kubernetes.Interface, ns string) (map[string][]string, error)
 			return pods, err
 		}
 		items = list.Items
-		ready := true
+		ready := len(items) > 0
```
rshriram: This should not be necessary since we now check if the containers are running before declaring the pod ready

andraxylia: I also have changes in the same fct in PR, where I check for pods/containers in multiple namespaces. You can revert this.

test/integration/tcp.go
andraxylia: Only the t->t scenario causes issues in minikube, I have this in my PR as well.

ijsnellf:
@andraxylia: since most of these changes are already in your PR, please either close this one, or lgtm it if you want the changes to be in their own PR.

PR #1106, opened 2017-08-25
Rollback breaking changes to target.service

costinm:

Revert recent change that removed 'target.service',it breaks raw VM and consul 
support for mixer. 

Code based on original before the breakage, with small adjustments for the new 
layout.

**Release note**:

```release-note
NONE
```


proxy/envoy/config.go
```diff
@@ -209,6 +210,7 @@ func buildSidecar(env proxy.Environment, sidecar proxy.Node) (Listeners, Cluster
 		clusters = append(clusters,
 			httpOutbound.clusters()...)
 		listeners = append(listeners,
+			buildHTTPListener(env.Mesh, sidecar, instances,nil, LocalhostAddress, int(env.Mesh.ProxyHttpPort), RDSAll, false))
```
ldemailly: missing , ? (doesn't compile)

also run "make fmt"

kyessenov:
This will break mixer AFAIK. Why is this not coordinated with me or mixer folks to begin with?

costinm:
Re. rename: this PR is strictly a rollback, it is bad practice to mix roll backs with backward incompatible API changes. 

Given the date - and impact on testing ( already have few engineers wasting hours due to URL format changes that made pilot incompatible with older envoy labes) - I think a safe rename is needed, not simply replacing the name (i.e. send both along, make sure there is no disruption on testing, etc).

ldemailly:
why close the pr?

bin/check.sh
```diff
@@ -19,7 +19,7 @@ gometalinter --concurrency=${NUM_CPU} --enable-gc --deadline=300s --disable-all\
   --enable=gotype\
   --enable=ineffassign\
   --enable=interfacer\
-  --enable=lll --line-length=120\
+  --enable=lll --line-length=160\
```
ldemailly: where does this stop ?

costinm:
Kuat, can you unblock ? Responded to the comments.

bin/check.sh
costinm: It stops when we have a consistent Istio project - there is no reason for mixer and istio to have a different code style from pilot. 

I would note that google go style, and normal go style doesn't have limits on line length - not clear why Istio needs to be different, or how such decision is made - but 160 wins.

A bigger problem is the ineffassign - which seems to prevent 
port := 0 - for no good reason. I think we need some process around such decisions.

kyessenov: Well, I'm a vim user and I like 120 characters. 160 is too long. Don't bundle these subtle checks into PRs, it's easy to miss and will surprise us later.

proxy/envoy/config.go
```diff
@@ -260,8 +261,24 @@ func buildHTTPListener(mesh *proxyconfig.ProxyMeshConfig, role proxy.Node, route
...
+		for _, instance := range instances {
+			serviceSet[instance.Service.Hostname] = true
+		}
+		services := make([]string, 0, len(serviceSet))
+		for service := range serviceSet {
+			services = append(services, service)
+		}
+
+		sort.Strings(services)
+		service = strings.Join(services, ",")
```
kyessenov: @douglas-reid Costin really needs to set target service in pilot. Can we coordinate on the choice of the target service among the list of instances? Is it first by alphabetic order?

bin/check.sh
rshriram: +1

ldemailly: re ineffassign, it's for a good reason it is not doing anything... it should be
```
	var port int
```
or better deleted and := in each block

what's the problem with that ?

proxy/envoy/config.go
```diff
@@ -260,8 +261,24 @@ func buildHTTPListener(mesh *proxyconfig.ProxyMeshConfig, role proxy.Node, route
 		Config: FilterRouterConfig{},
 	})
 
+	service := ""
```
rshriram: qualify: mixerSourceService
Also, add a comment please, so that in future, when we axe this code, we know what to replace it with - and don't cause confusion again.

proxy/envoy/mixer.go
```diff
@@ -101,6 +104,11 @@ func mixerHTTPRouteConfig(role proxy.Node) *FilterMixerConfig {
 		},
 		QuotaName: MixerRequestCount,
 	}
+	if len(service) > 0 {
+		r.MixerAttributes[AttrTargetService] = service
+	}
+
+	return r
```
rshriram: why can't this go into the MixerAttributes struct itself? I believe that struct has been declared with `omit_empty` thingy..

proxy/envoy/config.go
douglas-reid: After our offline discussion, I'm not entirely sure what the action here is. My understanding was that, for `target.service`, Mixer was going to get out of the business of trying to figure that out. If there are multiple services per pod, are we going to use an annotation to select the canonical Istio service (maybe "istio=" ?)?

rshriram: We essentially need to pass a ISTIO_SERVICE_NAME env var to pilot agent (vm or k8s).. and if this gets passed to envoy (via service cluster option), then we can infer the target service easily. This is a generic solution that is independent of the platform. the agent can default to istio-proxy if not present..

proxy/envoy/config.go
costinm: Agree, added the comment - this is a revert, so the original name and code as we had it 2 weeks ago was used.

The only change I made was to make it compile.

bin/check.sh
costinm: Reverted to 120 until we can have a discussion on style guides for Istio (instead of each repo doing its own random stuff).

costinm: Re. ineffassign: original code (port:=0) fails the linter, it has to be changed to "var port int". 
The problem is that port:=0 is a perfectly common pattern, and shouldn't be rejected.

proxy/envoy/mixer.go
costinm: I'm not entirely sure if setting a value to "" will cause it to be sent as empty value, or how omit_empty may interact.
This seems more readable and explicit.

bin/check.sh
ldemailly: bikeshedding here but having a variable defined at the minimum scope required is also a good thing (in this case  port wasn't really needed at that scope - eg proper fix should be https://github.com/istio/pilot/pull/1121/files)

one benefits of linter is there is no discussion/argument: just do what the machine tells you :-)

proxy/envoy/config.go
kyessenov: Except the late bound service problem... Let's not bring this one up again.

platform/kube/conversion.go
```diff
@@ -169,7 +169,7 @@ func convertProbePort(c v1.Container, handler *v1.Handler) (*model.Port, error)
 
 	var protocol model.Protocol
 	var portVal intstr.IntOrString
-	port := 0
+	var port int
```
ldemailly: if you rebase someone already fix that (though not the 'best' way)

bin/check.sh
kyessenov: Still see 160

proxy/envoy/config.go
costinm: For the general case - you can have different listeners, each with a different target.service. 

There are many ideas and suggestions on how to improve this - the point of the PR is to revert to the status
before it got broken, it will not be better than what we had 2 weeks ago when it broke.

IMHO we need a proper design and discussion before making this kind of changes, there are many 
possible solutions and tradeoffs.

PR #1110, opened 2017-08-27
Update code coverage processes

yutongz:
1. Fix pkg_check build failed because of deps conflicts
2. Update coverage.requirement with latest SHA and match new format (with default requirement)
3. Separate coverage processes (codecov.io, package check) with presubmit:
    1) Speed up presubmit
    2) Coverage will not block presubmit, we can easily make it required or not.

Connect to https://github.com/istio/test-infra/pull/422

**Release note**:

```release-note
none
```


prow/pilot-presubmit.sh
```diff
@@ -44,10 +44,6 @@ fi
 
 echo '=== Bazel Build ==='
 ./bin/install-prereqs.sh
-bazel build //...
```
yutongz: Removing "bazel build //..." because bin/init.sh already includes it.

yutongz:
I think it's ok to merge this in. As well as the pr in test-infra.

PR #1120, opened 2017-08-28
Cleanup Consul integration code

### LINKED ISSUE #1118: Eliminate dependency from proxy on platform
Re: https://github.com/istio/pilot/commit/c54e6ec0a5cab3bce8b67af05b67d464422bcf65
We require that proxy does not depend directly on the platform. All interactions should go through the model abstraction, to decoupling the behavior from the environment. 
I'm not sure why the PR got merged as-is, please re-work it to avoid bundling platform with the proxy.

GregHanson:
Follow up to premature merge of PR #1000 

Resolves #1118 

**Release note**:

```release-note
```


rshriram:
Need more explanation please. From what I gathered, zijun's PR was non
functional. What exactly is done in this PR?


-- 

~shriram

proxy/envoy/net.go
```diff
@@ -95,17 +70,3 @@ func getPrivateIPIfAvailable() net.IP {
...
-	for _, address := range addrs {
-		// check the address type and if it is not a loopback the display it
-		if ipnet, ok := address.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
-			if ipnet.IP.To4() != nil {
-				return ipnet.IP
-			}
-		}
-	}
-	return net.IPv4zero
-}
```
rshriram: Where is the replacement? You removed these and you are still using these functions in main

kyessenov: I'm not sure using local inet is a good idea. How do we pick the right interface if there are multiple available? What about ipv6?

rshriram: He is removing this function.. See this: https://github.com/istio/pilot/pull/1120/files#diff-10b44b0c268f7471a1cd69e4e0f59cc0L80 which is more generic.

rshriram: @GregHanson I feel that this file is not envoy specific. We could probably move this into proxy/

PR #1122, opened 2017-08-28
revert GrpcEcho -> Echo

rshriram:
Revert renaming of Echo gRPC service to GrpcEcho service (as some piece of s**t is caching old pb.go files and refusing to flush the cache).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/client/client.go
+++ b/test/client/client.go
@@ -155,12 +155,12 @@ func makeWebSocketRequest(client *websocket.Dialer) func(int) func() error {
 	}
 }
 
-func makeGRPCRequest(client pb.GrpcEchoTestServiceClient) func(int) func() error {
+func makeGRPCRequest(client pb.EchoTestServiceClient) func(int) func() error {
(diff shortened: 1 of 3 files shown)


ldemailly:
I think it'd be super valuable to figure out what's happening as it'll happen again if we don't

rshriram:
Bet is on caching. Bazel cache somewhere. Messing up the code. If you look at tests, you will find that its always going back to old version of echo.proto. Given that we have no script (other than bazel directive) to generate the pb.go files, I wager that somewhere, the cache is still holding on to older version and not picking up the changes. And given that the test passes sometimes, the node where my PR got processed is probably the node where the cache is upto date, while other nodes are still stale. This is a wild theory :).

ldemailly:
Another wild theory is that it doesn't get deployed properly - someone knowledgable (I'm not) should be able to tell what's happening from the output (or we should print image versions etc... on failing tests if we aren't already)

Issue #1132, opened 2017-08-29
Remove hardcoded `NamespaceAll` from service controller


kyessenov:
See https://github.com/istio/pilot/commit/720d751f5a82dcd746e3b36d8dadee96647c1d17#diff-9b493b4288200b30b3d0953052d6abdaR86
The service controller always listens for all namespaces, which will likely cause flakiness in our integration tests. This should be customizable just like other controllers (ingress and CRDs).

Issue #1131, opened 2017-08-29
Remove `IstioNamespace` from core config metadata


kyessenov:
This is against k8s standard metadata fields.
See https://github.com/istio/pilot/commit/720d751f5a82dcd746e3b36d8dadee96647c1d17#diff-2279f86634da3f787f279961af76713aR46

andraxylia:
Can you send a reference or more details explaining the issue?

ldemailly:
I think the idea is the only "namespace" we need is the watched namespace(s)

so it's the same issue as #1141 

I think we can live with whichever for 0.2 as long as it's clearly commented which is which

PR #1141, opened 2017-08-30
Eliminate `Namespace` and `AppNamespace` duplication from controller options 

andraxylia:
The controller should only have a single namespace option parameter for namespaces it monitors.
The namespace in which it runs can be obtained from env var POD_NAMESPACE.

Fixes istio/pilot#1130
Fixes istio/pilot#1131
Fixes istio/pilot#1132

**Release note**:

```release-note
```


adapter/config/ingress/status.go
```diff
@@ -58,18 +60,18 @@ func NewStatusSyncer(mesh *proxyconfig.ProxyMeshConfig, client kubernetes.Interf
...
+				return client.ExtensionsV1beta1().Ingresses(options.WatchedNamespace).Watch(opts)
 			},
 		},
 		&v1beta1.Ingress{}, options.ResyncPeriod, cache.Indexers{},
 	)
 
 	var publishService string
 	if mesh.IngressService != "" {
-		publishService = fmt.Sprintf("%v/%v", options.Namespace, mesh.IngressService)
+		publishService = fmt.Sprintf("%v/%v", os.Getenv("POD_NAMESPACE"), mesh.IngressService)
```
kyessenov: It's not a good practice to read environment variables from within libraries. Can this be passed as a parameter to the constructor?

andraxylia: Why is this not a good practice?

Then we are back to having two namespaces, and the whole point of this issue is moot.

kyessenov: It leads to unexpected dependencies (e.g. ingress library also requires this environment variable and panics without it). Would it make more sense to require that ingress service in mesh is fully qualified?

andraxylia: If reading env variable is not good for the reasons you mentioned, then we need to pass the pod namespace as a parameter, for the reasons I enumerated above, which leads to two namespaces, which makes the changes in this PR not needed.
Passing the fully qualified service for the ingress is a hack and would not apply properly to other uses.

ldemailly: IMO there should a global context (global part of the config that can't change like where is the process running) and a local context that (may) change for each controller

kyessenov: The global context is the mesh config, and we have already defined the `IngressService` name in there. Till now, it was not FQ since it's local to a namespace. If we do per-cluster installation, then the service name should be qualified.

ldemailly: k so can we add the istio-namespace there too ? or simply in a bag of objects

I don't think you'd want to get a fully qualified istio-pilot.namespace.svc.cluster.local and then try to parse out namespace, let's just have the namespace in it's own IstioNamespace somewhere ?

andraxylia:
closing this in favor of istio/pilot#1148

PR #1148, opened 2017-08-30
Put ingress and egress in the app namespace + auto create config map from istioctl kube-inject

andraxylia:

Fixes istio/pilot#1154 (cannot have multiple CA  in the same cluster, listening to all namespaces)
Solution is to limit the CA to the app namespace (it can only restrict to one namespace currently). Since CA also generates secrets for ingress and egress, in adition to apps, moved ingress and egress to the app namespace.

Works around istio/pilot#1153 to improve UX, but it is not the final fix.

Renames appNamespace to watchedNamespace and other minor fixes.

Fixes istio/pilot#1154
Fixes istio/pilot#1132
Fixes istioctl build from go, broken by some blackfriday dependency in md2man (?)
Upgrades mixer image so we get the pluralization.

**Release note**:

```release-note
```


platform/kube/controller.go
```diff
@@ -37,17 +37,16 @@ const (
...
 )
 
 // ControllerOptions stores the configurable attributes of a Controller.
 type ControllerOptions struct {
-	// Namespace to restrict controller to (empty to disable restriction)
-	Namespace    string
-	AppNamespace string
-	ResyncPeriod time.Duration
-	DomainSuffix string
+	// Namespace in which the controller runs. Can be empty if the controller does not run in a namespace
```
ldemailly: 2 points:

- "Can be empty if the controller does not run in a namespace" is that possible ? (controller outside of k8s ??)
- Is there a downside calling this IstioNamespace for clarity (vs the other one)

cmd/istioctl/main.go
```diff
@@ -41,9 +41,8 @@ const (
 var (
 	platform string
 
-	kubeconfig     string
-	namespace      string
-	istioNamespace string
```
ldemailly: so it's not needed anywhere (like kube-inject) ? how will it find the istio components ? (/what creates the meshconfig for a given namespace ?)

platform/kube/controller.go
andraxylia: Yes, for instance in rawVM, or in a standalone process in the computer.

We cannot call it istioNamespace, it is a generic field. "istio" prefix does not add much value, nor clarity.

cmd/istioctl/main.go
andraxylia: You have a point. As a previous step, I was creating the config map in every namespace with kubectl, but maybe we can do it programatically, in which case I need to revert this change.

platform/kube/controller.go
ldemailly: ok, thanks for the comment.

(semi rhetorical question: if it can be empty, what is it used for ? does it work with empty (there is a test for creating controller with empty namespace, yet it's used for ... ?) - ditto is there a test for empty vs non empty WatchedNamespace ?)

cmd/istioctl/main.go
ldemailly: I would think it'd be best to not denormalize the config map but have pointers in each managed namespace to which namespace is the istio one (managing the namespace)
ie only have 1 required istio-namespace property ? technically you don't even need to remember after injection or do you need that map at runtime in sidecars somehow ?

andraxylia: Thanks for bringing this up (the first part), it could affect the ux.
I added code in istioctl kube-inject to copy the configMap in the app namespace, until I have time to properly implemented with a namespace controller:
https://github.com/istio/pilot/issues/1153

We cannot de-normalize, it would still mean cross-namespace, and we need to map the actual data.

cmd/istioctl/inject.go
```diff
@@ -113,12 +113,24 @@ kubectl get deployment -o yaml | istioctl kube-inject -f - | kubectl apply -f -
...
 			if err != nil {
 				return fmt.Errorf("Istio configuration not found. Verify istio configmap is "+
 					"installed in namespace %q with `kubectl get -n %s configmap istio`",
-					namespace, namespace)
+					istioNamespace, istioNamespace)
 			}
+
+			// Temporary hack (few days), until this is properly implemented
+			// https://github.com/istio/pilot/issues/1153
+			_, _, err = inject.GetMeshConfig(client, namespace, meshConfig)
```
ayj: This is error prone, though perhaps not a huge deal if its very short lived.

Expected usage:
```bash
# Correct
kubectl apply -n NAMESPACE -f <(istioctl kube-inject -n NAMESPACE -f deployment.yaml)
istioctl kube-inject -n NAMESPACE -f deployment.yaml | kubectl appy -n NAMESPACE -f -

# Bug!
kubectl apply -n NAMESPACE -f <(istioctl kube-inject -f deployment.yaml)
istioctl kube-inject -f deployment.yaml | kubectl appy -n NAMESPACE -f -
```

platform/kube/inject/configmap.go
```diff
@@ -0,0 +1,76 @@
...
+// CreateMeshConfigMap copies a configMap in another namespace
+func CreateMeshConfigMap(kube kubernetes.Interface, namespace, name string, srcConfigMap *v1.ConfigMap) (*v1.ConfigMap, error) {
+
+	destConfigMap := &v1.ConfigMap{
+		TypeMeta:   metav1.TypeMeta{Kind: "ConfigMap", APIVersion: "v1"},
+		ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace},
+		Data:       srcConfigMap.Data,
+	}
+
+	destConfigMap, err := kube.CoreV1().ConfigMaps(namespace).Create(destConfigMap)
```
ayj: You return immediatly here, e.g.

```golang
return kube.CoreV1().ConfigMaps(namespace).Create(destConfigMap)
```

andraxylia:
Gopkg.lock and WORKSPACE are fixing the istioctl build from plain go, which is broken because of some blackfriday package.

platform/kube/inject/configmap.go
andraxylia: done.

cmd/istioctl/inject.go
andraxylia: Yes, user need to specify the namespace in the istioctl kube-inject command as well, in addition to the kubectl apply (so twice). They will really want to use the new initializer.

test/integration/infra.go
```diff
@@ -156,19 +156,21 @@ func (infra *infra) setup() error {
 	//
 	// See github.com/kubernetes/kubernetes/issues/49048 for k8s
 	// tracking issue.
-	if yaml, err := fill("initializer.yaml.tmpl", infra); err != nil {
-		return err
-	} else if err = infra.kubeDelete(yaml, infra.IstioNamespace); err != nil {
-		glog.Infof("Sidecar initializer could not be deleted: %v", err)
-	}
+	if infra.UseInitializer {
```
ayj: Why the change? InitializerConfiguration is clustered scope so each namespace needs to managed by an initializer even if the injection policy OFF. These were intentionally deployed regardless of InitializerConfiguration to avoid completely breaking shared test clusters.

andraxylia: reverted

adapter/config/ingress/controller.go
```diff
@@ -56,15 +56,16 @@ func NewController(client kubernetes.Interface, mesh *proxyconfig.ProxyMeshConfi
...
 
-	glog.V(2).Infof("Ingress controller watching namespaces %s", options.AppNamespace)
+	glog.V(2).Infof("Ingress controller running in namespace %s, watching namespaces %s",
+		options.Namespace, options.WatchedNamespace)
 	// informer framework from Kubernetes
 	informer := cache.NewSharedIndexInformer(
 		&cache.ListWatch{
 			ListFunc: func(opts meta_v1.ListOptions) (runtime.Object, error) {
-				return client.ExtensionsV1beta1().Ingresses(options.AppNamespace).List(opts)
+				return client.ExtensionsV1beta1().Ingresses(options.WatchedNamespace).List(opts)
```
rshriram: What if I want one ingress controller for all namespaces? Do I have an option for that ? That is the default and common installation.

cmd/istioctl/inject.go
rshriram: Not if their cluster admin has disabled initializers.

andraxylia: If they disabled initializers they will have to provide the extra params to istioctl kube-inject,

adapter/config/ingress/controller.go
ayj: `WatchedNamespace = metav1.NamespaceAll`?

andraxylia: In that case you do not pass the argument -a when starting the controller, and it will by default watch all Namespaces. I explicity added the Namespace all in -a arg, I hope this clarifies.

discoveryCmd.PersistentFlags().StringVarP(&flags.controllerOptions.WatchedNamespace, "app namespace",
		"a", metav1.NamespaceAll,
		"Restrict the applications namespace the controller manages; if not set, controller watches all namespaces")

rshriram: Yep. Thanks

Gopkg.lock
```diff
@@ -101,7 +101,7 @@
   branch = "master"
   name = "github.com/golang/protobuf"
   packages = ["jsonpb","proto","ptypes","ptypes/any","ptypes/duration","ptypes/struct","ptypes/timestamp"]
-  revision = "ab9f9a6dab164b7d1246e0e688b0ab7b94d8553e"
+  revision = "83cd65fc365ace80eb6b6ecfc45203e43edfbc70"
```
ldemailly: do those match bazel workspace ? would be neat if the tool would add the date

Gopkg.toml
```diff
@@ -93,3 +93,7 @@
 [[constraint]]
   name = "k8s.io/ingress"
   revision = "0c6f15e372c831de52fcc393932540bb3a6d51b5"
+
+[[constraint]]
+  name = "github.com/russross/blackfriday/"
```
ldemailly: why do we need that ? (markdown processor)

platform/kube/controller.go
```diff
@@ -43,11 +43,12 @@ const (
...
 // ControllerOptions stores the configurable attributes of a Controller.
 type ControllerOptions struct {
-	// Namespace to restrict controller to (empty to disable restriction)
-	Namespace    string
-	AppNamespace string
-	ResyncPeriod time.Duration
-	DomainSuffix string
+	// Namespace in which the controller runs. Can be "" if the controller does not run in a namespace
+	Namespace string
+	// Namespace the controller watches. If set to meta_v1.NamespaceAll (""), controller watches all namespaces
```
ldemailly: Thank you !

andraxylia:
PTAL and approve the commit , pilot e2e is passing.

e2e smoketest failures are in mixer tests and are unrelated to my changes.

platform/kube/inject/configmap.go
```diff
@@ -0,0 +1,74 @@
...
+	"k8s.io/client-go/kubernetes"
+
+	proxyconfig "istio.io/api/proxy/v1/config"
+	"istio.io/pilot/model"
+	"istio.io/pilot/proxy"
+)
+
+// GetMeshConfig fetches the ProxyMesh configuration from Kubernetes ConfigMap.
+func GetMeshConfig(kube kubernetes.Interface, namespace,
+	name string) (*v1.ConfigMap, *proxyconfig.ProxyMeshConfig, error) {
```
ldemailly: it seems (or my search failed) that you always ignore the first argument, so why return it and change to _ everywhere ?

also next time consider doing a cp + commit + changes so git and the diff would know this is coming from 
platform/kube/inject/inject.go

also why the new file ?

andraxylia: it's used here:
istioMeshConfigMap, istioMeshConfig, err := inject.GetMeshConfig(client, istioNamespace, meshConfigMapName)

andraxylia: new file so it can be re-used in the initializer, Jason asked it to be like this.

ldemailly: ok sorry I missed that / glossed over the "hack" section

afaik (unless there is some bazel magic) the file being in the same package there is 0 difference ?

Gopkg.toml
kyessenov: can you please not mix-up these  godep stuff with the actual PR?

cmd/pilot-discovery/main.go
```diff
@@ -193,8 +193,9 @@ func init() {
 		fmt.Sprintf("File name for Istio mesh configuration"))
 	discoveryCmd.PersistentFlags().StringVarP(&flags.controllerOptions.Namespace, "namespace", "n", "",
 		"Select a namespace for the controller loop. If not set, uses ${POD_NAMESPACE} environment variable")
-	discoveryCmd.PersistentFlags().StringVarP(&flags.controllerOptions.AppNamespace, "app namespace", "a", "",
-		"Restrict the applications namespace that controller manages, do n")
+	discoveryCmd.PersistentFlags().StringVarP(&flags.controllerOptions.WatchedNamespace, "app namespace",
```
kyessenov: why have both `a` and `n` flags that seemingly do same thing?

ldemailly:
/lgtm
(I'm sure you'll fix stuff it gets worse)

platform/kube/controller.go
kyessenov: Why did you change the meaning of Namespace field? The previous semantics was WatchedNamespace.

test/integration/driver.go
```diff
@@ -61,7 +61,7 @@ const (
 	caImage = "gcr.io/istio-testing/istio-ca:2baec6baacecbd516ea0880573b6fc3cd5736739"
 
 	// mixerImage specifies the default mixer docker image used for e2e testing *update manually*
-	mixerImage = "gcr.io/istio-testing/mixer:49e721e15d481cd5d92d9a2b30b5e8fcdcafdb63"
+	mixerImage = "gcr.io/istio-testing/mixer:986c6e1eb9c7dcff42e3df5aefa9ce858ece885e"
```
kyessenov: this is not a minor change, don't bundle the subtle things like this

test/integration/infra.go
```diff
@@ -190,7 +190,7 @@ func (infra *infra) setup() error {
 		}
 	}
 	if infra.Ingress {
-		if err := deploy("ingress-proxy.yaml.tmpl", infra.IstioNamespace); err != nil {
+		if err := deploy("ingress-proxy.yaml.tmpl", infra.Namespace); err != nil {
```
kyessenov: ingress proxy should be in istio namespace.

test/integration/infra.go
```diff
@@ -202,7 +202,7 @@ func (infra *infra) setup() error {
 		if err != nil {
 			return err
 		}
-		_, err = client.CoreV1().Secrets(infra.IstioNamespace).Update(&v1.Secret{
+		_, err = client.CoreV1().Secrets(infra.Namespace).Update(&v1.Secret{
```
kyessenov: secret should be in istio namespace.

test/integration/infra.go
```diff
@@ -214,7 +214,7 @@ func (infra *infra) setup() error {
 		}
 	}
 	if infra.Egress {
-		if err := deploy("egress-proxy.yaml.tmpl", infra.IstioNamespace); err != nil {
+		if err := deploy("egress-proxy.yaml.tmpl", infra.Namespace); err != nil {
```
kyessenov: same

test/integration/infra.go
```diff
@@ -391,7 +391,7 @@ func (infra *infra) applyConfig(inFile string, data map[string]string) error {
 	}
 
 	glog.Info("Sleeping for the config to propagate")
-	time.Sleep(3 * time.Second)
+	time.Sleep(20 * time.Second)
```
kyessenov: don't bundle this here. i have a fix in another PR. just wait for cluster to quiet down to pass tests.

Gopkg.toml
andraxylia: I cannot not mix it, because I need to build. Why does this bother you?

Gopkg.lock
andraxylia: yes, I changed it to match.

platform/kube/inject/configmap.go
andraxylia: there is a new function, plus changes in the original fct. Please have a better look.

cmd/pilot-discovery/main.go
andraxylia: They do not control the same thing.

test/integration/infra.go
andraxylia: The purpose of this change is to allow multiple runs, this is why ingress and egress have been moved back to the app namespace. When we will have isolated cluster, they can be run in IstioNamespace.

test/integration/infra.go
andraxylia: same as above.

test/integration/infra.go
andraxylia: I cannot wait...

test/integration/driver.go
andraxylia: We need the pluralization from mixer.

andraxylia: It's not subtle, it passed pilot e2e tests, it can be upgraded.

andraxylia: It it were easier to get a change committed, we would not end up bundling multiple fixes in one PR.

cmd/pilot-discovery/main.go
andraxylia: Example where they are different is ingress. This PR cleans up the misconception they are the same.
-n was already associated with POD_NAMESPACE, and the command line, and in general associated with the namespace where the k8s object is deployed, so I chose -a to restrict the namespace.

test/integration/infra.go
andraxylia: If your PR gets in faster, it will ask me to merge and it will disappear. What value do you use?

platform/kube/inject/configmap.go
ldemailly: would be easier to have a better look if it showed as a diff than a big green block

test/integration/infra.go
kyessenov: Yes you merged this? Really?

PR #1149, opened 2017-08-30
Ingress controller loads TLS cert/key from volume mount

### LINKED ISSUE #1134: Use volume mount to propagate TLS key/cert to Ingress
This is a security enhancement. Istio ingress should get the TLS cert/key for cluster-inbound traffic from a secure channel.

Currently, the flow of ingress TLS cert/key propagation is as follows:
The operator sets the TLS cert/key for ingress through a resource. Pilot reads the resource from apiserver, and exposes the cert/key through unsecure HTTP. Istio ingress's pilot-agent dynamically reads the HTTP endpoint and writes the TLS cert/key to the local file system for Envoy to consume.
(Tutorial: https://istio.io/docs/tasks/ingress.html#configuring-secure-ingress-https)

We plan to keep the same story for TLS cert/key propagation as how we propagate the in-cluster cert/key:
The operator sets the TLS cert/key for ingress through a secret. The secret is mounted as a volume when ingress starts. To support dynamically setting cert/key, Istio ingress's pilot-agent will restart Envoy when a new key/cert pair is set to the secret.

myidpt:
This PR removes the secret HTTP endpoint between Pilot and ingress controller, which is a security hole that the private key is transferred through HTTP. Instead, the operator should use volume mount to directly mount the key/cert from secret.

This PR, removing secret HTTP endpoint, partially resolves #1134. Another PR on istio/istio solves the volume mount part.

**Release note**:
```release-note
Ingress controller loads TLS cert/key from volume mount
```


myidpt:
@kyessenov PTAL. Thanks!

kyessenov:
Where's the code that writes to a secret and monitors it?

myidpt:
As we discussed yesterday, the secret is directly mounted to ingress controller. The operator should create/replace the secret. The secret is already monitored at https://github.com/istio/pilot/blob/720d751f5a82dcd746e3b36d8dadee96647c1d17/proxy/envoy/watcher.go#L101. We keep the watcher.
Initially, the ingress controller mounts an empty key/cert pair. See https://github.com/istio/istio/pull/623

rshriram:
Sorry, I didn't close this. I approved a different PR which incorrectly referenced this PR as the issue.

kyessenov:
There's a test in testdata that tries to submit TLS-enabled ingress. Can you please update the test and then later update the docs how to use ingress with TLS?

myidpt:
@kyessenov Sure thing, I will update the test.
What's the doc you want to update? I don't think we want to update istio.github.io because it's based on the V0.1 release, right?

ldemailly:
@myidpt istio.github.io is open for 0.2 updates, we need to get started some day or another
(v1 is archived in https://istio.io/v-0.1/docs/ so that's not an issue )
this being said I think you'd be the first one

myidpt:
@ldemailly I see. No problem, I will update the doc once this PR is merged :)

ldemailly:
I am surprised this doesn't break something as you meantioned a istio/istio change needed to mount the secret. are the tests not covering/using that http download ?

myidpt:
@ldemailly Actually I changed the e2e test to mount the secret, and the infra to change the key/cert to real ones. Look at the change in test/integration/testdata/ingress-proxy.yaml.tmpl, and test/integration/infra.go :)

PR #1151, opened 2017-08-31
Fix the bug that when no service account is specified in annotation, empty service accounts are returned.

myidpt:
```
NONE
```

### DIFF
--- a/platform/kube/conversion.go
+++ b/platform/kube/conversion.go
@@ -83,11 +83,15 @@ func convertService(svc v1.Service, domainSuffix string) *model.Service {
 
 	serviceaccounts := make([]string, 0)
 	if svc.Annotations != nil {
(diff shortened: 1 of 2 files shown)


myidpt:
@wattli PTAL. Thanks!

ldemailly:
what's the desired behavior when there is no annotation ? which service account is used?
cc @costinm we were just talking about this

myidpt:
@ldemailly @costinm 
The annotation is to mark the service accounts that run the service on VMs. Note that the service can be run by Kubernetes pods, in which case the pods' service accounts do not need to be annotated (since this info can be retrieved from apiserver).
So if the annotation is empty, the service may still be run by the K8s pods. And the secure naming info is a mapping from the service to the pods' service accounts.

wattli:
Sorry for the late response, good work, thanks!

PR #1155, opened 2017-08-31
[WIP] Support headless services

ijsnellf:
Support headless services. Proxies/routes connections to headless services to service instances.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: WILL UPDATE

This solution may not be practical due to the amount of querying that must be done.

Egress and auth are currently broken. Integration tests need to be added to the normal e2e flow (right now all services are headless).

**Release note**:

```release-note
Added support for headless services.
```


proxy/envoy/route.go
```diff
@@ -316,14 +316,16 @@ func sharedHost(parts ...[]string) []string {
...
-func buildTCPRoute(cluster *Cluster, addresses []string) *TCPRoute {
+func buildTCPRoute(cluster *Cluster, addresses []string, ports string) *TCPRoute {
 	// destination port is unnecessary with use_original_dst since
 	// the listener address already contains the port
 	route := &TCPRoute{
-		Cluster:    cluster.Name,
-		clusterRef: cluster,
+		Cluster:           cluster.Name,
+		DestinationIPList: make([]string, 0, len(addresses)),
+		DestinationPorts:  ports,
```
rshriram: Does tcp route have multiple ports?

ijsnellf: It can. But we are only doing single port right now.

kyessenov:
Agree, sounds like the only way to handle the pod churn is by marking the endpoint port, reserving it as unique per cluster, and routing anything on 0.0.0.0:ENDPOINT_PORT as passthrough. This actually has the benefit of localizing the requirement to the pod deployment, the user just needs to pick a high port for the pod startup flags.

ijsnellf:
I'm closing this PR and opening https://github.com/istio/pilot/pull/1197 since the new logic is entirely different.

PR #1158, opened 2017-09-01
add webhook validation

ayj:

Add the `istio.io/pilot/platform/kube/admit` package which implements the ExternalAdmissionWebhook for server-side validation of Pilot configuration. This includes a script to enable the workaround described [here](https://github.com/kubernetes/kubernetes/issues/49987#issuecomment-319739227).

This feature requires 1.7.x with [external-admission-webhooks](https://kubernetes.io/docs/admin/extensible-admission-controllers/#external-admission-webhooks). General instructions to enable are [here](https://kubernetes.io/docs/admin/extensible-admission-controllers/#enable-external-admission-webhooks). Specific testing instructions for GKE and Minikube are documented [here](https://github.com/istio/pilot/blob/master/doc/testing.md#create-development-cluster).

Fixes https://github.com/istio/pilot/issues/1066.

UPDATED: this PR includes integration with the istio-pilot discovery service. The admission registration is disabled by default since it requires an additional workaround for GKE (see platform/kube/admit/webhook-workaround.sh). 

Users can experiment with this feature on GKE using the e2e pre-submit test with the following sequence of commands.

1) run `bin/e2e.sh -use-admission-webhook`
2) cancel the test after the deployment and test pods are created
3) run `platform/kube/admit/webhook-workaround.sh` and specify the server-name, secret-name, namespace, and port number of the pilot discovery service, e.g.

```bash
platform/kube/admit/webhook-workaround.sh \ 
    --service-name istio-pilot \
    --secret-name pilot-webhook \
    --namespace istio-test-w1g0f \
    --port 443
```

This script converts the istio-pilot discovery service to type=LoadBalancer, waits for ingress IP, creates the workaround external service/endpoint, generates required CA and server certs for webhook, and creates the corresponding k8s secret which is consumed by the admission webhook server.

4) wait until the LoadBalancer IP forwarding rule is propagated in GCP. You can try curl'ing `<ExternalIP>:8000/v1/registration` until a valid response is returned.

5) Create a (in)valid pilot configuration via kubectl (or istioctl). Invalid configuration should be rejected by apiserver with appropriate error message. 

**Release note**:

```release-note
Add the `istio.io/pilot/platform/kube/admit` package which implements the ExternalAdmissionWebhook for server-side validation of Pilot configuration. This includes a script to enable the workaround described [here](https://github.com/kubernetes/kubernetes/issues/49987#issuecomment-319739227).

This feature requires 1.7.x with [external-admission-webhooks](https://kubernetes.io/docs/admin/extensible-admission-controllers/#external-admission-webhooks). General instructions to enable are [here](https://kubernetes.io/docs/admin/extensible-admission-controllers/#enable-external-admission-webhooks). Specific testing instructions for GKE and Minikube are documented [here](https://github.com/istio/pilot/blob/master/doc/testing.md#create-development-cluster).
```


platform/kube/admit/admit.go
```diff
@@ -0,0 +1,248 @@
...
+	// DefaultAdmissionHookName is the default name for the
+	// ExternalAdmissionHooks.
+	DefaultAdmissionHookName = "pilot.config.istio.io"
+
+	// DefaultAdmissionServiceName is the default service of the
+	// validation webhook.
+	DefaultAdmissionServiceName = "istio-pilot-config"
+)
+
+// Admit implements the external admission webhook for validation
```
kyessenov: validation _of_

platform/kube/admit/admit.go
```diff
@@ -0,0 +1,248 @@
...
+	// ExternalAdmissionHookConfiguration.
+	DefaultAdmissionHookConfigName = "pilot-config"
+
+	// DefaultAdmissionHookName is the default name for the
+	// ExternalAdmissionHooks.
+	DefaultAdmissionHookName = "pilot.config.istio.io"
+
+	// DefaultAdmissionServiceName is the default service of the
+	// validation webhook.
+	DefaultAdmissionServiceName = "istio-pilot-config"
```
kyessenov: should this be pilot itself? seems unnecessary to allocate a service just for validation.

platform/kube/admit/admit.go
```diff
@@ -0,0 +1,248 @@
...
+	serviceName      string
+	serviceNamespace string
+	caBundle         []byte
+
+	// unconditionally validate all config that is not in this list of
+	// configuration.
+	validateNamespaces []string
+}
+
+// GetAPIServerExtensionCACert gets the CA cert that will signed the
```
kyessenov: will _be_ signed _by_

platform/kube/admit/admit.go
```diff
@@ -0,0 +1,248 @@
...
+// cert used by the "GenericAdmissionWebhook" plugin admission
+// controller.
+func GetAPIServerExtensionCACert(cl kubernetes.Interface) ([]byte, error) {
+	const name = "extension-apiserver-authentication"
+	c, err := cl.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(name, metav1.GetOptions{})
```
kyessenov: Why not a secret to store CA cert?

platform/kube/admit/admit.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+	return &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		ClientCAs:    caCertPool,
+		ClientAuth:   tls.RequireAndVerifyClientCert,
+	}, nil
+}
+
+// New creates a new instance of the admission webhook controller.
+func New(descriptor model.ConfigDescriptor, hookConfigName, hookName, serviceName, serviceNamespace string, validateNamespaces []string, caBundle []byte) *Admit { // nolint: lll
```
kyessenov: nit: constructors are usually called `Create` or `Make` around here. Too bad there's no standard name for the pattern. Also, `Admit` does not sound good as a thing, maybe call it `AdmissionController`.

platform/kube/admit/admit.go
kyessenov: Is kube-system the right place to store the config? Is the name of the config standard?

platform/kube/admit/admit.go
```diff
@@ -0,0 +1,248 @@
...
+	// validation webhook.
+	DefaultAdmissionServiceName = "istio-pilot-config"
+)
+
+// Admit implements the external admission webhook for validation
+// pilot configuration.
+type Admit struct {
+	descriptor       model.ConfigDescriptor
+	hookConfigName   string
+	hookName         string
```
kyessenov: what are these hook things? comments?

platform/kube/admit/admit.go
```diff
@@ -0,0 +1,248 @@
...
+		hookConfigName:     hookConfigName,
+		hookName:           hookName,
+		serviceName:        serviceName,
+		serviceNamespace:   serviceNamespace,
+		caBundle:           caBundle,
+		validateNamespaces: validateNamespaces,
+	}
+}
+
+// Unregister registers the external admission webhook
```
kyessenov: _unregisters_

platform/kube/admit/admit.go
```diff
@@ -0,0 +1,248 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: a.hookConfigName,
+		},
+		ExternalAdmissionHooks: []admissionregistrationv1alpha1.ExternalAdmissionHook{
+			{
+				Name: a.hookName,
+				Rules: []admissionregistrationv1alpha1.RuleWithOperations{{
+					Operations: []admissionregistrationv1alpha1.OperationType{
+						admissionregistrationv1alpha1.Create,
+						admissionregistrationv1alpha1.Update,
```
kyessenov: PATCH?

platform/kube/admit/admit.go
```diff
@@ -0,0 +1,248 @@
...
+func (a *Admit) Unregister(client admissionClient.ExternalAdmissionHookConfigurationInterface) error {
+	return client.Delete(a.hookConfigName, nil)
+}
+
+// Register registers the external admission webhook for pilot
+// configuration types.
+func (a *Admit) Register(client admissionClient.ExternalAdmissionHookConfigurationInterface) error {
+	var resources []string
+	for _, schema := range a.descriptor {
+		resources = append(resources, schema.Plural)
```
kyessenov: is this expecting the kind name? then it should "RouteRules", not "route-rules"

platform/kube/admit/admit.go
```diff
@@ -0,0 +1,248 @@
...
+					Service: admissionregistrationv1alpha1.ServiceReference{
+						Namespace: a.serviceNamespace,
+						Name:      a.serviceName,
+					},
+					CABundle: a.caBundle,
+				},
+			},
+		},
+	}
+	client.Delete(webhook.Name, nil) // nolint: errcheck
```
kyessenov: might be worth logging even if ignoring the error, for debugging purposes

platform/kube/admit/testcerts.go
```diff
@@ -0,0 +1,138 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This file was generated using openssl by the gencerts.sh script
+// and holds raw certificates for the webhook tests.
+
+package admit
```
kyessenov: admit_test is probably better to make it clear these are test artifacts

platform/kube/admit/admit.go
ayj: Yes, though we'll need a new port since this is HTTPS. We'll also need to temporary use a different service name as a workaround for the issue described [here](https://github.com/kubernetes/kubernetes/issues/49987#issuecomment-319739227), but that should eventually go away (k8s 1.8+).

platform/kube/admit/admit.go
ayj: That is where Kubernetes stores the aggregate apiserver's client CA certificate. We don't control this.

```bash
$ kubectl -n kube-system get cm extension-apiserver-authentication 
NAME                                 DATA      AGE
extension-apiserver-authentication   6         17d
```

platform/kube/admit/admit.go
ayj: This is the resource name, e.g. `pods` and not `Pod` (see https://kubernetes.io/docs/admin/extensible-admission-controllers/#configure-initializers-on-the-fly)

platform/kube/admit/admit.go
kyessenov: I see, then it makes sense but deserves a note in the comment.

platform/kube/admit/admit.go
ayj: There is no `PATCH` operation for admission (see https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/admissionregistration/types.go#L186

platform/kube/admit/admit.go
kyessenov: Please mention this. It might appear as if we're choosing these names.

platform/kube/admit/admit.go
kyessenov: then it's probably to remove "-" if the library code doesn't deal with it well.

rshriram:
cc @GregHanson .. you might want to update your k8s image to 1.7.4

bin/check.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 
-buildifier -showlog -mode=check $(find . -type f \( -name 'BUILD' -or -name 'WORKSPACE' -or -wholename '.*bazel' -or -wholename '.*bzl' \) -print )
+buildifier -showlog -mode=check $(find . -type f \( -name 'BUILD' -or -name 'WORKSPACE' -or -wholename '.*bazel$' -or -wholename '.*bzl$' \) -print )
```
ldemailly: good fix!

ayj:
I went ahead and integrated the webhook with istio-pilot discovery service. It's off by default and requires an additional helper script to work on GKE (see platform/kube/admit/webhook-workaround.sh). I set things up so the majority of the code can be used as-is with some minimal changes to cert derivation/distribution w/o `webhook-workaround.sh` once k8s 1.8 on GKE is available.

platform/kube/admit/admit.go
```diff
@@ -0,0 +1,394 @@
...
+	// good. This is useful when multiple validators are running in
+	// the same cluster managing different sets of namespaces
+	// (e.g. shared test clusters).
+	ValidateNamespaces []string
+
+	// // CAbundle is the PEM encoded CA bundle which will be used to
+	// // validate webhook's service certificate.
+	// CABundle []byte
+
+	// DomainSuffix is the DNS domain suffix for Istio CRD resources, e.g. local.cluster.
```
kyessenov: nit: cluster.local

test/integration/testdata/rbac-beta.yaml.tmpl
```diff
@@ -29,6 +29,11 @@ rules:
 - apiGroups: [""]
   resources: ["namespaces", "nodes", "secrets"]
   verbs: ["get", "list", "watch"]
+{{if .UseAdmissionWebhook}}
```
kyessenov: nit: may drop this, since it's a superset of RBAC permissions.

rshriram:
@ayj please merge with master.. and upload test logs from your local test cluster run.. Need e2e tests from istio/istio to pass.

kyessenov:
rebase please.

PR #1165, opened 2017-09-01
Don't clear pending initializers when -use-initializer=false

ayj:

Don't clear pending initializers when -use-initializer=false

Clearing the list of pending initializers is required for admin bypass
of the initialization process for the initializer deployment. This
works correctly on k8s clusters with Initializer plugin enabled. It
doesn't work when that plugin is disabled, e.g. non-alpha GKE, default
minikube, and results in the following error:

        The Deployment “istio-sidecar-initializer” is invalid:
        metadata.initializers.pending: Invalid value: “null”: must be
        non-empty when result is not set

**Release note**:

```release-note
NONE
```


andraxylia:
Try to re-base, there have been changes that help with the smoke test.

PR #1171, opened 2017-09-02
make the virtual host's name as the key of the egress rule

vadimeisenbergibm:
It improves the debuggabilty of egress-rule's virtual hosts

**Release note**:

```NONE
```


rshriram:
commit already included in #1174

PR #1172, opened 2017-09-04
This is basically == master (no merge necessary)

ldemailly:
Let’s see if the tests pass…

```release-note
NONE
```


ldemailly:
as expected it didn't pass (503s, https://github.com/istio/istio/issues/653 , let's try another time but will probably fail again as it hasn't been passing for days)

/retest

PR #1173, opened 2017-09-04
Added quote for inject policy config.

gyliu513:
"Off" without quote will be treated as false, so we need to make
sure the injection policy config parameters include a quote so
that we can handle "off" well.

Check here for detail about "off": http://yaml.org/type/bool.html

**Release note**:

```release-note
NONE
```

/cc @ayj

### DIFF
--- a/test/integration/testdata/initializer-configmap.yaml.tmpl
+++ b/test/integration/testdata/initializer-configmap.yaml.tmpl
@@ -4,18 +4,18 @@ metadata:
   name: istio-inject
 data:
   config: |-
-    policy: {{.Policy}}
+    policy: "{{.Policy}}"
     namespaces:
 {{ range $namespace := .Namespaces}}
-    - {{$namespace}}
+    - "{{$namespace}}"
 {{ end }}
     params:
-      initImage: {{.Params.InitImage}}
-      proxyImage: {{.Params.ProxyImage}}
+      initImage: "{{.Params.InitImage}}"
+      proxyImage: "{{.Params.ProxyImage}}"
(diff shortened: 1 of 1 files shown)


test/integration/testdata/initializer-configmap.yaml.tmpl
```diff
@@ -4,18 +4,18 @@ metadata:
...
       SidecarProxyUID: {{.Params.SidecarProxyUID}}
-      Version: {{.Params.Version}}
+      Version: "{{.Params.Version}}"
       EnableCoreDump: {{.Params.EnableCoreDump}}
-      MeshConfigMapName: {{.Params.MeshConfigMapName}}
-      ImagePullPolicy: {{.Params.ImagePullPolicy}}
-      IncludeIPRanges: {{.Params.IncludeIPRanges}}
+      MeshConfigMapName: "{{.Params.MeshConfigMapName}}"
+      ImagePullPolicy: "{{.Params.ImagePullPolicy}}"
+      IncludeIPRanges: "{{.Params.IncludeIPRanges}}"
```
rshriram: Need quotes here, because we can have a comma separated list I think.

rshriram:
Merging based on @ayj 's approval.

test/integration/testdata/initializer-configmap.yaml.tmpl
gyliu513: Yes, not only because of list, there are still possibility that the end user may use some key words in go yaml, such as `off` etc. ;-)

PR #1174, opened 2017-09-04
Reject egress rules that conflict on some domain

### LINKED ISSUE #1166: Ensure that egress rules do not introduce duplicate domain / virtualhost names to RDS
This is a follow-up from https://github.com/istio/pilot/pull/1142. 
Now that egress rules are sent to all proxies, we should be careful not to push breaking config to all sidecars. That can happen if there are two egress rules for the same domain. We should do stricter validation for the egress HTTP routes and pro-actively eliminate duplicates during config generation.

vadimeisenbergibm:
Prevent conflicting egress rules from causing envoy to crash.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1166

**Release note**:

```
```

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strconv"
 
 	"github.com/golang/protobuf/proto"
 	multierror "github.com/hashicorp/go-multierror"
@@ -496,17 +497,29 @@ func RejectConflictingEgressRules(egressRules map[string]*proxyconfig.EgressRule
 	}
 	sort.Strings(keys)
 
+	// domains - a map where keys are of the form domain:port and values are the keys of
+	// egress-rule configuration objects
 	domains := make(map[string]string)
-	for _, key := range keys {
-		egressRule := egressRules[key]
+	for _, egressRuleKey := range keys {
+		egressRule := egressRules[egressRuleKey]
 		conflictingRule := false
+
+	DomainsAndPortsLoop:
 		for _, domain := range egressRule.Domains {
-			var keyOfAnEgressRuleWithTheSameDomain string
-			keyOfAnEgressRuleWithTheSameDomain, conflictingRule = domains[domain]
-			if conflictingRule {
-				errs = multierror.Append(errs, fmt.Errorf("rule %q conflicts with rule %q on domain %s, is rejected",
-					key, keyOfAnEgressRuleWithTheSameDomain, domain))
-				break
+			for _, port := range egressRule.Ports {
+				portNumber := int(port.Port)
+				domainsKey := domain + ":" + strconv.Itoa(portNumber)
+
+				var keyOfAnEgressRuleWithTheSameDomain string
+				keyOfAnEgressRuleWithTheSameDomain, conflictingRule = domains[domainsKey]
+				if conflictingRule {
+					errs = multierror.Append(errs,
+						fmt.Errorf("rule %q conflicts with rule %q on domain "+
+							"%s and port %d, is rejected", egressRuleKey,
+							keyOfAnEgressRuleWithTheSameDomain, domain,
+							portNumber))
+					break DomainsAndPortsLoop
+				}
 			}
 		}
 
@@ -515,9 +528,13 @@ func RejectConflictingEgressRules(egressRules map[string]*proxyconfig.EgressRule
 		}
 
 		for _, domain := range egressRule.Domains {
-			domains[domain] = key
+			for _, port := range egressRule.Ports {
+				portNumber := int(port.Port)
+				domainsKey := domain + ":" + strconv.Itoa(portNumber)
+				domains[domainsKey] = egressRuleKey
+			}
 		}
-		filteredEgressRules[key] = egressRule
+		filteredEgressRules[egressRuleKey] = egressRule
 	}
 
 	return filteredEgressRules, errs
--- a/model/config.go
+++ b/model/config.go
@@ -480,12 +480,17 @@ func (store *istioConfigStore) Policy(instances []*ServiceInstance, destination 
 // According to Envoy's virtual host specification, no virtual hosts can share the same domain.
 // The following code rejects conflicting rules deterministically, by a lexicographical order -
 // a rule with a smaller key lexicographically wins.
+// Here the key of the rule is the key of the Istio configuration objects - see
+// `func (meta *ConfigMeta) Key() string`
 func RejectConflictingEgressRules(egressRules map[string]*proxyconfig.EgressRule) ( // long line split
 	map[string]*proxyconfig.EgressRule, error) {
 	filteredEgressRules := make(map[string]*proxyconfig.EgressRule)
 	var errs error
 
 	var keys []string
+
+	// the key here is the key of the Istio configuration objects - see
+	// `func (meta *ConfigMeta) Key() string`
 	for key := range egressRules {
 		keys = append(keys, key)
 	}
--- a/model/config.go
+++ b/model/config.go
@@ -484,7 +484,7 @@ func RejectConflictingEgressRules(egressRules map[string]*proxyconfig.EgressRule
 	var errs error
 
 	var keys []string
-	for key, _ := range egressRules {
+	for key := range egressRules {
 		keys = append(keys, key)
 	}
 	sort.Strings(keys)
--- a/model/config.go
+++ b/model/config.go
@@ -479,7 +479,8 @@ func (store *istioConfigStore) Policy(instances []*ServiceInstance, destination 
 // According to Envoy's virtual host specification, no virtual hosts can share the same domain.
 // The following code rejects conflicting rules determenistically, by a lexicographical order - a rule with
 // a smaller name lexicographically wins.
-func RejectConflictingEgressRules(egressRules map[string]*proxyconfig.EgressRule) (map[string]*proxyconfig.EgressRule, error) {
+func RejectConflictingEgressRules(egressRules map[string]*proxyconfig.EgressRule) ( // long line split
+	map[string]*proxyconfig.EgressRule, error) {
 	filteredEgressRules := make(map[string]*proxyconfig.EgressRule)
 	var errs error
 
--- a/model/config.go
+++ b/model/config.go
@@ -475,10 +475,11 @@ func (store *istioConfigStore) Policy(instances []*ServiceInstance, destination 
 	return out
 }
 
-// RejectConflictingEgressRules rejects rules that have a domain which same to domains of some other rule.
+// RejectConflictingEgressRules rejects rules that have a domain which is equal to
+// a domain of some other rule.
 // According to Envoy's virtual host specification, no virtual hosts can share the same domain.
-// The following code rejects conflicting rules determenistically, by a lexicographical order - a rule with
-// a smaller name lexicographically wins.
+// The following code rejects conflicting rules deterministically, by a lexicographical order -
+// a rule with a smaller key lexicographically wins.
 func RejectConflictingEgressRules(egressRules map[string]*proxyconfig.EgressRule) ( // long line split
 	map[string]*proxyconfig.EgressRule, error) {
 	filteredEgressRules := make(map[string]*proxyconfig.EgressRule)
--- a/model/config.go
+++ b/model/config.go
@@ -20,6 +20,7 @@ import (
 	"sort"
 
 	"github.com/golang/protobuf/proto"
+	multierror "github.com/hashicorp/go-multierror"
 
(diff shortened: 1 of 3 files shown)


proxy/envoy/config.go
```diff
@@ -672,6 +672,12 @@ func buildEgressFromSidecarVirtualHostOnPort(rule *proxyconfig.EgressRule, mesh
 
 func buildEgressFromSidecarHTTPRoutes(mesh *proxyconfig.ProxyMeshConfig, egressRules map[string]*proxyconfig.EgressRule,
 	httpConfigs HTTPRouteConfigs) HTTPRouteConfigs {
+	egressRules, errs := model.RejectConflictingEgressRules(egressRules)
+
+	if errs != nil {
+		glog.Warningf("Rejected rules: %", errs)
```
ldemailly: %v (or %s)

ldemailly:
`make fmt` should fix most of issues. run `bin/check.sh` too before next update.

model/config.go
```diff
@@ -473,3 +474,44 @@ func (store *istioConfigStore) Policy(instances []*ServiceInstance, destination
...
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+
+	domains := make(map[string]string)
+	for _, key := range keys {
+		egressRule := egressRules[key]
+		conflictingRule := false
+		for _, domain := range egressRule.Domains {
+			keyOfAnEgressRuleWithTheSameDomain, conflictFound := domains[domain]
```
ldemailly: reuse `conflictingRule` instead of defining a new bool (and setting `conflictingRule` to true seperately) ?

vadimeisenbergibm:
Closing temporarily

ldemailly:
?

vadimeisenbergibm:
@ldemailly installing buildifier and gometalinter, running bin/check.sh and make fmt, will fix the issues and reopen the PR soon, sorry for the confusion, thank you for your review.

ldemailly:
no problem; bin/init.sh and/or install-prereqs.sh should do that for you

try rebasing your fork as it seems prow can't merge/test it

vadimeisenbergibm:
@ldemailly OK, will rebase.

vadimeisenbergibm:
@ldemailly fixed presubmit.sh errors.

model/config.go
```diff
@@ -473,3 +474,45 @@ func (store *istioConfigStore) Policy(instances []*ServiceInstance, destination
 
 	return out
 }
+
+// RejectConflictingEgressRules rejects rules that have a domain which same to domains of some other rule.
```
rshriram: Typo in sentence

proxy/envoy/config.go
```diff
@@ -664,15 +664,22 @@ func buildEgressFromSidecarVirtualHostOnPort(rule *proxyconfig.EgressRule, mesh
 	externalTrafficRoute := buildDefaultRoute(externalTrafficCluster)
 
 	return &VirtualHost{
-		Name:    rule.Domains[0] + "-" + strconv.Itoa(port.Port),
+		Name:    ruleKey + ":" + strconv.Itoa(port.Port),
```
ldemailly: why this change ? can you describe the effect here and/or in the PR summary

going from `foo.com-123` to `fooRuleKey:123` (if I understand both the old and new code correctly)

shouldn't it be `host.foo.com:123` - why was it - before ?

model/config_test.go
```diff
@@ -475,3 +475,156 @@ func TestPolicy(t *testing.T) {
...
+		got, errs := model.RejectConflictingEgressRules(c.in)
+		if (errs == nil) != c.valid {
+			t.Errorf("RejectConflictingEgressRules failed on %s: got valid=%v but wanted valid=%v",
+				c.name, errs == nil, c.valid)
+		}
+		if !reflect.DeepEqual(got, c.out) {
+			t.Errorf("RejectConflictingEgressRules failed on %s: got=%v but wanted %v: %v",
+				c.name, got, c.in)
+		}
+	}
```
rshriram: Need a test case with duplicate domains across different ports

model/config.go
```diff
@@ -473,3 +474,45 @@ func (store *istioConfigStore) Policy(instances []*ServiceInstance, destination
...
+// According to Envoy's virtual host specification, no virtual hosts can share the same domain.
+// The following code rejects conflicting rules determenistically, by a lexicographical order - a rule with
+// a smaller name lexicographically wins.
+func RejectConflictingEgressRules(egressRules map[string]*proxyconfig.EgressRule) ( // long line split
+	map[string]*proxyconfig.EgressRule, error) {
+	filteredEgressRules := make(map[string]*proxyconfig.EgressRule)
+	var errs error
+
+	var keys []string
+	for key := range egressRules {
```
rshriram: Add a comment explaining what is the key here

proxy/envoy/config.go
vadimeisenbergibm: @ldemailly this one is a commit from https://github.com/istio/pilot/pull/1171, reviewed by @kyessenov. I cherry-picked this commit into this PR for testing, forgot to reset it before submitting a PR. Once https://github.com/istio/pilot/pull/1171 is merged, this commit will not appear in this PR. Sorry for the confusion. 

The idea is to use the key of the rule as a name of the virtual host, and not the first domain of the rule. It is more clear to associate a virtual host with a rule. Since `:` are allowed in the names of Envoy virtual hosts, and this way it is more clear that it is a port, I used `:`.

model/config_test.go
vadimeisenbergibm: Done.

vadimeisenbergibm:
@rshriram thank you for your review, I have applied your comments.

vadimeisenbergibm:
@rshriram Good catch, I have fixed the test - making ports into consideration when calculating conflicting rules.

model/config.go
```diff
@@ -473,3 +475,67 @@ func (store *istioConfigStore) Policy(instances []*ServiceInstance, destination
...
+
+				var keyOfAnEgressRuleWithTheSameDomain string
+				keyOfAnEgressRuleWithTheSameDomain, conflictingRule = domains[domainsKey]
+				if conflictingRule {
+					errs = multierror.Append(errs,
+						fmt.Errorf("rule %q conflicts with rule %q on domain "+
+							"%s and port %d, is rejected", egressRuleKey,
+							keyOfAnEgressRuleWithTheSameDomain, domain,
+							portNumber))
+					break DomainsAndPortsLoop
```
rshriram: What are go semantics for labels ? Is it possible to do this without labels ?

model/config.go
```diff
@@ -473,3 +475,67 @@ func (store *istioConfigStore) Policy(instances []*ServiceInstance, destination
...
+				}
+			}
+		}
+
+		if conflictingRule {
+			continue
+		}
```
rshriram: If we break out of the entire top level for loop, why is this needed?

proxy/envoy/config.go
```diff
@@ -664,15 +664,22 @@ func buildEgressFromSidecarVirtualHostOnPort(rule *proxyconfig.EgressRule, mesh
...
 
 func buildEgressFromSidecarHTTPRoutes(mesh *proxyconfig.ProxyMeshConfig, egressRules map[string]*proxyconfig.EgressRule,
 	httpConfigs HTTPRouteConfigs) HTTPRouteConfigs {
-	for _, rule := range egressRules {
+
+	egressRules, errs := model.RejectConflictingEgressRules(egressRules)
+
+	if errs != nil {
+		glog.Warningf("Rejected rules: %v", errs)
+	}
```
rshriram: I don't think we should be doing validation in the config generation phase. Please take a look at validation.go. We validate when users submit rules, so that we can provide immediate feedback. All rule validations need to be done before we enter config generation.

model/config.go
```diff
@@ -473,3 +475,67 @@ func (store *istioConfigStore) Policy(instances []*ServiceInstance, destination
 
 	return out
 }
+
+// RejectConflictingEgressRules rejects rules that have a domain which is equal to
+// a domain of some other rule.
+// According to Envoy's virtual host specification, no virtual hosts can share the same domain.
+// The following code rejects conflicting rules deterministically, by a lexicographical order -
+// a rule with a smaller key lexicographically wins.
+// Here the key of the rule is the key of the Istio configuration objects - see
```
rshriram: Please move this to validation.go.

proxy/envoy/config.go
vadimeisenbergibm: This is the problem - the validation is performed on a single configuration object only. This way you cannot find conflicts between different configuration objects. 

I propose the following - today perform this check in the configuration generation check. Later consider implementing validation of conflicts of all Istio configuration objects, probably on the server-side, using web-hooks. The check during the configuration generation could still be relevant, as a precaution, in case any bugs happen between updating configuration objects and reading them. 

CC @kyessenov .

model/config.go
vadimeisenbergibm: The semantics is to go after the end of the loop:
```
for _, domain := range egressRule.Domains {
			for _, port := range egressRule.Ports {
```

Once we find the conflict, there is no point to continue the both loops. Without the loop, there will be a need for an additional check and an additional break.

```
for _, domain := range egressRule.Domains {
                     if conflictinRule {
                               break
                     }			
                     for _, port := range egressRule.Ports {
                               ...
                               if conflictinRule {
                                    break
                               }
```

model/config.go
vadimeisenbergibm: We break from the loop with the label `DomainsAndPortsLoop`, not from the loop with this condition.

model/config.go
vadimeisenbergibm: Please see my previous comment - I put this function in config.go since it is different from the functions in validation.go where validation of single configuration objects is performed. If you think it still should be in validation.go, I will copy that function and the test to validation.go/validation_test.go.

proxy/envoy/config.go
rshriram: I see. In that case, I think we should normalize the rules (merge) and not reject. This is because, we do not have any actions on the egress rules. Its just a list of domains to pass through from the sidecar. In case there are two egress rules with different settings for use_egress_proxy, just take the first one and ignore the rest. This is the mode we adopt throughout this code. I do not see any downsides to this, especially since there is no action associated with the egress rule.

Also, in your case, you are rejecting an entire rule even if one domain conflicts. If you follow the merge procedure (which is a simple change in the function above), then you would be basically collecting a set of virtualhost:port entries along with the respective setting for use_egress_proxy. This can then be used to generate the virtual host blocks, etc.

What do you think @kyessenov ?

vadimeisenbergibm: I would go with the simple solution. In the documentation, state that egress rules cannot share the same domain with the same port. Period. Then we just remove conflicting rules altogether, instead of implementing complex merge semantics. This way the users are protected from breaking their Istio deployment, but their conflicting rules will be rejected and will appear in the pilot logs as such. Once rejected, it will be more clearly noticeable to the users - they will get 404 for the domains in the rule. If we start partially allowing conflicting rules, it will be harder to notice and to understand what was merged and why.

vadimeisenbergibm: We can go even further and for simplicity require that no two rules can share the same domain, even for different ports. This way one domain can appear in one rule only, and if there is a need to add a port, the containing rule will have to be updated.

rshriram: @frankbu what is your opinion on this?
Egress rule is basically a white list.

frankbu: Since there is only one attribute (use_egress_proxy) that may conflict, it seems that we could get away without validating this for now and simply merge. I think the "safe" merge rule is that if any rule requests use_egress_proxy == true for a domain then use_egress_proxy will be true (OR semantics) for that domain. It seems that use_egress_proxy == false is just an optimization.

If we do that for now, we can always add validation in the future, if this causes problems.

vadimeisenbergibm: @frankbu Please note that also protocols on the port can conflict. For example, you can have three rules on the same domain, that would claim that some port on this domain supports protocols `http`, `http2` or `https`. The question is which protocol to select while merging these three rules into one.

1. Should `https` win over `http` and `http2` ? The outbound cluster will have SSL context so Envoy will perform TLS origination.
2. Should `http2` win over `http`? The outbound cluster will have `http2` feature defined.

rshriram: Lets merge this as it is today.. Once we update egress rule to contain a single domain, this should go away.

rshriram:
merging this - as this is the only PR with all green in pilot!

PR #1176, opened 2017-09-05
in make setup: do not create platform/kube/config if it already exists

vadimeisenbergibm:
Today, running `make setup` twice fails due to an attempt to create platform/kube/config (soft link), while it already exists. 

**Release note**:

```NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 SHELL := /bin/bash
 
 .PHONY: setup
-setup: kubeconfig
+setup: platform/kube/config
 	@bin/install-prereqs.sh
 	@bin/init.sh
 
@@ -51,6 +51,6 @@ gazelle:
 racetest:
 	@bazel test --features=race //...
 
-kubeconfig:
+platform/kube/config:
 	@ln -s ~/.kube/config platform/kube/
 


Makefile
```diff
@@ -15,7 +15,7 @@
 SHELL := /bin/bash
 
 .PHONY: setup
-setup: kubeconfig
+setup: platform/kube/config
```
gyliu513: @vadimeisenbergibm can you please explain more for why does this fix the issue? I have a try and this works.

vadimeisenbergibm: Otherwise, running `make setup` twice fails - the second `ln -s` fails for me, on ubuntu.
My change also will allow the user to set platform/kube/config differently, and `make setup` will not reset it.

vadimeisenbergibm: @gyliu513 regarding why it fixes the issue - make will check if `platform/kube/config` exists, and if it does, make will not perform `ln -s`.

gyliu513: Got it, thanks @vadimeisenbergibm

gyliu513:
@rshriram can you pls help approve?

Makefile
gyliu513: @vadimeisenbergibm you may also want to update the document a bit here:

doc/build.md:To avoid a `missing input file '//platform/kube:config'` error you need to `ln -s ~/.kube/config platform/kube/` or `make kubeconfig`, see [testing](testing.md) for details/why.

vadimeisenbergibm: @gyliu513 yep, I want to rewrite this document to make it more clear.

kadamvandita: where shd make kubeconfig command be run? getting  missing input file '//platform/kube:config'   error

kadamvandita: Regardless of the above change the code fails. From where is the file ~/.kube/config  generated?Do we need to install kubernetes for this?

PR #1178, opened 2017-09-05
rewrite build instructions

vadimeisenbergibm:
The current build instructions are too lengthy, confusing and contain too much information. Make the instructions short.

**Release note**:

```NONE
```

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -1,35 +1,31 @@
-# Build instructions for Linux
+# Build instructions
 
-See also [Istio Development](https://github.com/istio/istio/blob/master/devel/README.md) for setup and general tips.
 
-(For Windows and Mac, we recommend using a Linux virtual machine and/or [Vagrant-specific build instructions](build-vagrant.md) or [minikube build pod](minikube.md); Go code compiles on Mac but docker and proxy tests will fail on Mac)
 
-We are using [Bazel 0.5.2](https://github.com/bazelbuild/bazel/releases) as the main build system in Istio Pilot. The following command builds all targets in Istio Pilot:
+## Environment
+We build on Linux. For Windows and Mac, we recommend using a Linux virtual machine and/or [Vagrant-specific build instructions](build-vagrant.md) or [minikube build pod](minikube.md). Go code compiles on Mac but docker and proxy tests will fail.
 
-To avoid a `missing input file '//platform/kube:config'` error you need to `ln -s ~/.kube/config platform/kube/` or `make kubeconfig`, see [testing](testing.md) for details/why.
+## Prerequisites
 
-    make build
+* [Setting up Go](https://github.com/istio/istio/blob/master/devel/README.md#setting-up-go)
+* [Setting up Bazel](https://github.com/istio/istio/blob/master/devel/README.md#setting-up-bazel)
 
-Bazel uses `BUILD` files to keep track of dependencies between sources.  If you
-add a new source file or change the imports  please run the following command
-in the repository root to update all `BUILD` files:
+## Initial Setup
 
-    make gazelle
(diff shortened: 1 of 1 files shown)


doc/build.md
```diff
@@ -1,35 +1,31 @@
-# Build instructions for Linux
+# Build instructions
```
kyessenov: These instructions may work for Mac, but need to be validated. Bazel gained native support for Mac recently.

doc/build.md
```diff
@@ -1,35 +1,31 @@
...
 
-    make gazelle
-
-Data dependencies such as the ones used by tests require manual declaration in
-the `BUILD` files.
+## Initial Setup
 
-## Go tooling compatibility
+1. [Create a forked repository](https://github.com/istio/istio/blob/master/devel/README.md#fork-the-main-repository) in `$GOPATH/src/istio.io`.
+2. Run `make setup`. It well install the required tools and will [vendorize](https://golang.org/cmd/go/#hdr-Vendor_Directories) the dependencies.
```
kyessenov: Please specify the exact version of Bazel. Project will not build with versions prior to 0.5.2.

kyessenov: The magic involved in linking files and auto-vendorizing tends to be brittle and not guaranteed to work. I'm afraid by hiding both bazel and the brittle script, we're opening the doors to lots of questions about why the build fails.

doc/build.md
vadimeisenbergibm: This is stated in the Environment section. Putting "for Linux" in the heading seems like repetition for me.

doc/build.md
```diff
@@ -1,35 +1,31 @@
-# Build instructions for Linux
+# Build instructions
 
-See also [Istio Development](https://github.com/istio/istio/blob/master/devel/README.md) for setup and general tips.
```
ldemailly: I wonder if there is a way to improve the general document rather than diverge between istio repos

I would in theory remove this doc entirely in favor of the the shared doc ?

rshriram:
Given all the bazel madness, my opinion is to have a lower bar to entry and then let people warm up to bazel complexities, rather than scaring them away immediately.

kyessenov:
Don't remove the complexity though :) This is the place of record for which version of bazel runs this, and the right sequence of steps. I'm ok moving the instructions somewhere else, but we can't just remove them.

doc/build.md
vadimeisenbergibm: Added mentioning 0.5.2 version. So what are the sources of brittleness? What can be improved? If Bazel and the script are brittle, the questions will be asked anyway.

Please note that Makefile is already used in mixer and is mentioned in [Developing for Istio guide](https://github.com/istio/istio/tree/master/devel). This PR only tries to improve the existing instructions, which used Makefile before.

vadimeisenbergibm: Regarding dual/triple build systems - it seems this is the situation in mixer as well, and also in Kubernetes itself.
You have:
1. Bazel BUILD files
2. Scripts
3. Makefile

Makefile seems to be used to run scripts and to invoke Bazel. I do not know why this situation happened. My guess is that different tools are better at different tasks.

doc/build.md
vadimeisenbergibm: @ldemailly @rshriram @kyessenov Please see the shortened version of the build instructions. I just referenced the general document and stated the differences of Pilot.

What I learned - there is duplication of build artifacts between pilot and mixer. Both have Makefiles which are slightly different and both have `bin/bazel_to_go.py` which are different as well. I would propose to consider creating an `istio/build` repositiory and putting Makefile and scripts there. Then this repository can be shared between Istio repositories as git submodule or in some other way, so all the Istio repositories will use the same Makefiles and scripts, and the development guide will be the same for all.

rshriram: and we could just build off the istio/build ;)

ldemailly: https://github.com/istio/test-infra/issues/91

PR #1182, opened 2017-09-06
Rename gobuild script to gocompile to avoid bazelify horror

andraxylia:
Bazelify wrongly belived a file called gobuild is a bazel build file.

### DIFF
--- a/bin/gocompile-and-push-images
+++ b/bin/gocompile-and-push-images
@@ -17,7 +17,7 @@
 ################################################################################
 #
 # This script creates a base image for the init image that includes iptables.
-#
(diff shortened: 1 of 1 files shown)


kyessenov:
please add `.sh` suffix to avoid problems like this.

ldemailly:
it passed the step that was broken before so please merge - ty

kyessenov:
The PR content does not match description.

ldemailly:
for the record it's not buildifier but our check.sh as @ayj mentioned/fixed in https://github.com/istio/pilot/pull/1158/files#diff-2ea24c113876f7ba80e101dd5e3e77ca

PR #1184, opened 2017-09-06
Create istio.deps to activate dependency update

chxchx:
This file defines a list of repos as dependencies of the parent repo. It only concerns about repos under the istio project umbrella. Automated dependency update uses this file to learn what dependencies are and pull their latest versions to compare with what has been used on master of parent repo. It might be useful to take a look at https://github.com/istio/test-infra/blob/master/toolbox/util/dependency.go

**Release note**:

```release-note
None
```


istio.deps
```diff
@@ -0,0 +1,9 @@
+[
+	{
+		"name": "ISTIO_PROXY_BUCKET",
+		"repoName": "proxy",
+		"prodBranch": "stable",
+		"file": "WORKSPACE",
+		"lastStableSHA": "62755c63d9c80c043ece26504c4288845a929cc4"
+	}
+]
```
rshriram: This is not enough.. We need to update the SHAs on docker images as well.

chxchx:
PTAL
One question I have is that in https://github.com/istio/pilot/blob/master/test/integration/driver.go#L57 we have SHAs from auth and mixer, but it says update manually at the moment. Do we want to auto update those as well?

kyessenov:
No need to update those. Mixer needs to be replaced with a fake mixer but don't have hands to do that. CA is rather simple at that SHA. 

Where's the script that does the update? Can that script be hosted here, so we know what's happening?

chxchx:
The script to update dependency is at https://github.com/istio/test-infra/blob/master/toolbox/deps_update/main.go

I think hosting it here is not ideal.
* It creates duplication. Any changes to the original script must also be copied over to all duplicates.
* The script uses a lot of shared utility in test-infra such as GitHub API client, which means they have to be copied over as well.
* The script is meant to be generic. It can update the dependencies of any repo under the istio umbrella by simply changing the flag.

kyessenov:
Would be nice to have backreference to the script. Maybe make it YAML so we can put comments into it?

chxchx:
I include a comment with reference to the update script in the latest commit. Since we have enabled auto dependency update on other repos (each has a `istio.deps` file like this one), it is hard to switch to yaml at the moment. If you strongly believe yaml is perferred in this case, I encourage that you file an github issue with the label `asks-to-engprod`.

An example PR created by the script to update dependencies of pilot can be found here
https://github.com/chxchx/pilot/pull/5/files

chxchx:
@ldemailly This helps to get proxy into pilot as requested. Would love to have your input.

ldemailly:
you could add some comments but 
/lgtm
we need to get the bot in place asap so let's get this in

PR #1187, opened 2017-09-06
disable generate request ID when tracing is disabled

rshriram:
Do not enable request ID generation if Tracing is disabled. Saves a lot of CPU.

**Release note**:

```
NONE
```


kyessenov:
Request ID has value independent of tracing. Maybe have a separate flag?

rshriram:
per offline conversation, X-Request-ID is generated as long as tracing is enabled - which is the case in our tests. For prod/perf cases, this should be disabled as its a big CPU hog.

PR #1195, opened 2017-09-07
Increase default ResyncPeriod from 1 to 60 seconds

ayj:
Pilot currently uses 1 second resync period which puts additional load
on apiserver. This is not necessary and doesn't scale well in shared
test clusters. Increase to 60 seconds for now with the possibility of
further increases if this doesn't cause test instability.

See
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-sig-api-machinery/PbSCXdLDno0/v9gH3HXVDAAJ
for recommendations on setting resync period.

**Release note**:

```release-note
```

### DIFF
--- a/cmd/pilot-discovery/main.go
+++ b/cmd/pilot-discovery/main.go
@@ -184,7 +184,7 @@ func init() {
 	discoveryCmd.PersistentFlags().StringVarP(&flags.controllerOptions.WatchedNamespace, "app namespace",
 		"a", metav1.NamespaceAll,
 		"Restrict the applications namespace the controller manages; if not set, controller watches all namespaces")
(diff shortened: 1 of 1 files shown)


ldemailly:
thx for the explanation kuat- then why isn't it hours ? why is a full resync ever needed if the connection/subscription doesn't error out ?

kyessenov:
It doesn't hurt either. We will gradually ramp it up. Jumping 1s to 1d is too much.

ldemailly:
any chance that was what triggered some of the 503s / the race you mentioned ?
(which then will become harder to detect but still present?)

kyessenov:
Yes, RDS/CDS async is a known issue solved by ADS that's being implemented. It results in 404 I think, not 503.

Issue #964, opened 2017-07-27
Add direct network endpoint routes for TCP headless services


kyessenov:
Rich client libraries assume direct connectivity (e.g. kafka libraries).

PR #1197, opened 2017-09-07
TCP k8s headless service support

ijsnellf:
Adds support for k8s TCP [headless services](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services). Also enables discovery of and routing to headless HTTP/HTTP2 services.

**Which issue this PR fixes**: fixes https://github.com/istio/pilot/issues/964 

Additional work needed for other protocols (HTTP/gRPC etc). Currently HTTP/HTTP2 requests are routed / load balanced based on their host headers. If this behavior needs to be modified, it can be addressed in follow-on PRs.

**Release note**:

```release-note
Fixed bug that prevented istio services from talking to headless services in kubernetes
```


rshriram:
this tackles just TCP right? Lets get this PR passing and ship it. Do gRPC in another PR unless its an easy fix.

proxy/envoy/config.go
```diff
@@ -524,25 +524,47 @@ func buildOutboundHTTPRoutes(mesh *proxyconfig.ProxyMeshConfig, sidecar proxy.No
...
+				if service.Address != "" {
+					cluster := buildOutboundCluster(service.Hostname, servicePort, nil)
+					route := buildTCPRoute(cluster, []string{service.Address})
+					config := &TCPRouteConfig{Routes: []*TCPRoute{route}}
+					listener := buildTCPListener(
+						config, service.Address, servicePort.Port, servicePort.Protocol)
+					tcpClusters = append(tcpClusters, cluster)
+					tcpListeners = append(tcpListeners, listener)
+				} else {
+					if originalDstCluster == nil {
```
kyessenov: the order of Services is non-det and this seems to pick the first one that has no address. Can we change the logic so that we pick one original destination cluster for each servicePort.Port (by sorting services by name)?
We also need to be careful with merging these listeners with HTTP listeners as they can collide.

test/integration/grpc.go
```diff
@@ -48,14 +48,17 @@ func (t *grpc) run() error {
 }
 
 func (t *grpc) makeRequests() error {
-	testPods := []string{"a", "b"}
+	srcPods := []string{"a", "b"}
+	dstPods := []string{"a", "b"}
 	if t.Auth == proxyconfig.ProxyMeshConfig_NONE {
 		// t is not behind proxy, so it cannot talk in Istio auth.
-		testPods = append(testPods, "t")
+		srcPods = append(srcPods, "t")
```
kyessenov: add "t" to dstPods too?

kyessenov: or maybe just add "headless" to "testPods"?

test/integration/headless.go
```diff
@@ -0,0 +1,71 @@
...
+
+	proxyconfig "istio.io/api/proxy/v1/config"
+)
+
+type headless struct {
+	*infra
+}
+
+func (t *headless) String() string {
+	return "headless-reachability"
```
kyessenov: headless-tcp-reachability

test/integration/headless.go
```diff
@@ -0,0 +1,71 @@
...
+	for _, src := range srcPods {
+		for _, dst := range dstPods {
+			for _, port := range []string{":10090", ":19090"} {
+				for _, domain := range []string{"", "." + t.Namespace} {
+					name := fmt.Sprintf("TCP connection from %s to %s%s%s", src, dst, domain, port)
+					funcs[name] = (func(src, dst, port, domain string) func() status {
+						url := fmt.Sprintf("http://%s%s%s/%s", dst, domain, port, src)
+						return func() status {
+							resp := t.clientRequest(src, url, 1, "")
+							if t.Auth == proxyconfig.ProxyMeshConfig_MUTUAL_TLS && src == "t" {
```
kyessenov: skip mTLS check as there was one prior?

test/integration/grpc.go
ijsnellf: We can't use "headless" as a source (unless we explicitly remap it to the b pod).

ijsnellf:
@kyessenov: I added explicit checks for TCP wildcard listener duplicates and made TCP listener ordering deterministic. Since TCP listener ordering is deterministic, TCP/HTTP wildcard listener conflicts should be de-dupped by `listeners.normalize()`, but that will be done silently (no warning messages).

PR #1202, opened 2017-09-08
DebugMode for kube-inject

ldemailly:
Fixes issue #1029
```release-note
Use istioctl kube-inject —debug=false to get production proxy image
Default is true, which gives image you can sudo apt-get as needed
```


platform/kube/inject/http_test.go
```diff
@@ -32,20 +32,21 @@ import (
 )
 
 var (
-	mesh = proxy.DefaultMeshConfig()
-
-	httpTestConfig = &Config{
+	mesh              = proxy.DefaultMeshConfig()
+	unitTestDebugMode = true
```
ayj: Can you put this with unitTestHub/unitTestTag in platform/kube/inject/inject_test.go?

platform/kube/inject/inject.go
```diff
@@ -148,6 +155,7 @@ type Params struct {
 	SidecarProxyUID   int64                        `json:"sidecarProxyUID"`
 	Version           string                       `json:"version"`
 	EnableCoreDump    bool                         `json:"enableCoreDump"`
+	DebugMode         bool                         `json:"debugMode"`
```
ldemailly: linter made me put here for alignment; saves some memory

platform/kube/inject/inject.go
```diff
@@ -200,10 +208,10 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, name string) (*C
 		c.Policy = DefaultInjectionPolicy
 	}
 	if c.Params.InitImage == "" {
-		c.Params.InitImage = InitImageName(DefaultHub, version.Info.Version)
+		c.Params.InitImage = InitImageName(DefaultHub, version.Info.Version, c.Params.DebugMode)
```
ldemailly: having both InitImage and the debug mode depending on a flag which should agree when invoked is weird.
I started putting XxxImageName as method on Param (SetProxyImageName() which would use the flag but that doesn't work well with initialization of the struct and the rest of the code, an alternative would be to pass in the hub and version and flag instead of the image name but that's a bigger change)

platform/kube/inject/http_test.go
ldemailly: it's not really a constant like the others (I have it set to true in some tests and false in others but sure... not convinced a gloabal for true is useful (like we have `privTrue := true` somewhere, should be `priv := true` or just `true`)
but will change

ldemailly:
@ayj one interesting question is if you want to support that switch from the auto inject and how ?

test/integration/infra.go
```diff
@@ -134,19 +134,20 @@ func (infra *infra) setup() error {
 	if err != nil {
 		return err
 	}
-
+	debug := true
```
ayj: Add this to `infra` and expose as a flag in driver.go?

platform/kube/inject/http_test.go
ayj: I see what you mean. It looks like `unitTestDebugMode` is only used once in http_test.go? Maybe just remove it altogether.

test/integration/infra.go
ldemailly: EnableCoreDump, Version, MeshConfigMapName don't seem to have flags ? is it really necessary for this one?
but let me look

ayj:
> @ayj one interesting question is if you want to support that flag from the auto inject and how ?

`readOnly` and `privileged` will be set correctly w/auto-inject. The proxy/init image is set explicitly in the ConfigMap so debugMode won't have any effect on that. For consistency it might be better to add explicit `--proxyImage` and `initImage` flags to kube-inject and have debugMode only affect the privileges, logging, etc.

test/integration/infra.go
ayj: `DebugMode` is something people might want to set for tests. `Version` and `MeshConfigMapName` shouldn't need to changed for e2e tests.

ldemailly: fixed

platform/kube/inject/http_test.go
ldemailly: too late, fixed :-) the reason it's in a variable is it is used 3 times (for 2 images and the mode) I wouldn't want to put a literal and have it diverge where only one of the 3 is one way...

ldemailly:
> readOnly and privileged will be set correctly w/auto-inject.
what's the definition of correctly ? debug image needs non readonly and priviledge to be actually useful for debugging which is what this pr is about

platform/kube/inject/inject.go
```diff
@@ -132,12 +132,19 @@ const (
 )
 
 // InitImageName returns the fully qualified image name for the istio
-// init image given a docker hub and tag
-func InitImageName(hub, tag string) string { return hub + "/proxy_init:" + tag }
+// init image given a docker hub and tag and debug flag
```
gyliu513: Why add a debug flag here? Seem it is not used

platform/kube/inject/inject.go
```diff
@@ -132,12 +132,19 @@ const (
...
+func InitImageName(hub string, tag string, _ bool) string {
+	return hub + "/proxy_init:" + tag
+}
 
 // ProxyImageName returns the fully qualified image name for the istio
-// proxy image given a docker hub and tag.
-func ProxyImageName(hub, tag string) string { return hub + "/proxy_debug:" + tag }
+// proxy image given a docker hub and tag and whether to use debug or not.
```
gyliu513: s/proxy image given a docker hub and tag and whether to use debug or not./proxy image given a docker hub, tag and whether to use debug or not.

gyliu513:
Also how about put the content on top to release note?

```
Use istioctl kube-inject —debug=false to get production proxy image
Default is true, which gives image you can sudo apt-get as needed
```

ldemailly:
> Also how about put the content on top to release note? @ldemailly

I don't understand ? what content on top ? those are the release notes...

platform/kube/inject/inject.go
ldemailly: because eventually we will have debug/opt for all images we produce

platform/kube/inject/inject.go
ldemailly: thanks ! but sorry already merged, next pr/time; I think it's readable enough/I was minimizing the diff so not critical to fix/followup for "and" -> ","

ldemailly:
> https://github.com/istio/pilot/blob/master/test/integration/testdata/initializer-configmap.yaml.tmpl
ah thanks for this one ! didn't know/was asking jason about it.
I think we'll have another doc/pr for what to do to the config map yes

gyliu513:
@ldemailly I see the release note, I was originally thinking that only in such format can a release note be added.

image

If we can also get release note without the keyword of `Release note:`, that would be fine.

ldemailly:
yes the bot looks for ```releasenote
blocks

gyliu513:
>yes the bot looks for ```release-note
blocks

I see, thanks @ldemailly

PR #1205, opened 2017-09-08
Separate proxy-specific configs from mesh-wide configs

kyessenov:
This PR splits statically configured settings in agent from dynamically configured settings in discovery. This would allow auth and mixer to be turned on/off dynamically.

**Release note**:

```release-note
```


adapter/config/crd/conversion.go
```diff
@@ -45,7 +45,7 @@ func convertObject(schema model.ProtoSchema, object IstioObject, domain string)
 
 // convertConfig translates Istio config to k8s config JSON
 func convertConfig(schema model.ProtoSchema, config model.Config) (IstioObject, error) {
-	spec, err := schema.ToJSONMap(config.Spec)
+	spec, err := model.ToJSONMap(config.Spec)
```
andraxylia: Renaming/refactoring should be done in a separate PR with no functional changes.

andraxylia:
Do we need this functionality in 0.2?

kyessenov:
Bug fix for inability to turn off some settings live.

model/validation.go
```diff
@@ -1042,66 +1042,95 @@ func ValidateConnectTimeout(timeout *duration.Duration) error {
...
 
-	// discovery address is mandatory since mutual TLS relies on CDS.
-	// strictly speaking, proxies can operate without RDS/CDS and with hot restarts
-	// but that requires additional test validation
-	if mesh.DiscoveryAddress == "" {
-		errs = multierror.Append(errs, errors.New("discovery address must be set to the proxy discovery service"))
-	} else if err := ValidateProxyAddress(mesh.DiscoveryAddress); err != nil {
-		errs = multierror.Append(errs, multierror.Prefix(err, "invalid discovery address:"))
-	}
-
```
andraxylia: Why is this check removed?

andraxylia: Nevermind, it has been moved below.

test/integration/testdata/config.yaml.tmpl
```diff
@@ -4,13 +4,16 @@ metadata:
   name: istio
 data:
   mesh: |-
-    discoveryAddress: istio-pilot.{{.IstioNamespace}}:8080
+    defaultConfig:
```
andraxylia: What is defaultConfig about and why is this needed? Once it is a configMap, it can be overridden.

test/integration/testdata/egress-proxy.yaml.tmpl
```diff
@@ -48,18 +58,12 @@ spec:
               apiVersion: v1
               fieldPath: metadata.namespace
         volumeMounts:
-        - name: config-volume
```
andraxylia: Won't this conflict kube-inject initializer and istioctl kube-inject, who still rely on the configMap? @ayj

andraxylia:
Can you elaborate on the settings and the functionality this provides?

cmd/pilot-agent/main.go
```diff
@@ -99,12 +83,46 @@ var (
...
+				{
+					Directory: proxy.AuthCertsPath,
+					Files:     []string{proxy.CertChainFilename, proxy.KeyFilename, proxy.RootCertFilename},
+				},
+			}
+
+			if role.Type == proxy.Ingress {
+				certs = append(certs, envoy.CertSource{
+					Directory: proxy.IngressCertsPath,
+					Files:     []string{"tls.crt", "tls.key"},
```
ayj: `tls.crt` and `tls.key` should also be defined in `proxy` package for consistency with the sidecar key/certs?

cmd/pilot-agent/main.go
```diff
@@ -99,12 +83,46 @@ var (
...
+				})
+			}
+
+			glog.V(2).Infof("monitored certs: %#v", certs)
```
ayj: nit: s/monitored/Monitored

model/validation.go
```diff
@@ -1042,66 +1042,95 @@ func ValidateConnectTimeout(timeout *duration.Duration) error {
 }
 
 // ValidateProxyMeshConfig checks that the mesh config is well-formed
-func ValidateProxyMeshConfig(mesh *proxyconfig.ProxyMeshConfig) (errs error) {
+func ValidateProxyMeshConfig(mesh *proxyconfig.MeshConfig) (errs error) {
```
ayj: nit: s/ValidateProxyMeshConfig/ValidateMeshConfig for consistency with `ValidateProxyConfig()` below and the input to this function is `MeshConfig`.

cmd/pilot-agent/main.go
```diff
@@ -118,24 +136,18 @@ var (
 )
 
 func init() {
+	proxyCmd.PersistentFlags().StringVar(&config, "config", "",
+		"Configuration overrides for the proxy as YAML/JSON")
 	proxyCmd.PersistentFlags().StringVar((*string)(&serviceregistry), "serviceregistry",
 		string(platform.KubernetesRegistry),
 		fmt.Sprintf("Select the platform for service registry, options are {%s, %s}",
 			string(platform.KubernetesRegistry), string(platform.ConsulRegistry)))
-	proxyCmd.PersistentFlags().StringVar(&meshconfig, "meshconfig", "/etc/istio/config/mesh",
```
costinm: Backward compat please. Easy to copy old flags to new.

cmd/pilot-agent/main.go
```diff
@@ -118,24 +136,18 @@ var (
 )
 
 func init() {
+	proxyCmd.PersistentFlags().StringVar(&config, "config", "",
```
ayj: Can you update description to indicate the value of `config` is the YAML/JSON content itself and not a filename? It wasn't obvious to me until I reviewed the initializer code that uses it.

proxy/context.go
```diff
@@ -102,27 +102,51 @@ const (
...
+	KeyFilename = "key.pem"
+
+	// RootCertFilename is mTLS root cert
+	RootCertFilename = "root-cert.pem"
+)
+
+// DefaultProxyConfig for individual proxies
+func DefaultProxyConfig() proxyconfig.ProxyConfig {
+	return proxyconfig.ProxyConfig{
+		ConfigPath:             "/etc/istio/proxy",
```
costinm: /var/lib/istio/...

proxy/envoy/route.go
```diff
@@ -40,9 +41,9 @@ const (
 // buildListenerSSLContext returns an SSLContext struct.
 func buildListenerSSLContext(certsDir string) *SSLContext {
 	return &SSLContext{
-		CertChainFile:            certsDir + "/" + certChainFilename,
-		PrivateKeyFile:           certsDir + "/" + keyFilename,
-		CaCertFile:               certsDir + "/" + rootCertFilename,
+		CertChainFile:            certsDir + proxy.CertChainFilename,
```
costinm: Would be good to be defensive - have a function to add "/" to certsDir if missing, or similar. 
In particular since before the "/" at the end of the dir was not required ( and it's usually not required in most config formats)

test/integration/testdata/egress-proxy.yaml.tmpl
```diff
@@ -35,7 +35,17 @@ spec:
...
         ports:
         - containerPort: 80
-        args: ["proxy", "egress", "-v", "{{.Verbosity}}"]
+        args:
+        - proxy
+        - egress
+        - -v
+        - "{{.Verbosity}}"
+        - --config
+        - |
```
costinm: Not familiar with this - does it generate a temp file ?

cmd/pilot-agent/main.go
```diff
@@ -99,12 +83,46 @@ var (
 
 			glog.V(2).Infof("proxy role: %#v", role)
 
-			envoyProxy := envoy.MakeProxy(mesh, servicecluster, role.ServiceNode(), configpath, binarypath)
-			agent := proxy.NewAgent(envoyProxy, proxy.DefaultRetry)
-			watcher, err := envoy.NewWatcher(mesh, agent, role)
-			if err != nil {
+			proxyConfig := proxy.DefaultProxyConfig()
+			if config != "" {
+				if err := model.ApplyYAML(config, &proxyConfig); err != nil {
```
costinm: Do you mean the yaml file will be passed as a string in the config ???
That doesn't make any sense - it'll be quite bad on a VM ( there are also line limits on command, etc )...

test/integration/testdata/egress-proxy.yaml.tmpl
ayj: The egress/ingress proxy are exempt from injection (see ` policy.sidecar.istio.io: "policy.sidecar.istio.io/force-off"`)

WORKSPACE
```diff
@@ -478,8 +478,8 @@ go_proto_library(
     ],
 )
     """,
-    commit = "88dd261709ccd8fa8a43bec5e3c15071d39bfc7c",  # Sep 07, 2017
-    remote = "https://github.com/istio/api.git",
+    commit = "d68fa2d4cd9b022a1778401a43259a711c914530",  # Sep 07, 2017
+    remote = "https://github.com/kyessenov/api.git",
```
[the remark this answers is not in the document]
kyessenov: Yes, this passes locally as-is before that PR gets merged.

cmd/pilot-agent/main.go
kyessenov: The problem is that admission controller needs to configure the agent. We could define 20 flags and then set the values for each in the pod template. But that's error prone. This suggests to pass a proto inline in YAML form, so we can  be backwards compatible, and have stronger validation. It would look like: `./agent --config "discoveryAddress: pilot:8080"`

cmd/pilot-agent/main.go
kyessenov: meshconfig is being deprecated and into two resources...

proxy/context.go
kyessenov: Let's not change values. I specifically wanted to maintain the same golden test outputs to avoid breaking anything.

proxy/envoy/route.go
kyessenov: Yes, but that is a good small task to do in another PR. This PR is already big and not changing the logic much.

test/integration/testdata/egress-proxy.yaml.tmpl
kyessenov: Yes, as I mentioned this is how it looks when generated by kube-inject or admission controller. Ingress is really a dummy pod injected with a proxy.

cmd/pilot-agent/main.go
kyessenov: Ok, will fix.

cmd/pilot-agent/main.go
kyessenov: Ok, I am trained to use lower case for messages so they can be merged correctly.

cmd/pilot-agent/main.go
kyessenov: Yes, will fix.

model/validation.go
kyessenov: Yes, will fix.

kyessenov:
Thanks for the review. Let me address comments.

cmd/pilot-agent/main.go
rshriram: this is not a convenient thing to launch the agent with an entire YAML as a string. Can you also add a flag to read from a file? We don't have to use it but it will become useful in non pod scenarios, where I just want to run the agent with a bunch of settings.

cmd/pilot-agent/main.go
```diff
@@ -118,24 +136,18 @@ var (
 )
 
 func init() {
+	proxyCmd.PersistentFlags().StringVar(&config, "config", "",
+		"Configuration overrides for the proxy as YAML/JSON")
```
rshriram: Rename this to configValues and add a config-file option. So that we can do the fugly thing behind the scenes but have a nicer option for humans

proxy/context.go
rshriram: Lets not change please.

proxy/context.go
```diff
@@ -102,27 +102,51 @@ const (
...
+		ConnectTimeout:        ptypes.DurationProto(1 * time.Second),
 		IngressClass:          "istio",
-		IngressControllerMode: proxyconfig.ProxyMeshConfig_STRICT,
-
-		AuthPolicy:    proxyconfig.ProxyMeshConfig_NONE,
-		AuthCertsPath: "/etc/certs",
+		IngressControllerMode: proxyconfig.MeshConfig_STRICT,
+		AuthPolicy:            proxyconfig.MeshConfig_NONE,
+		RdsRefreshDelay:       ptypes.DurationProto(1 * time.Second),
+		ZipkinAddress:         "",
```
rshriram: convert to boolean

cmd/pilot-agent/main.go
kyessenov: Done

cmd/pilot-agent/main.go
kyessenov: done

proxy/envoy/route.go
kyessenov: added `path.Join` which does the right thing

cmd/pilot-agent/main.go
kyessenov: renamed to `configInline`

model/validation.go
kyessenov: Done

cmd/pilot-agent/main.go
kyessenov: renamed to `configInline` and adding a flag for `configFile` to read from a file.

proxy/context.go
kyessenov: Done

kyessenov:
I strongly suggest we don't change default values in this same PR. That is much easier to review in a separate PR.

cmd/pilot-agent/main.go
costinm: My point was that passing a yaml file as an 'inline' command line parameter is not how 
either CLI or yaml were meant to be used.

kyessenov: Please suggest a way without enumerating all proto fields and creating flags for each.

ayj: Any better if we pass JSON? That's a common pattern with other command line tools, e.g. 

```bash
curl -H "Content-Type: application/json" -X POST -d '{"username":"xyz","password":"xyz"}' http://localhost:3000/api/login
```

costinm: And as a general comment - using a proto and yaml to pass configuration seems super complex given that in practice we have a dozen or so settings that could be cleanly expressed as flags or env ( both supported by cobra ). 

Just because someday we may have nested config doesn't require both proto and yaml - the flags have self-included doc and defaults, are well understood, etc. 

I'm not aware of any benefit of having the proto in the api directory that justifies the added compexity.

kyessenov: YAML is a superset of JSON. Flags are in-code, proto is a declarative spec. We generated doc pages from the proto, can't be done that easily for flags. Constraints are also easier to express over proto than flags. Anyways, since we do processing of the values, it's better to treat them as first class API objects, rather than a bunch of strings in some flags.

costinm: I'm not sure what is the problem with having one field/flag for each field - it's a dozen or so - 
having them in the .proto or in the code is pretty much the same amount of code. Yaml may be
best for complex configs ( not our case ), proto is great for network transport - but it seems in
this case flags would be a much better user experience. Even k8s allow pretty much all operations to be done using flags.

kyessenov: Kubernetes is also trying to switch to configmaps whenever possible to let them change dynamically. We're in a better position here with not having a bazillion flags.

cmd/pilot-agent/main.go
rshriram: Thanks!

kyessenov:
@costinm is there something you want me to address here?

andraxylia:
Please update branch.

kyessenov:
Ugh, I miss the old days of functioning test infra...

cmd/pilot-agent/main.go
```diff
@@ -97,14 +84,62 @@ var (
...
-			glog.V(2).Infof("proxy role: %#v", role)
+			glog.V(2).Infof("Proxy role: %#v", role)
 
-			envoyProxy := envoy.MakeProxy(mesh, servicecluster, role.ServiceNode(), configpath, binarypath)
-			agent := proxy.NewAgent(envoyProxy, proxy.DefaultRetry)
-			watcher, err := envoy.NewWatcher(mesh, agent, role)
-			if err != nil {
+			proxyConfig := proxy.DefaultProxyConfig()
+			if config != "" && configFile != "" {
+				return errors.New("cannot specify both configuration content and file at the same time")
```
costinm: I still don't think passing a yaml file as command line parameter is reasonable - and I'm not aware of any other app doing this. 

--config "foo: bar\n  blah: | foo\n" etc is not readable, doesn't look good when you do "ps", and it is completely 
unneeded and unexpected for a user. We use yaml config files everywhere else.

Please remove this.

kyessenov: Can we treat this as an implementation detail for Admission Controller? I agree it looks awful, but that's the easier way to pass config from AC to pods. Otherwise, we have to allocate space to write files and then distribute those files to each pod.

kyessenov: I added configFile for the normal case

ayj: s/YAML/JSON and the readability issue with newlines goes away? Its functionally equivalent albeit still a bit verbose.

costinm: BTW - there is also viper ( same repo as cobra ) - that adds support for watching/reloading configs, env, and a bunch of formats - and works well with cobra flags.

kyessenov: Files are difficult in the container world. Pod template is the best place to put these configs, and embedded YAML looks pretty nice actually.

kyessenov: @costinm what is your constructive solution that works for k8s AC? let's not block PRs with generic criticisms.

rshriram: Just change to flags? Keep the config file option around.

kyessenov: Ok, that can be done later, right. I'm trying to get to the state that mesh config is dynamic because that's a functional improvement. The switch from inline config from flags is a mechanical conversion process. I don't believe agent is part of our public API, it's an implementation detail of how Pilot operates.

andraxylia:
If something is changed in a configMap, a process can watch the configMap, process & apply the change without restart.

If the same info is passed as flags when the process is started, any change would require a restart. This is not desirable.

kyessenov:
We don't watch files. Once you implement the machinery to push and garbage collect config maps, we can use that.

kyessenov:
Here's the dozen flags. PTAL

rshriram:
please merge after tests pass

cmd/pilot-agent/main.go
costinm: No problem with deprecating - as long as it's kept around for a while before it is removed. As is common with any other 
project exposing flags or configs or APIs ( and required in most google3 projects).

platform/kube/inject/testdata/auth.cert-dir.yaml.injected
```diff
@@ -30,6 +30,28 @@ spec:
         - sidecar
         - -v
         - "2"
+        - --configPath
```
costinm: What happened with the default values ? We are past freeze date, plenty of engineers testing and having different configs ( including the deb for raw VM ) - there is no valid reason for requiring any of those params.

This is not the first program to not be used primarily by end users - the common practice is to have sensible defaults.

kyessenov: It's the result of admission controller processing the flags. We can't be sure agent and discovery on defaults, so it's better to be explicit.

costinm:
Pilot agent is used by raw VM as well, not only admission controller.
Please add back the default values in separate PR.

andraxylia:
The use case and design for this should have been reviewed and approved. Please add more details in the description so one can understand the use case, and how can turn on or off authentication and mixer.

With the limited description provided, my understanding is this PR replaces the configMap with configs sent as individual arguments in the command line for starting the process. ConfigMaps are an elegant way of creating config for a process, that do not require restart. This change will cause traffic loss.

In addition, it is error prone, since for every new field changed in the config, one needs to change in a few other places. Not all the fields are now there, for instance, the EgressProxyAddress is missing. This will break cluster wide. Check that all fields in the config are passed, or if not, document why they do not need to be passed. If more configs are added in the future, it will go beyond the breaking point of the command line.

Please manually test istioctl kube-inject, since the pilot e2e tests do not cover the full command, just one of the internal APIs.

kyessenov:
@costinm Pilot agent has the same default values as before if you do not pass the flags: 
https://github.com/istio/pilot/blob/master/cmd/pilot-agent/main.go#L187
The complete list flags in admission controller is due to proto deserialization artifact (inability to distinguish missing vs empty after we read the YAML).

kyessenov:
@andraxylia 
ConfigMaps are not watched and cannot be simultaneously changes across namespaces. If the proxy agent resides in one namespace, and discovery in another, we need to update both to turn on authentication flag. This is a better solution since it makes the dynamic flags a singleton config unit, while the static flags remain per-proxy and immutable (like before).

kyessenov:
@andraxylia 
I'm not sure what you mean by traffic loss. Currently, to turn on auth, you have to reinstall the entire Istio.

Issue #1164, opened 2017-09-01
Simplify initializer injection policy


ayj:
### current
```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  annotations:
    policy.sidecar.istio.io: policy.sidecar.istio.io/force-on
  name: my-deployment 
# etc ... 
```
### proposed
```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  annotations:
    sidecar.istio.io/inject-override: true
  name: my-deployment 
# etc ... 
```

`sidecar.istio.io/inject-override` becomes a tristate boolean with the following semantics. 

```yaml
# Override default global policy and inject sidecar. Has no affect if global policy is `opt-out`.
sidecar.istio.io/inject-override: true

# Override default global policy and do not inject sidecar. Has no affect if global policy is `opt-in`.
sidecar.istio.io/inject-override: false

# No annotation (i.e. missing key) or unknown value results in default behavior.
```

frankbu:
This looks good to me, but maybe it would be a little better to name it `sidecar.istio.io/inject` since true means inject, and false means don't inject. The name `sidecar.istio.io/inject-override` could be interpreted to mean "override the specified cluster injection policy", yes (true) or no (false).

ayj:
> The name sidecar.istio.io/inject-override could be interpreted to mean "override the specified cluster injection policy", yes (true) or no (false).

That is what it means :) It overrides the default injection policy of the cluster. This was my original intent for using `force-on`/`force-off` instead of a boolean - to make the intent explicit.

frankbu:
I understand, but my point is that there are two possible interpretations of the values true and false.

Option 1
--------
true => do inject when the default policy is to not inject
false => don't inject when the default policy is to inject

Option 2
---------
true => invert the default policy (if default is inject, then don't, if default is don't inject, then do)
false => do what the default policy says.

I was thinking that the name `-override` might be a little less obvious that the behavior is Option 1. Either way, I think it's fine.

linsun:
It could be just me... i have mentally associated opt-out to opt-out injection thus no injection.  opt-in means yes i want the injection.  BTW, what would be the default values for these after user installed istio?  

i like sidecar.istio.io/inject better.  i think given it is specified in the deployment, it will be considered an override by default to the cluster level config.

BTW, do you intend to allow user to specify this at the deployment level?  I asked because i could have a deployment but i'm not exposing it as a service.

[...]

PR #1208, opened 2017-09-08
Simplify initializer injection policy and annotation naming

ayj:

Simplify the initializer injection policy and annotation naming based on the discussion [here](https://github.com/istio/pilot/issues/1164)

* fixes https://github.com/istio/pilot/issues/1164
* Remove deprecated alpha annotation before istio/istio installs files are updated and 0.2 release
* Fix initializer bug introduced by https://github.com/istio/pilot/pull/1188

This PR is mostly just renaming things per the scheme below:

```bash
s/InjectionPolicyOptOut/InjectionPolicyEnabled
s/InjectionPolicyOptIn/InjectionPolicyDisabled
s/policy.sidecar.istio.io/sidecar.istio.io/inject
s/status.sidecar.istio.io/sidecar.istio.io/status
s/istioSidecarAnnotationPolicyValueForceOn/true
s/istioSidecarAnnotationPolicyValueForceOff/false
```

**Release note**:

```release-note
```


platform/kube/inject/inject.go
```diff
@@ -236,41 +222,47 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
...
 	} else {
-		var ok bool
-		if resourcePolicy, ok = annotations[istioSidecarAnnotationPolicyKey]; !ok {
-			resourcePolicy = istioSidecarAnnotationPolicyValueDefault
+		if value, ok := annotations[istioSidecarAnnotationPolicyKey]; !ok {
+			useDefault = true
+		} else {
+			// http://yaml.org/type/bool.html
+			switch strings.ToLower(value) {
+			case "y", "yes", "true", "on":
```
rshriram: may be add enabled ?

ayj: I would prefer to limit the number of valid values if possible. http://yaml.org/type/bool.html seemed like a reasonable compromise and is already standardized. If we can always extend later if necessary without backwards compat issues.

PR #1230, opened 2017-09-10
adding --service-zone flag to Envoy

### LINKED ISSUE #1212: Start Envoy with --service-zone
8 21:22:32.781]    proxy/envoy/envoy  [--parent-shutdown-time-s <uint64_t>] [--drain-time-s
I0908 21:22:32.781]                       <uint64_t>] [--file-flush-interval-msec <uint64_t>]
I0908 21:22:32.781]                       [--service-zone <string>] [--service-node <string>]
I0908 21:22:32.781]                       [--service-cluster <string>] [--hot-restart-version]
I0908 21:22:32.781]                       [--restart-epoch <uint64_t>] [-l <string>] [-c
I0908 21:22:32.781]                       <string>] [--concurrency <uint32_t>] [--base-id
I0908 21:22:32.781]                       <uint64_t>] [--] [--version] [-h]
I0908 21:22:32.781] 

Envoy startup has service zone option that indicates the local availability zone. Not sure if we start Envoy with the service zone option as I don't recall seeing one.

rshriram:
This PR adds an optional --availabilityZone flag to pilot-agent (which gets translated into --service-zone flag for Envoy). This lets the end user specify the availability zone associated with an envoy instance during startup time. It is possible to dynamically infer this from kubernetes, but it will bloat the pilot agent binary (40Mb+). If the flag is not specified, the service-zone flag will not be passed (this is the default for backward compatibility).

When injecting the pod from initializer or kube-inject, this flag will not be specified (and az aware routing will thus remain disabled). However, users wishing to manually inject the sidecar can specify the flag to take advantage of Envoy's az-aware routing.
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1212 

**Release note**:

```release-note
Ability to explicitly specify the availability zone associated with an Envoy sidecar in a pod
```

### DIFF
--- a/proxy/envoy/watcher.go
+++ b/proxy/envoy/watcher.go
@@ -126,8 +126,8 @@ func (proxy envoy) args(fname string, epoch int) []string {
 		"--service-node", proxy.node,
 	}...)
 
-	if len(proxy.config.ServiceZone) > 0 {
-		startupArgs = append(startupArgs, []string {"--service-zone", proxy.config.ServiceZone}...)
+	if len(proxy.config.AvailabilityZone) > 0 {
+		startupArgs = append(startupArgs, []string{"--service-zone", proxy.config.AvailabilityZone}...)
 	}
 
 	return startupArgs
--- a/proxy/envoy/watcher.go
+++ b/proxy/envoy/watcher.go
@@ -117,13 +117,20 @@ func NewProxy(config proxyconfig.ProxyConfig, node string) proxy.Proxy {
 }
 
 func (proxy envoy) args(fname string, epoch int) []string {
-	return []string{"-c", fname,
+	startupArgs := make([]string, 0)
+	startupArgs = append(startupArgs, []string{"-c", fname,
 		"--restart-epoch", fmt.Sprint(epoch),
 		"--drain-time-s", fmt.Sprint(int(convertDuration(proxy.config.DrainDuration) / time.Second)),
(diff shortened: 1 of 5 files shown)


proxy/envoy/watcher.go
```diff
@@ -117,13 +117,20 @@ func NewProxy(config proxyconfig.ProxyConfig, node string) proxy.Proxy {
 }
 
 func (proxy envoy) args(fname string, epoch int) []string {
-	return []string{"-c", fname,
+	startupArgs := make([]string, 0)
```
ldemailly: make and first append not needed

kyessenov: yeah, just store as startupArgs := []string{...} and then append.

ldemailly:
it seems harmless (famous last words after code freeze) so I'm ok with the change if it passes all tests - but - where is the (e2e) functional test ? (ie some test script with at least 2 zones and some test that the 2 zones are indeed behaving as expected (which I assume means send to same zone in preference to other zone? is the wanted behavior described ? how does it differ from k8s?
also does our installation support (and we have tests) for deploying pods in each zone for istio-system control plane?

rshriram:
K8s does not support zone-aware load balancing/routing AFAIK. Its just for failure domain isolation, and influencing scheduling of pods in particular AZs.

PR #1238, opened 2017-09-11
Fix RBAC: Unique name for ClusterRoleBinding when running on shared c…

andraxylia:
This matches the same fix istio/istio#716

Fixes istio/istio#712


kyessenov:
Link to the original issue, please.

Issue #1239, opened 2017-09-11
pilot e2e tests do not clean clusterrolebinding


andraxylia:
teardown deletes the namespace, but clusterrolebindings are not namespaced.

It should instead do kubectl delete rbac.yaml

PR #1240, opened 2017-09-11
RBAC delete ClusterRoleBindings

andraxylia:
Fixes istio/pilot#1239

### DIFF
--- a/test/integration/infra.go
+++ b/test/integration/infra.go
@@ -308,6 +308,14 @@ func (infra *infra) deployApp(deployment, svcName string, port1, port2, port3, p
 }
 
 func (infra *infra) teardown() {
+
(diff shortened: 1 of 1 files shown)


test/integration/infra.go
```diff
@@ -308,6 +308,14 @@ func (infra *infra) deployApp(deployment, svcName string, port1, port2, port3, p
 }
 
 func (infra *infra) teardown() {
+
+	if yaml, err := fill("rbac-beta.yaml.tmpl", infra); err != nil {
```
ldemailly: why is that still needed now that the rbac is inlined into istio.yaml ?

andraxylia: because the e2e pilot processes and applies the templates individually, not the istio.yaml.
I am making a similar change in istio/istio e2e, part of the tests do the same (I know this is not correct, I had opened an issue).

PR #1242, opened 2017-09-11
Build and push Debian package as part of the release

rvkubiak:
```
Build and push Debian package as part of the release
```


bin/push-debian.sh
```diff
@@ -0,0 +1,33 @@
...
+
+if [ -z "${BAZEL_ARGS}" ] || [ -z "${ISTIO_VERSION}" ]; then
+  usage
+fi
+
+set -ex
+
+bazel ${BAZEL_STARTUP_ARGS} build ${BAZEL_ARGS} "//tools/deb:istio-agent"
+gsutil -m cp -r \
+  bazel-bin/tools/deb/istio-agent_${ISTIO_VERSION}_amd64.* \
```
kyessenov: we should try to use static build to be consistent with docker files.

costinm: Good idea - but we can update it in 0.3, not critical I think.

bin/push-debian.sh
```diff
@@ -0,0 +1,33 @@
...
+  gs://istio-release/releases/${ISTIO_VERSION}/deb
```
sebastienvas: Please make this configurable like in:
https://github.com/istio/pilot/blob/master/bin/upload-istioctl#L23

bin/push-debian.sh
rshriram: Its just one small option.. --output..=static.

ldemailly:
Is it easy to also sign the packages with istio.io credentials? (Follow up PR maybe as we are short on time)

bin/push-debian.sh
rvkubiak: Added --output_groups=static to the bazel build args.

bin/push-debian.sh
rvkubiak: Done.

bin/push-debian.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if [ -z "${BAZEL_CONFIG}" ] || [ -z "${ISTIO_VERSION}" ] || [ -z "${GCS_PATH}" ]; then
+  usage
+fi
+
+set -ex
+
+bazel ${BAZEL_STARTUP_ARGS} build \
+  ${BAZEL_CONFIG} --output_groups=static \
+  "//tools/deb:istio-agent"
```
kyessenov: does deb rule properly handle static output group? there's a bug in bazel for doing something similar for docker, and it doesn't do the right thing.

rvkubiak: Removed.

kyessenov: Ok, since it's urgent, that's ok, but we should consider doing a workaround in bazel with file groups in the future.

rvkubiak: I have no sense of the urgency.  I only added it because @rshiraram suggested this was a trivial change.  Since I'm unfamiliar with the Bazel bug or the urgency of this change I'm not going to go back and forth anymore - whoever is invested in this change can make it.

kyessenov:
cc @costinm please fix to use static build

bin/push-debian.sh
kyessenov: Understood, cc'ed the right person.

rshriram: Oh, I didn't know that bazel had another bug for static output for debian files as well. :(. Will the script above work?

PR #1243, opened 2017-09-11
Change default intervals for discovery, drain duration, etc.

rshriram:
Change the default refresh interval, envoy drain duration. The current interval is suitable only for testing (1s). Increasing the default to 30s results in lesser CPU utilization.

**Release note**:

```release-note
NONE
```

FYI @gnirodi @suryadu


rshriram:
cant the two be in sync? Also, setting in istio/istio means the tests will take a while to run due to 30s RDS refresh.

ayj:
Pilot already has the notion of default mesh configuration. It seems reasonable for these hard-coded defaults to be optimized for a typical deployment rather than our unit test environment.

kyessenov:
Sure, but we also want to avoid hidden defaults. Hence, let's make all defaults explicit in istio/istio, and then the change here becomes a no-op.

test/integration/testdata/config.yaml.tmpl
```diff
@@ -22,3 +22,5 @@ data:
     mixerAddress: istio-mixer.{{.IstioNamespace}}:9091
 {{end}}
     authPolicy: {{ .Auth }}
+    disablePolicyChecks: false
+    enableTracing: true
```
ayj: Are these necessary to changing the discovery/refresh duration?

rshriram: Eh no.. sorry, I just wanted to include the default values to make sure that these are being parsed by Pilot.

ayj:
Agreed that explicit defaults should be updated in istio.istio. At the same time, hidden defaults can't be completely avoided unless hardcoded defaults are removed. It seems better to error on the side of redundancy here and avoid a configuration that isn't appropriate for anything but testing.

test/integration/testdata/config.yaml.tmpl
ayj: ah, right - I forgot what the defaults were here

kyessenov:
Please update both.

test/integration/testdata/config.yaml.tmpl
```diff
@@ -22,3 +22,5 @@ data:
     mixerAddress: istio-mixer.{{.IstioNamespace}}:9091
 {{end}}
     authPolicy: {{ .Auth }}
+    disablePolicyChecks: false
```
andraxylia: rdsRefreshDelay: 30s missing

ayj:
(1) - https://github.com/istio/istio/blob/master/install/kubernetes/templates/istio-pilot.yaml.tmpl#L8 is out of sync with https://github.com/istio/api/blob/master/proxy/v1/config/proxy_mesh.proto#L22

(2) - The defaults for ProxyConfig in MeshConfig aren't applied properly, i.e. they're completely overwritten. This results in the following error

```
* config path must be set
* binary path must be set
* service cluster must be set - Re-run kube-inject with `-i <istioSystemNamespace> and ensure valid MeshConfig exists
```

We haven't been able to distinguish between these errors (bug vs. out-of-sync configmap) because the underlying errors were hidden. That issue is fixed by https://github.com/istio/pilot/pull/1247 hence the reason for this PR

rshriram:
Someone care to run refresh goldens on this PR? I am on the road

kyessenov:
Will do

PR #1251, opened 2017-09-12
Make sure ProxyConfig defaults are applied

ayj:

`ProxyConfig` defaults are not applied properly to user defined configuration. The current code applies defaults by decoding the user supplied MeshConfig YAML into a MeshConfig protobuf with expected default values. This works for top-level fields in the protobuf but breaks with nested pointer types, i.e. `ProxyConfig`. 

The fix implemented here uses the same strategy but explicitly applies it to the nested ProxyConfig after the top-level configuration is decoded.

**Release note**:

```release-note
```

### DIFF
--- a/cmd/BUILD
+++ b/cmd/BUILD
@@ -5,7 +5,6 @@ go_library(
     srcs = ["cmd.go"],
     visibility = ["//visibility:public"],
     deps = [
-        "//model:go_default_library",
         "//proxy:go_default_library",
         "//tools/version:go_default_library",
         "@com_github_golang_glog//:go_default_library",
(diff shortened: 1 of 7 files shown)


rkpagadala:
Do we need a test for this?

ayj:
Please hold off on merging this. As @rkpagadala said, this should have proper unit tests and we need to fix the same problem in `ReadMeshConfig`.

kyessenov:
@rshriram Jason specifically requested not to merge this.

ayj:
I think @rshriram was reading my mind (or the PR details). I addressed the two issues I noted, but hadn't updated the PR description yet.

PR #1260, opened 2017-09-13
Use unique ClusterRole for shared clusters

andraxylia:
Use unique ClusterRole for shared clusters


ldemailly:
this file looks a lot like the one in isitio/istio we shouldn't have 2 (in 0.3)

andraxylia:
there is an issue open to track this: istio/istio#706

PR #1263, opened 2017-09-13
Recommend 1.7.5 for GKE and Minikube

ayj:

Recome

**Release note**:

```release-note
```

### DIFF
--- a/doc/testing.md
+++ b/doc/testing.md
@@ -46,6 +46,8 @@ kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-ad
 
 * Minikube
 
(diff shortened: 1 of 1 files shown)


linsun:
is this a recommendation or requirement?  wanted to make sure it is not really a requirement else I'd like to know why.

ayj:
Its a requirement for GKE since new clusters can only be created with 1.7.5, 1.6.9, or 1.6.7 (see https://cloud.google.com/container-engine/release-notes).

its a recommendation for Minikube so we use the latest patch release available.

linsun:
ok thanks!  so istio itself doesn't have any hard requirement for 1.7.5 that you are aware of.

ayj:
Correct

PR #1265, opened 2017-09-13
Updating protobuf archive checksum

sebastienvas:
The checksum of  https://github.com/google/protobuf/archive/v3.2.0.tar.gz changed under us. Probably because protobuf team rebuilt the package.

It seems it happened before 
https://github.com/google/protobuf/issues/3087

```release-note
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -450,7 +450,7 @@ go_repository(
 
 http_archive(
     name = "com_github_google_protobuf",
-    sha256 = "2a25c2b71c707c5552ec9afdfb22532a93a339e1ca5d38f163fe4107af08c54c",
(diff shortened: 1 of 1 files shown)


linsun:
Ah cool, didn't know we could update this!  thanks @sebastienvas~

xfxyjwf:
You'd better not use  https://github.com/google/protobuf/archive/v3.2.0.tar.gz because it's created automatically by github and according to https://github.com/libgit2/libgit2/issues/4343#issuecomment-328631745 it may change in the future again.

sebastienvas:
@xfxyjwf, which one should we use ?

kyessenov:
@xfxyjwf wouldn't re-uploading archives break rules_go in general then? There is a similar `http_archive` rule used by the rule_go macros.

sebastienvas:
https://github.com/istio/istio/pull/748 should fix the issue we were seeing in smoketest,

PR #1266, opened 2017-09-13
Egress routing fix

### LINKED ISSUE #1180: Re-enable routing rules for egress domains
Support for external services through egress rules has omitted routing rules/destination policies on these rules. Egress rule should be considered as just another service spec and so, the destination in route/destinationpolicy should be matched with the destination in the egress rule to generate the appropriate settings for timeouts/retries/circuit breakers, etc.

Without this, apps will not be able to use istio routing rules or any of Envoy's features when talking to external services.

rshriram:
Enable routing and destination policies for egress rules
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1180 

**Release note**:

```release-note
NONE
```

Its not correct. The goldens are not getting updated properly, even though all the clusters are being added to the common pool. Worse still, running the refresh goldens command, updates the circuit breaker settings for world service!

### DIFF
--- a/model/config.go
+++ b/model/config.go
@@ -456,7 +456,9 @@ func (store *istioConfigStore) Policy(instances []*ServiceInstance, destination 
 		return nil
 	}
 
-	var out *Config
+	// ugly go-ism
+	var out Config
+	var found bool
 	for _, config := range configs {
 		policy := config.Spec.(*proxyconfig.DestinationPolicy)
 		if !MatchSource(config.ConfigMeta, policy.Source, instances) {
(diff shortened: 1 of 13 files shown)


linsun:
We really need to simplify our egress configuration (hopefully with this PR).  I have been learning how to use it with help from @rshriram.  this feature in alpha is really hard to use.   I am working with a user and he decided not to use it because it is complicated.  He is just calling simple watson services and it times out in 15s so his app didn't work.  In order to configure that timeout, we require user to create a k8s external service, plus change app code, plus create the route rule for the destination.  It is just complicated while people will expect to work out of box, btw, it does work out of box with k8s without istio.

PR #1271, opened 2017-09-14
Support proper destination format in weighted rule

rshriram:
Fixes the TODO (and regression) in route.go, where destination field in DestinationWeight was not honored.

**Release note**:

```release-note
NONE
```


proxy/envoy/route.go
```diff
@@ -150,9 +150,11 @@ func buildHTTPRoute(config model.Config, service *model.Service, port *model.Por
 	if len(rule.Route) > 0 {
 		route.WeightedClusters = &WeightedCluster{}
 		for _, dst := range rule.Route {
-			// TODO: add destination service override from the route weights
-
-			cluster := buildOutboundCluster(destination, port, dst.Labels)
+			real_destination := destination
```
kyessenov: go convention is mixed caps (`actualDestination`)

model/conversion_test.go
```diff
@@ -47,11 +47,15 @@ func TestProtoSchemaConversions(t *testing.T) {
 		"precedence": 5,
 		"route": [
 		{
-			"destination": "bar",
+			"destination": {
+                "name" : "bar"
```
rkpagadala: fix indent, here and below

rshriram: It looks fine in Gogland.. any idea why this is happening?

kyessenov: Maybe tabs?

kyessenov: Actually github messes up whitespace. It might be fine despite what web page shows.

ldemailly: you have tabs you shouldn't use tabs

ldemailly:
it's odd this PR didn't pick up the breakage it is causing in consul
we should probably figure out why (off by one on linters vs workspace ?)
https://github.com/istio/pilot/pull/1275#issuecomment-329653627

rshriram:
Determined that this is probably not the cause of breakage. @GregHanson thoughts?

ldemailly:
yeah it's unrelated- seems to be a linter upgrade

PR #1274, opened 2017-09-14
Remove 'istioctl mixer' sub command and Issue an error message

mandarjog:
**What this PR does / why we need it**:
Mixer no longer uses the old `istioctl mixer` method of configuring Mixer.
At present we use it is replaced by `kubectl`

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:
`mixer` is removed from the list of `Available commands`.
```
Usage:
  istioctl [command]

Available Commands:
  completion  Generate bash completion for Istioctl
  create      Create policies and rules
  delete      Delete policies or rules
  get         Retrieve policies and rules
  help        Help about any command
  kube-inject Inject Envoy sidecar into Kubernetes pod resources
  register    Registers a service instance (e.g. VM) joining the mesh
  replace     Replace existing policies and rules
  version     Display version information
```
If mixer command is run we get the following
```
$ bazel-bin/cmd/istioctl/istioctl mixer
Error: this method of configuring Mixer is removed in favor of kubectl
$ echo $?
255
```
If a command from the old documentation is used
```
$ bazel-bin/cmd/istioctl/istioctl mixer rule create global ratings.default.svc.cluster.local -f ratelimit.yaml
Error: this method of configuring Mixer is removed in favor of kubectl
$ echo $?
255
```

**Release note**:
```release-note
`istioctl mixer` subcommand has been removed. Istioctl issues a specific error message if the command is used.
```

PR #1275, opened 2017-09-14
Load balanced TCP

ijsnellf:
Currently when the TCP port of a service does not have a load balancer address (`service.Address`), we create a pass-through cluster for the port. This maps well to k8s headless services, where no load balancing is desired. For Eureka/Consul TCP services, load balancing is expected. This PR adds load balancing for non-headless TCP services without a load balancing address.

**Which issue this PR fixes**: fixes TODO

As with headless TCP services, this solution requires that the TCP port be unique.

The behavior in Kubernetes should be unchanged with this PR. This change is only for Eureka/Consul services.

**Release note**:

```release-note
```


ldemailly:
```
I0914 23:02:12.556] platform/consul/controller_test.go:134::error: arg inst for printf verb %q of wrong type: *istio.io/pilot/model.ServiceInstance (vet)
I0914 23:02:12.556] platform/consul/controller_test.go:190::error: arg services for printf verb %q of wrong type: []*istio.io/pilot/model.Service (vet)
I0914 23:02:12.556] platform/consul/controller_test.go:208::error: arg services[0] for printf verb %q of wrong type: *istio.io/pilot/model.ServiceInstance (vet)
```
it's best to build before making a pr

also this seems like a significant change that has a chance to cause destabilization ?

rshriram:
This is a bug fix enabling tcp load balancing for non k8s deployments. This error seems totally off, given that controller_test does not have any rules. will investigate tomorrow.

rshriram:
Also, the last PR passed with flying green colors (https://github.com/istio/pilot/pull/1274), which means some cache is messed up somewhere.

ldemailly:
yeah it's odd - is it that somehow the String() function disappeared on those types ?

ijsnellf:
Checking this morning --I get the issue locally. Diagnosing...

ijsnellf:
Maybe `go vet` got updated and became more strict? Whatever the case, I fixed the issues.

I put the fixes in this PR, but I can put them in a separate PR if necessary.

platform/consul/controller_test.go
```diff
@@ -131,8 +131,7 @@ func TestInstances(t *testing.T) {
 			}
 		}
 		if !found {
-			t.Errorf("Instances() did not match by tag => %q, want tag {%q:%q}",
-				inst, filterTagKey, filterTagVal)
+			t.Errorf("Instances() did not match by tag {%q:%q}", filterTagKey, filterTagVal)
```
ijsnellf: Go vet decided it doesn't like code to format structs using `%q`.

ldemailly: %v then?

ijsnellf: Yep, that works. Which seems silly, because the behavior for `%q` is basically equivalent to `"%v"`. Whatever, I just removed the struct formating entirely instead.

PR #1276, opened 2017-09-15
Remove copying of meshconfig to app namespace

### LINKED ISSUE #1272: ConfigMap copy issue
In 0.1.x series of Pilot, the proxy agent generated configuration for TCP, and needed access to the core startup configuration of pilot discovery service. This resulted in a shared configuration unit (a configmap) that was meant to stay in sync between the agent and the service, as inconsistencies would likely result in inability to communicate between sidecars.

In 0.2.x series of Pilot, the situation has improved as most of the configuration is now centrally generated in the service. The agent still requires small amount of settings (that have not yet been worked into xDS APIs). These settings are mostly static (e.g. drain durations, etc), and need not be changed for the duration of pod existence. The ideal situation is full elimination of these settings when xDS fully covers the entire surface of Envoy config. These settings are currently injected by kube-inject or initializer or CI/CD or human operator as startup-flags. There's simply no need to access Pilot service config from the agent directly.

Since the agent does not read the config map, there's no longer need to copy ConfigMap from installation namespace (where pilot resides) to application namespace (where agent resides). Remember that ConfigMaps are only visible within a single namespace. 

Currently, kube-inject copies the configmap over to the application namespace as part of injection process. I believe we should remove the copying bit since the produced configmap is not being consumed.

@rshriram @andraxylia @costinm @linsun @louiscryan @frankbu

### LINKED ISSUE #1273: Ingress controller should reside in installation namespace
Current installation templates put ingress proxies in the application namespace (cc @andraxylia ). 
I recommend we put all ingress and egress proxies into the installation namespace, as they are managed by istio and handle cluster-wide ingresses, rather than having independent identical proxies in each application namespace.

@rshriram @linsun @costinm @louiscryan

kyessenov:
This PR removes mounting of configmap in kube-injected pods since that configmap is not used by the agent. This eliminates the need to copy the configmap to application namespaces.

Also moves ingress proxies to istio installation. That requires enabling cluster-wide service registry (since it has to listen to both app and istio namespaces). That should not cause problems on shared cluster (that's how we operated during 0.1.x development). The ingresses are contained by marking each resource with namespace, and setting the controller mode to STRICT.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1272, fixes #1273 

**Release note**:

```release-note
```


costinm:
- Are you also removing the ability to run ingress per namespace ? What is the relation with configmap mount and copy removal, and why is it necessary ? 
- Is it the smallest safe change ? It seems a lot of renames and removals - can you instead just suppress the generation of the map with a flag or constant, so we can add it back quickly ? Since this change seems pretty big - it may be tricky to revert when we start adding back the config map.

kyessenov:
1. To answer your first point, configmap mount requires copy or the pod does not start.
2. Linters do a good job finding unused variables. Once you remove the use, you start pulling the string.

adapter/config/ingress/status.go
```diff
@@ -69,7 +71,7 @@ func NewStatusSyncer(mesh *proxyconfig.MeshConfig, client kubernetes.Interface,
 
 	var publishService string
 	if mesh.IngressService != "" {
-		publishService = fmt.Sprintf("%v/%v", options.Namespace, mesh.IngressService)
+		publishService = fmt.Sprintf("%v/%v", ingressNamespace, mesh.IngressService)
```
kyessenov: ingress namespace and ingress service values are fixed-form to satisfy the ingress status syncer (a core k8s library)

platform/kube/inject/inject.go
```diff
@@ -328,7 +319,8 @@ func injectIntoSpec(p *Params, spec *v1.PodSpec) {
 		Command: []string{"/bin/sh"},
 		Args: []string{
 			"-c",
-			fmt.Sprintf("sysctl -w kernel.core_pattern=%s/core.%%e.%%p.%%t && ulimit -c unlimited", EnvoyConfigPath),
+			fmt.Sprintf("sysctl -w kernel.core_pattern=%s/core.%%e.%%p.%%t && ulimit -c unlimited",
+				p.Mesh.DefaultConfig.ConfigPath),
```
kyessenov: This is a minor bug fix: both core dumps and mount volume were not respecting `ConfigPath` field in the proxy config.

test/integration/infra.go
```diff
@@ -128,9 +128,6 @@ func (infra *infra) setup() error {
 	if err := deploy("rbac-beta.yaml.tmpl", infra.IstioNamespace); err != nil {
 		return err
 	}
-	if err := deploy("config.yaml.tmpl", infra.Namespace); err != nil {
```
kyessenov: Not needed to deploy in app namespace.

test/integration/ingress.go
```diff
@@ -75,16 +75,16 @@ func (t *ingress) run() error {
...
+		{"a", fmt.Sprintf("https://%s.%s:443/http", ingressServiceName, t.IstioNamespace), ""},
+		{"b", fmt.Sprintf("https://%s.%s:443/pasta", ingressServiceName, t.IstioNamespace), ""},
+		{"a", fmt.Sprintf("http://%s.%s/lucky", ingressServiceName, t.IstioNamespace), ""},
+		{"b", fmt.Sprintf("http://%s.%s/lol", ingressServiceName, t.IstioNamespace), ""},
+		{"a", fmt.Sprintf("http://%s.%s/foo", ingressServiceName, t.IstioNamespace), "foo.bar.com"},
+		{"a", fmt.Sprintf("http://%s.%s/bar", ingressServiceName, t.IstioNamespace), "foo.baz.com"},
+		{"a", fmt.Sprintf("grpc://%s.%s:80", ingressServiceName, t.IstioNamespace), "api.company.com"},
+		{"a", fmt.Sprintf("grpcs://%s.%s:443", ingressServiceName, t.IstioNamespace), "api.company.com"},
+		{"", fmt.Sprintf("http://%s.%s/notfound", ingressServiceName, t.IstioNamespace), ""},
+		{"", fmt.Sprintf("http://%s.%s/foo", ingressServiceName, t.IstioNamespace), ""},
```
kyessenov: Ingress by default is in istio-system.

platform/kube/inject/inject_test.go
```diff
@@ -166,23 +164,18 @@ func TestIntoResourceFile(t *testing.T) {
...
-				DebugMode:         c.debugMode,
+				InitImage:       InitImageName(unitTestHub, unitTestTag, c.debugMode),
+				ProxyImage:      ProxyImageName(unitTestHub, unitTestTag, c.debugMode),
+				ImagePullPolicy: "IfNotPresent",
+				Verbosity:       DefaultVerbosity,
+				SidecarProxyUID: DefaultSidecarProxyUID,
+				Version:         "12345678",
+				EnableCoreDump:  c.enableCoreDump,
+				Mesh:            &mesh,
+				DebugMode:       c.debugMode,
```
ldemailly: wtb github diff -w (ignore whitespaces) - this is a lot harder to read as a diff than it should be

kyessenov:
Waiting for ack from @andraxylia and @costinm

test/integration/testdata/egress-proxy.yaml.tmpl
```diff
@@ -41,10 +41,10 @@ spec:
         - -v
         - "{{.Verbosity}}"
         - --discoveryAddress
-        - istio-pilot.{{.IstioNamespace}}:8080
+        - istio-pilot:8080
 {{if .Zipkin}}
         - --zipkinAddress
-        - zipkin.{{.IstioNamespace}}:9411
+        - zipkin:9411
```
andraxylia: General remark about zipkin, unrelated to the changes in this PR:

If the user has zipkin already deployed in another namespace, he will not be able to configure Istio to use it, even if he changes the configMap, because this is hardcoded in the templates.

test/integration/testdata/ingress.yaml.tmpl
```diff
@@ -3,7 +3,7 @@ kind: Ingress
 metadata:
   name: secured-ingress
   annotations:
-    kubernetes.io/ingress.class: istio
+    kubernetes.io/ingress.class: {{ .IstioNamespace }}
```
andraxylia: Nice idea for restricting which ingresses are handled by the controller, but a bit unorthodox, the class is supposed to be something like "istio", "ngnx", "traeffik".

Can we add another annotation for the restriction and keep the original one as "istio"? Still in 0.2, since it will be harder to change this later.

andraxylia: I understand the above would require a few more lines in the controller.

test/integration/testdata/egress-proxy.yaml.tmpl
kyessenov: ConfigMap is also a template. We need folks to change zipkin address in two places in the installation yaml.

test/integration/testdata/ingress.yaml.tmpl
kyessenov: It's set `istio` by default. This test overrides it to a different value.

andraxylia: istio/istio ingress template must be changed in the same commit as that of upgrading pilot to this SHA. It has to be done manually, do not rely on the bot.

PR #1280, opened 2017-09-15
teach istioctl non-hyphenated resource names

ayj:

Teach istioctl non-hyphenated resource names for proxy CRD types, e.g. `istioctl get routerules`.

fixes https://github.com/istio/pilot/issues/1236

Question: support for hyphenated names is still retained. Shall we remove that to avoid being stuck with multiple naming variants? 

**Release note**:

```release-note
```

### DIFF
--- a/adapter/config/crd/client.go
+++ b/adapter/config/crd/client.go
@@ -160,7 +160,7 @@ func (cl *Client) RegisterResources() error {
 				Scope:   apiextensionsv1beta1.NamespaceScoped,
 				Names: apiextensionsv1beta1.CustomResourceDefinitionNames{
 					Plural: ResourceName(schema.Plural),
-					Kind:   kabobCaseToCamelCase(schema.Type),
+					Kind:   KabobCaseToCamelCase(schema.Type),
 				},
 			},
 		}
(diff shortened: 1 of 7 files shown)


mandarjog:
We should detect and specifically error out on the old `hyphenated-names`, that way everyone is forced to update them, and has clear direction on what to update it with.

mandarjog:
Thanks @ayj 
/lgtm

frankbu:
LGTM. I would vote for removing the old hyphenated support ASAP, since it will just make it harder to be sure that people are using the new version and their rules are the new format.

ayj:
`egressrule` is already supported

I also updated the simple output to be roughly similar to kubectl instead of outputting the encoded config key. PTAL

```bash
$ istioctl get routerule
NAME			KIND		NAMESPACE
details-default		route-rule	default
productpage-default	route-rule	default
ratings-default		route-rule	default
ratings-test-delay	route-rule	default
reviews-default		route-rule	default
```

ayj:
@kuat, I added the fully qualified apiversion to the `knowntypes` in adapter/config/crd/types.go so that `istioctl get` shows this:

```bash
apiVersion: config.istio.io/v1alpha2
kind: RouteRule
metadata:
<...>
```
... instead of this:

```bash
apiVersion: v1alpha2
kind: RouteRule
metadata:
<...>
```

kyessenov:
Sounds good. Thanks for heads up.

kyessenov:
ok to merge?

ayj:
Yes. It passed tests but was blocked on CLA. I rebase'd to fix CLA issue and tests are re-running.

PR #1291, opened 2017-09-17
Use regex field instead of overloading header match for regex routes

rshriram:
We currently use header match for regex matching of routes. Instead, switch to the regex field introduced as a first class entity in Envoy config.
In addition, this PR uses prefix based matching for Ingress paths, instead of exact path based matches, as most users are used to the standard nginx ingress behavior (prefix match) [most people using istio ingress were confused as to why things didn't work]

**Release note**:

```release-note
Istio ingress defaults to prefix based URL match, similar to Kubernetes ingress.
```

Do not merge until Proxy sha is updated.


kyessenov:
What happened to two approvals?

kyessenov:
This doesn't work for combining ingress and route rules.
If I create a rule saying to route to version v2 if route matches certain regex, the rule will be ignored. Please follow-up with a fix or revert.

K8s spec requires path to be a regex. Nginx does not follow the spec, so we should not follow the spec, too, in your opinion?

rshriram:
A subsequent PR will resolve the combination issue between ingress rule and route rules.

ldemailly:
we need TWO reviews before making significant changes

PR #1295, opened 2017-09-18
use cloud.google.com instead of www.google.com in egress test

vadimeisenbergibm:
in countries other than US, www.google.com redirects to local versions
of google e.g. www.google.co.il, www.google.de etc.

**Which issue this PR fixes** *(optional, in `fixes 

this redirection fails the egress test. cloud.google.com does not redirect and works OK outside of US

#<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Release note**:

```NONE
```

### DIFF
--- a/test/integration/egress.go
+++ b/test/integration/egress.go
@@ -57,7 +57,7 @@ func (t *egress) setup() error {
 		},
 		Spec: v1.ServiceSpec{
 			Type:         "ExternalName",
-			ExternalName: "www.google.com",
+			ExternalName: "cloud.google.com",
 			Ports: []v1.ServicePort{{
 				Port: 443,
 				Name: "https", // important to define protocol


rshriram:
This should not even access anything.. Is it actually accessing www.google.com ? If so, we can revert to simple httpbin.org

vadimeisenbergibm:
@rshriram it does accesses httbin.org and www.google.com and expects to receive 200 status code. In countries outside of US, request to www.google.com, will return 302 with redirection to local google site and the test will fail.

ijsnellf:
I think @rshriram was suggesting using https://httpbin.org instead of google.

vadimeisenbergibm:
@ijsnellf The test currently tests both httbin.org and google.com. I think it is a good idea to check more than external site.

PR #1298, opened 2017-09-18
bugfix: remove istio initializer from istio-system workloads

ayj:
istio-system namespace is handled differently than normal user
namespaces and should not have sidecar's injected. This fixes a bug in
that logic where the sidecar was not injected (correct) but the
pending list of initializers was not cleared (bug). This caused
istio-system components to not be deployed in cluster-wide
installations to istio-system name.

**Release note**:

```release-note
```

### DIFF
--- a/platform/kube/inject/inject.go
+++ b/platform/kube/inject/inject.go
@@ -446,6 +446,13 @@ func intoObject(c *Config, in interface{}) (interface{}, error) {
 		return nil, err
 	}
 
+	// skip special kubernetes system namespaces
+	for _, namespace := range ignoredNamespaces {
+		if obj.GetNamespace() == namespace {
+			return out, nil
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


platform/kube/inject/inject.go
```diff
@@ -446,6 +446,13 @@ func intoObject(c *Config, in interface{}) (interface{}, error) {
 		return nil, err
 	}
 
+	// skip special kubernetes system namespaces
+	for _, namespace := range ignoredNamespaces {
+		if obj.GetNamespace() == namespace {
+			return out, nil
+		}
+	}
+
```
rshriram: This is just moving from initializer to kube-inject.. is it because initializer calls into inject.go ?

ayj: Yes. Resources in istio-system need their pending initializer list to be cleared without the sidecar actually being injected.

PR #1299, opened 2017-09-18
Enable regex based route combination in ingress

rshriram:
Combine ingress rule and route rule based on path, prefix and regex.
If regex is present in ingress, pick routes with paths/prefixes that match.
If regex is present in routes, pick ingresses with paths/prefixes that match.
if regex is present in ingress and route, return concatenated regex (intersection)

Also, reverts the default match behavior in ingress, back to "exact" match

**Release note**:

```release-note
NONE
```


proxy/envoy/resources.go
```diff
@@ -252,34 +254,126 @@ func (route *HTTPRoute) CatchAll() bool {
...
-		return route
-	case prefix == "" && route.Path == "" && strings.HasPrefix(path, route.Prefix):
-		// if mixed, pick if route prefix satisfies the path and change route to path
-		route.Path = path
-		route.Prefix = ""
-		return route
-	default:
+	case path != "":
+		ingressMatch = MatchPath
+	case prefix != "":
```
kyessenov: Is it still the case that either path or prefix is set, even for regex?

proxy/envoy/ingress_test.go
```diff
@@ -44,16 +48,23 @@ func TestRouteCombination(t *testing.T) {
 		{path3, prefix1, nil},
 		{path3, prefix2, nil},
 		{path3, prefix3, path3},
+		{path1, regex1, path1},
+		{prefix1, regex1, prefix1},
```
kyessenov: Shouldn't this be `path1`? Regex only matches `/xyz` or `/xya`

proxy/envoy/resources.go
rshriram: yes. Only one is allowed. path/prefix/regex

proxy/envoy/ingress_test.go
rshriram: Yes, the prefix1 is /xyz

rshriram: Oh I see what you mean. But we can't know ahead of time. prefix matching a regex does not mean we convert the prefix into an exact path, because that prefix might be an invalid path

kyessenov: The test is incorrect then. Please fix the test and the logic. I suggest converting prefix into a regex.

proxy/envoy/resources.go
kyessenov: I see, that's not quite obvious from here https://github.com/istio/pilot/blob/master/proxy/envoy/header.go#L28

PR #1306, opened 2017-09-19
Adds istio.deps for API

sebastienvas:
```release-note
none
```


istio.deps
```diff
@@ -1,5 +1,12 @@
 [
 	{
+		"name": "ISTIO_API",
```
ldemailly: oh that's how the magic works, it looks for some variable name...

PR #1307, opened 2017-09-19
Test mixer

kyessenov:
Introduces a test mixer that echoes received attributes to stdout. The plan is to mock mixer behavior and add pre-defined check policies, for testing pilot/proxy behavior.
Lightens the load on the test cluster as the test mixer is very lightweight at the moment.
The tests can now validate that all attributes are set correctly, without requiring full-fledged mixer processing.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: 
contributed to #533

**Release note**:

```release-note
```


kyessenov:
I wanted to avoid a direct dependency on mixer codebase since it's moving fast, and I'm not using most of Mixer functionality in my test. I think the ideal solution is to pimp up istio/api with codegen stuff. Is that what you propose? Note that most of sed-ding is to get rid of gogoproto annotations since they're very hard to build with current rules_go.

costinm:
Don't we have a test mixer in the proxy integration tests ? If they serve
different purpose it's ok to have 2, but
should probably be shared, it doesn't make sense to have one test style per
git repo.

istio/istio has common test infra - may be the best place.

kyessenov:
They serve different purpose.

douglas-reid:
improving istio/api sounds good. i was also hopeful that we could have a forcing function to improve `@io_istio_mixer//test` such that users wouldn't be impacted by changes to mixer proper.  but we can work on that separately.

test/mixer/server.go
```diff
@@ -0,0 +1,121 @@
...
+	bag := s.decode(msg.Attributes)
+	fmt.Printf("[%d] words.size=%d\n", s.counter, len(msg.Attributes.Words))
+	for k, v := range bag {
+		fmt.Printf("[%d] %s=%s\n", s.counter, k, v)
+	}
+	return &istio_mixer_v1.CheckResponse{
+		Precondition: &istio_mixer_v1.CheckResponse_PreconditionResult{
+			Status: &google_rpc.Status{Code: 0},
+		},
+	}, nil
```
rshriram: Is it possible to mock this or add a parameter to this such that we can decide when Mixer should fail or accept a check request? Would come in handy during testing stability

test/mixer/wordlist_process.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+cat <<EOF
```
rshriram: missing copyright..
Also, please add a comment on what this shell script does.

test/mixer/pb/proto_process.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
rshriram: copyright

rshriram:
Largely LGTM, missing copyrights in couple of files. And some high level comment on the shell script that is generating go code. I thought you hated shell scripts!
Might also want to update our docs (testing.md) stating that we use a mock mixer for tests.

rshriram:
and I am going by @douglas-reid 's reviews for the mixer replacement. So, if the replacement is good with mixer folks, fine by me.

test/mixer/server.go
kyessenov: Yeah, we can do absolutely anything now that we can have a server endpoint. This is just a start to get data path validation, before enforcing control behavior. I'm thinking of adding control HTTP endpoints on this to mark start/end of a test, and then send rules for asserting check response.

test/mixer/wordlist_process.sh
kyessenov: This is auto-generated code. I'll add one copyright on the script.

test/mixer/pb/proto_process.sh
kyessenov: Same

costinm:
So is the expectation that any project using mixer will create its own mock
mixer ? Or will they depend on pilot (or proxy) and reuse their mixer ?

Depending on mixer repo will be required for any project developing plugins
- given the WORKSPACE diffs, it would be quite tricky to depend on
both mixer and pick the mock from pilot repo.

istio/api or istio/istio seem better places for code test infra that may be
reused - the fact proxy also needs a mock mixer suggest it's going to be
a common need. I suppose a mock pilot will be great to have in the same
place as well.

kyessenov:
The more mixers in the world the better. The protocol is easy to implement, and we need wire-level observation to validate that we program mixerclient correctly. 

Mixer plugins have entirely different needs. They will use templates and bazel machinery, and should be disallowed access to raw attributes.

kyessenov:
PTAL. Added copyrights and documentation for shell scripts.

rshriram:
@ZackButcher this is non functional change, to improve our test suites for better mixer tests. @douglas-reid LGTM'ed it as well.

Issue #1232, opened 2017-09-11
Integration test coverage in 0.2.x


rshriram:
The following features are not covered by Pilot's integration tests. Need to fix this in the coming weeks to improve test coverage
- [ ] Mongodb/redis filters (@rshriram)
- [x] Egress rules (@rshriram, @vadimeisenbergibm, @elevran  )
- [ ] Consul ( @GregHanson )
- [x] Eureka ( @ijsnellf )

(modified by @kyessenov)

kyessenov:
Issue moved to [istio/istio #1454](https://github.com/istio/istio/issues/1454) via [**ZenHub**](https://www.zenhub.com/)

PR #1316, opened 2017-09-19
Eureka integration tests

ijsnellf:
Eureka integration tests.

**Which issue this PR fixes**: contributes to https://github.com/istio/pilot/issues/1232

I'm looking for quick feedback on the overall approach. When `bin/e2e.sh -registry Eureka` is called, I spin up an Eureka server and an accompanying mirroring process that listens for k8s endpoints and registers them with the Eureka server. Then e2e tests proceed mostly as normal. This approach has the advantage requiring very minimal changes to the e2e tests.

TODO:
- [x] improve synchronization logic in the Eureka mirror.
- [x] fix ingress/egress support (perhaps a follow-on PR)

TODO future PRs:
- remove Eureka client duplication (switch to `hudl/fargo` or another 3rd party Eureka library?)
- enable automated the Eureka tests

Note that normal e2e tests will be unaffected by this PR.

**Release note**:

```release-note
```


docker/Dockerfile.eurekamirror
```diff
@@ -0,0 +1,3 @@
+FROM scratch
```
rshriram: Add a comment stating that this container is used only for internal testing and what it does.

test/integration/egress_rules.go
```diff
@@ -39,6 +40,14 @@ func (t *egressRules) setup() error {
 
 // TODO: test negatives
 func (t *egressRules) run() error {
+	if !t.Egress {
+		glog.Info("skipping test since egress is missing")
+		return nil
+	}
+	if platform.ServiceRegistry(t.Registry) == platform.EurekaRegistry { // TODO: re-enable once Eureka egress is working
```
rshriram: This should work out of the box in Eureka as well right? These are egress rules that come directly from CRDs. No service registry lookup.

test/integration/ingress.go
```diff
@@ -39,6 +40,9 @@ func (t *ingress) setup() error {
 	if !t.Ingress {
 		return nil
 	}
+	if platform.ServiceRegistry(t.Registry) == platform.EurekaRegistry { // TODO: re-enable once Eureka ingress is working
+		return nil
+	}
```
rshriram: Same applies here. Unless the ingress has to know anything about Eureka, this should not be any different. It is referring to services by name. (Imagine there is some foo system where you can create ingress elements, outside k8s. the config would look similar to k8s ingress config). 
Or am I missing something?

test/integration/tcp.go
```diff
@@ -36,6 +37,10 @@ func (t *tcp) teardown() {
 }
 
 func (t *tcp) run() error {
+	// TCP in Eureka is tested by the headless service test.
```
rshriram: 👍

test/integration/testdata/eureka.yaml.tmpl
```diff
@@ -0,0 +1,38 @@
...
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: eureka
+    spec:
+      containers:
+      - name: eureka
+        image: netflixoss/eureka:1.3.1
```
rshriram: 👍 nice!

test/integration/ingress.go
ijsnellf: Yes, you are correct. It should be working. I'm still trying to figure out what is going wrong.

ijsnellf:
@rshriram: ingress/egress are currently broken. I'm trying to figure out the cause.

kyessenov:
Nice work! Can you add some comments for the purpose of eurekamirror?

test/eurekamirror/eurekamirror.go
```diff
@@ -0,0 +1,475 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"time"
+
+	log "fmt"
```
GregHanson: remove duplicate import and format statments

test/integration/egress_rules.go
```diff
@@ -39,6 +39,11 @@ func (t *egressRules) setup() error {
 
 // TODO: test negatives
 func (t *egressRules) run() error {
+	if !t.Egress {
```
ijsnellf: @vadimeisenbergibm: please verify this is correct

test/integration/testdata/pilot.yaml.tmpl
```diff
@@ -81,6 +81,12 @@ spec:
         args:
         - discovery
         - --v={{.Verbosity}}
+        - -a
+        - {{.Namespace}}
+        - --registries
+        - {{.Registry}}
+        - --eurekaserverURL
```
kyessenov: is this safe to supply if eureka mirror is not deployed?

ijsnellf:
@kyessenov, @ayj: PTAL

adapter/config/ingress/BUILD
```diff
@@ -42,7 +42,9 @@ go_test(
         "//proxy:go_default_library",
         "//test/mock:go_default_library",
         "//test/util:go_default_library",
+        "@com_github_golang_glog//:go_default_library",
```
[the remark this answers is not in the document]
ijsnellf: nice catch

test/integration/testdata/pilot.yaml.tmpl
ijsnellf: yes, the value isn't used

PR #1318, opened 2017-09-19
Simplify use_bazel_go with bazel info

sebastienvas:
As @kyessenov suggested this is less error prone.

```release-note
none
```

### DIFF
--- a/bin/use_bazel_go.sh
+++ b/bin/use_bazel_go.sh
@@ -1,22 +1,18 @@
 # This file should be sourced before using go commands
 # it ensures that bazel's version of go is used
 
-ROOT=$( cd "$(dirname "${BASH_SOURCE[0]}")/.." ; pwd -P )
-DIR_NAME="$(basename ${ROOT})"
+EXEC_ROOT="$(bazel info execution_root)"
 
-TARGETBIN="${ROOT}/bazel-${DIR_NAME}"
-if [[ ! -e $TARGETBIN ]]; then
-  echo "*** $TARGETBIN does not exist - did you forget to bazel build ... ?"
+if [[ ! -e ${EXEC_ROOT} ]]; then
+  echo "*** ${EXEC_ROOT} does not exist - did you forget to bazel build ... ?"
   exit 1
 fi
 
-BDIR=$(dirname $(dirname $(readlink $TARGETBIN)))
-
-export GOROOT="$(find ${BDIR}/external -type d -name 'go1_*')"
-export PATH=$GOROOT/bin:$PATH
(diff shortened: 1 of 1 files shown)


bin/use_bazel_go.sh
```diff
@@ -1,22 +1,18 @@
...
+if [[ ! -e ${EXEC_ROOT} ]]; then
+  echo "*** ${EXEC_ROOT} does not exist - did you forget to bazel build ... ?"
   exit 1
 fi
 
-BDIR=$(dirname $(dirname $(readlink $TARGETBIN)))
-
-export GOROOT="$(find ${BDIR}/external -type d -name 'go1_*')"
-export PATH=$GOROOT/bin:$PATH
+export GOROOT="$(find ${EXEC_ROOT}/external -type d -name 'go1_*')"
```
rshriram: is there a way by which this can be fixed? if find returns multiple go versions, this is going to come back Empty as it happens on my Mac.
```bash
shriram@Shrirams-MacBook-Pro:~/istio.io/pilot $ find /private/var/tmp/_bazel_shriram/bc806ee45d4ae0d0f6926c893dc9a8c0/external -type d -name 'go1_*'
/private/var/tmp/_bazel_shriram/bc806ee45d4ae0d0f6926c893dc9a8c0/external/go1_8_3_darwin_amd64
/private/var/tmp/_bazel_shriram/bc806ee45d4ae0d0f6926c893dc9a8c0/external/go1_9_darwin_amd64
```

And the rest of the script fails. Is it atleast possible to do something like
```bash
BAZEL_GOROOT="$(find ...)"
if !-z BAZEL_GOROOT 
GOROOT=BAZEL_GOROOT
fi
```

Or some such fallback logic?

sebastienvas: can you bazel clean on your mac ?
https://docs.google.com/document/d/1sFIlrJCoAbIQCuzYIMi5dnY86Au2Nl_neNhB1m14xxU/edit

kyessenov: I have multiple go libraries in that path, too

rshriram: I did all the clean expunge stuff.
I still have the two go directories :(.

sebastienvas: I think failure in this case is a feature, otherwise you don't know what you are building with. There should not be 2 and if there is this is an indication that we should clean the cache.

rshriram: Okay, I did `bazel clean --expunge` and it seems to have blown away everything and rebuilding from scratch now. Seems to be downloading go 1.9 (bazel 0.5.4).

PR #1323, opened 2017-09-20
Context-create command for istioctl

GregHanson:
adds `config-create` command to istioctl.  Components of istio require a kubeconfig file.  To run istio in environment other than kubernetes, user would need to install and use `kubectl` to generate this file.  The new command creates/modifies the kubeconfig file so `kubectl` is not needed for non-k8s environments.  Uses k8s libraries for generation and writing to file

**Release note**:

```release-note
```

### DIFF
--- a/cmd/istioctl/main.go
+++ b/cmd/istioctl/main.go
@@ -429,20 +429,31 @@ and destination policies.
 				return err
 			}
 
-			config := clientcmdapi.Config{
-				Clusters: map[string]*clientcmdapi.Cluster{
-					configName: {Server: u.String()},
-				},
-				Contexts: map[string]*clientcmdapi.Context{
-					configName: {Cluster: configName, AuthInfo: ""},
-				},
-				CurrentContext: configName,
+			configAccess := clientcmd.NewDefaultPathOptions()
+			// use specified kubeconfig file for the location of new config
+			configAccess.GlobalFile = kubeconfig
+
+			config, err := configAccess.GetStartingConfig()
+			if err != nil {
+				return err
 			}
 
-			pathOpts := clientcmd.NewDefaultPathOptions()
-			// use specified kubeconfig file for the location of new config
-			pathOpts.GlobalFile = kubeconfig
-			err = clientcmd.ModifyConfig(pathOpts, config, false)
+			cluster, exists := config.Clusters[configName]
+			if !exists {
+				cluster = clientcmdapi.NewCluster()
+			}
+			cluster.Server = u.String()
+			config.Clusters[configName] = cluster
+
+			context, exists := config.Contexts[configName]
+			if !exists {
+				context = clientcmdapi.NewContext()
+			}
+			context.Cluster = configName
+			config.Contexts[configName] = context
+
+			config.CurrentContext = configName
+			err = clientcmd.ModifyConfig(configAccess, *config, false)
 
 			return err
 		},
--- a/cmd/istioctl/main.go
+++ b/cmd/istioctl/main.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"strings"
 	"text/tabwriter"
@@ -32,16 +33,14 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
-
-	"net/url"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	"istio.io/pilot/adapter/config/crd"
 	"istio.io/pilot/cmd"
 	"istio.io/pilot/model"
 	"istio.io/pilot/platform/kube"
 	"istio.io/pilot/tools/version"
-	"k8s.io/client-go/tools/clientcmd"
-	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
 const (
@@ -54,6 +53,7 @@ var (
 	kubeconfig     string
 	namespace      string
 	istioNamespace string
+	contextName    string
 
 	// input file name
 	file string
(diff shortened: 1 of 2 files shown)


GregHanson:
bug in modifying existing config instead of overwriting - DO NOT MERGE

cmd/istioctl/main.go
```diff
@@ -403,6 +407,58 @@ and destination policies.
...
+	configCmd = &cobra.Command{
+		Use:   "config-create <ip>:<port>",
+		Short: "Create base kubeconfig file for a server",
+		Example: `
+		# Create a config file for the api server.
+		istioctl config-create http://127.0.0.1:8080
+		`,
+		RunE: func(c *cobra.Command, args []string) error {
+
+			configName := "istio"
```
rshriram: Make this as an option and call it istio context name with default being istio

cmd/istioctl/main.go
```diff
@@ -403,6 +407,58 @@ and destination policies.
...
+			if err != nil {
+				c.Println(c.UsageString())
+				return err
+			}
+
+			configAccess := clientcmd.NewDefaultPathOptions()
+			// use specified kubeconfig file for the location of new config
+			configAccess.GlobalFile = kubeconfig
+
+			config, err := configAccess.GetStartingConfig()
```
rshriram: double checking. Why this kubeconfig thing in GlobalFile ? is that a fallback if we couldn't create a config ourselves? 
And does this function call fail if local config file is not present, but global one is present? I guess not.

rshriram:
@mandarjog @jmuk fyi.

ayj:
Neat. General question, what happens if user already has an existing kubeconfig? Does this overwrite it or produce some meaningful error? 

Also, it would be good to document that this only supports unauthenticated access to local apiserver. Alternatively, maybe add additional flags for `--certificate-authority`, `--client-key`, `--client-certificate` a. la https://coreos.com/kubernetes/docs/latest/configure-kubectl.html

rshriram:
Also, is it possible to have an --api-server CLI option as subflag of config create? 
```
istioctl --kubeconfig /tmp/gihanson-config config-create --context foobar --api-server http://172.28.0.13:8080
```

GregHanson:
@ayj the `config-create` command will create a kubeconfig file specified by `--kubeconfig` flag or use the default ~/.kube/config location.  

If the file already exists, it will append the new server cluster and context and switch the context to the newly created one. the context and config are created with name `istio` by default or the value specified by `--context` flag.  If a context and config already exist with the specified name, it will modify those.  I followed the behavior from the `kubectl config create-context/cluster` commands:
https://github.com/kubernetes/kubernetes/blob/5c558ddb185a7c120e147f24d69b7470bf7457e2/pkg/kubectl/cmd/config/create_cluster.go#L101

GregHanson:
I modified the command to print a statement if there is an existing kubeconfig file with current context already set.  So the user will be notified that their kubeconfig file settings have changes if they are using it

kyessenov:
Can we rename `config-create` to something different? At first, I thought this is about creating route rules.  Perhaps, we should name it `kubeconfig-create` or `context-create`?

rshriram:
context-create sounds good.

cmd/istioctl/main.go
```diff
@@ -403,6 +408,66 @@ and destination policies.
 		},
 	}
 
+	configCmd = &cobra.Command{
+		Use:   "context-create --api-server <ip>:<port>",
```
jmuk: nit: this use line is slightly different from the example section below; `--api-server http://<ip>:port` ?

cmd/istioctl/main.go
```diff
@@ -403,6 +408,66 @@ and destination policies.
...
+		# Create a config file for the api server.
+		istioctl context-create --api-server http://127.0.0.1:8080
+		`,
+		RunE: func(c *cobra.Command, args []string) error {
+			if istioAPIServer == "" {
+				c.Println(c.UsageString())
+				return fmt.Errorf("specify the the Istio api server IP")
+			}
+
+			u, err := url.ParseRequestURI(istioAPIServer)
```
jmuk: As far as I tried, `<ip>:<port>` format will be invalid with `ParseRequestURI` or `Parse` (see https://play.golang.org/p/BbVPMSpKkh). Are we okay with that?  Since it's set through `u.String()` below, I'm unsure why it needs to be parsed here.

GregHanson: @jmuk are you suggesting we not parse the server url value?  Or are you saying it is letting invalid URL's through or throwing errors for valid URLs?

GregHanson:
updated command:
```
istioctl --kubeconfig /tmp/gihanson-config context-create --context foobar --api-server http://172.28.0.13:8080
```

cmd/istioctl/main.go
kyessenov: What is the expected format for the `Server` a URI or an authority?

cmd/istioctl/main.go
```diff
@@ -403,6 +408,66 @@ and destination policies.
 		},
 	}
 
+	configCmd = &cobra.Command{
+		Use:   "context-create --api-server http://<ip>:<port>",
+		Short: "Create base kubeconfig file for a server",
```
ldemailly: I'd insert something to make it clear it's not to be used for k8s
for instance
```
"Create a kubeconfig for a non kubernetes cluster",
```
or similar (better english smithing welcome but the concept should be there I think)

cmd/istioctl/main.go
jmuk: I'm saying `127.0.0.1:8080` would fall into parse error by `url.ParseRequestURI`. It needs to be `http://127.0.0.1:8080`. It is not clear to me this is intentional.

If it is intentional, please change the `Use` line above. If it's not, maybe we should fix this line. For the fixing, I'm wondering what format is expected in the `Server` field.  Are we okay to directly passing a string `"127.0.0.1:8080"`?

cmd/istioctl/main.go
rshriram: He has left for the day. Will fix in subsequent PR.

PR #1331, opened 2017-09-21
Fix proxy agent lock-up

### LINKED ISSUE #1329: Ingress stopped working after 30mins
I was testing with bookinfo app, everything was working for the first 30 mins and then ingress stopped responding.

environment: GKE 1.7.5
```
export CA_HUB="gcr.io/istio-testing"
export CA_TAG="87901f278dd40e864b083409dfee9ca72e6a27e9"
export MIXER_HUB="gcr.io/istio-testing"
export MIXER_TAG="5c3f8f14085870d8af05ca5f26e30227bd6b3e72"
export ISTIOCTL_URL="https://storage.googleapis.com/istio-artifacts/pilot/e5681371e971d20b89823e55e394b2f773e7f07a/artifacts/istioctl"
export PILOT_HUB="gcr.io/istio-testing"
export PILOT_TAG="e5681371e971d20b89823e55e394b2f773e7f07a"
export PROXY_TAG="c7785bd97bd3d1610bafdda9ceaa3a061a641bad"
export ISTIO_NAMESPACE="istio-system"
export AUTH_DEBIAN_URL="https://storage.googleapis.com/istio-artifacts/auth/87901f278dd40e864b083409dfee9ca72e6a27e9/artifacts/debs"
export PILOT_DEBIAN_URL="https://storage.googleapis.com/istio-artifacts/pilot/e5681371e971d20b89823e55e394b2f773e7f07a/artifacts/debs"
export PROXY_DEBIAN_URL="https://storage.googleapis.com/istio-artifacts/proxy/c7785bd97bd3d1610bafdda9ceaa3a061a641bad/artifacts/debs"
```

ingress and service are up and running.
```bash
mjog@mjog0:~/GOHOME/src/istio.io/istio/samples/apps/bookinfo/rules$ kubectl get ing -o wide
NAME      HOSTS     ADDRESS          PORTS     AGE
gateway   *         35.202.121.196   80        2h
mjog@mjog0:~/GOHOME/src/istio.io/istio/samples/apps/bookinfo/rules$ k g svc -listio=ingress -o wide
kubectl --namespace istio-system get svc -listio=ingress -o wide
NAME            CLUSTER-IP      EXTERNAL-IP      PORT(S)                      AGE       SELECTOR
istio-ingress   10.47.244.164   35.202.121.196   80:32240/TCP,443:30792/TCP   3h        istio=ingress
mjog@mjog0:~/GOHOME/src/istio.io/istio/samples/apps/bookinfo/rules$ curl -v 35.202.121.196
* Rebuilt URL to: 35.202.121.196/
* Hostname was NOT found in DNS cache
*   Trying 35.202.121.196...
* connect to 35.202.121.196 port 80 failed: Connection refused
* Failed to connect to 35.202.121.196 port 80: Connection refused
* Closing connection 0
curl: (7) Failed to connect to 35.202.121.196 port 80: Connection refused
mjog@mjog0:~/GOHOME/src/istio.io/istio/samples/apps/bookinfo/rules$ k g pods -listio=ingress -o wide
kubectl --namespace istio-system get pods -listio=ingress -o wide
NAME                             READY     STATUS    RESTARTS   AGE       IP            NODE
istio-ingress-2193820168-81wb5   1/1       Running   0          3h        10.44.2.135   gke-mixer3-default-pool-a91d241b-dtrv
```

kyessenov:
The last change to the logic to abort-all envoys introduced a scenario when envoy could dead lock on rapid config changes (as is possible with firing config scheduling for the same config too often as done by the cert watcher).

This fixes the case by always reconciling on any error (abort or not). The reconciliation logic will apply in case the current config aborts immediately after reconciliation.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1329 

**Release note**:

```release-note
```

### DIFF
--- a/proxy/agent.go
+++ b/proxy/agent.go
@@ -214,18 +214,26 @@ func (a *agent) Run(ctx context.Context) {
 			// cleanup for the epoch
 			a.proxy.Cleanup(status.epoch)
 
-			// schedule a retry for a transient error and skip aborts
-			if status.err != nil && status.err != errAbort && !reflect.DeepEqual(a.desiredConfig, a.currentConfig) {
-				if a.retry.budget > 0 {
-					delayDuration := a.retry.InitialInterval * (1 << uint(a.retry.MaxRetries-a.retry.budget))
-					restart := time.Now().Add(delayDuration)
-					a.retry.restart = &restart
-					a.retry.budget = a.retry.budget - 1
-					glog.V(2).Infof("Updated retry delay to %v, budget to %d", delayDuration, a.retry.budget)
+			// schedule a retry for an error.
+			// the current config might be out of date from here since its proxy might have been aborted.
+			// the current config will change on abort, hence retrying prior to abort will not progress.
+			// that means that aborted envoy might need to re-schedule a retry if it was not already scheduled.
+			if status.err != nil {
+				// skip retrying twice by checking retry restart delay
+				if a.retry.restart == nil {
+					if a.retry.budget > 0 {
+						delayDuration := a.retry.InitialInterval * (1 << uint(a.retry.MaxRetries-a.retry.budget))
+						restart := time.Now().Add(delayDuration)
+						a.retry.restart = &restart
+						a.retry.budget = a.retry.budget - 1
+						glog.V(2).Infof("Epoch %d: set retry delay to %v, budget to %d", status.epoch, delayDuration, a.retry.budget)
+					} else {
+						glog.Error("Permanent error: budget exhausted trying to fulfill the desired configuration")
+						a.proxy.Panic(a.desiredConfig)
+						return
+					}
 				} else {
-					glog.Error("Permanent error: budget exhausted trying to fulfill the desired configuration")
-					a.proxy.Panic(a.desiredConfig)
-					return
+					glog.V(2).Infof("Epoch %d: restart already scheduled", status.epoch)
 				}
 			}
 
--- a/proxy/agent_test.go
+++ b/proxy/agent_test.go
@@ -364,3 +364,41 @@ func TestCascadingAbort(t *testing.T) {
 	a.ScheduleConfigUpdate(2)
 	<-ctx.Done()
 }
+
+// TestLockup plays a scenario that may cause a deadlock
+//  * start epoch 0
+//  * start epoch 1 (wait till sets desired and current config)
+//  * epoch 0 crashes, triggers abort
+//  * epoch 1 aborts
+//  * progress should be made
+func TestLockup(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	crash := make(chan struct{})
+	try := 0
+	start := func(config interface{}, epoch int, abort <-chan error) error {
+		if try >= 2 {
+			cancel()
+			return nil
+		}
+		try = try + 1
+		switch epoch {
+		case 0:
+			<-crash
+			return errors.New("crash")
+		case 1:
+			close(crash)
+			err := <-abort
+			return err
+		}
+		return nil
+	}
+	a := NewAgent(TestProxy{start, func(_ int) {}, nil}, testRetry)
+	go a.Run(ctx)
+	a.ScheduleConfigUpdate(0)
+	a.ScheduleConfigUpdate(1)
+	select {
+	case <-ctx.Done():
+	case <-time.After(1 * time.Second):
+		t.Error("liveness check failed")
+	}
+}


proxy/agent_test.go
```diff
@@ -364,3 +364,41 @@ func TestCascadingAbort(t *testing.T) {
...
 	<-ctx.Done()
 }
+
+// TestLockup plays a scenario that may cause a deadlock
+//  * start epoch 0
+//  * start epoch 1 (wait till sets desired and current config)
+//  * epoch 0 crashes, triggers abort
+//  * epoch 1 aborts
+//  * progress should be made
+func TestLockup(t *testing.T) {
```
ldemailly: does the test fail without the fix ?

kyessenov: Yes

kyessenov: Aborts are triggered by an abnormal exit code.

ldemailly: what trigger those ?

kyessenov: Any malfunction in envoy. Exit on draining is not abnormal. The reason for aborts is that we found out that once envoy starts crashing, the regular draining protocol breaks down, so we have to kill all active proxies (draining and new ones).

ldemailly: once detected, would it be easier to restart the agent too (and kill all child before that) or let kube restart the container maybe ?

kyessenov: Yes, but that would also restart application container.

mandarjog:
@kyessenov @rshriram as the name of the branch suggests, this is not the first time something like this has happened.
This PR fixes a specific race condition, but we need a backstop. 
In this case agent knew something was wrong, Envoy was dead at which point, the pod should have died or health check should have failed.

We need to let higher level nannies do their job.

kyessenov:
That's precisely the problem here. Agent doesn't know at all times whether proxy is dead or is just slow to initialize. This particular happened probably because some DNS resolution was slow, and envoy took longer than normal to initializing. If you pay attention to kube-dns, you can see it crashing from time to time. If you happen to make a DNS resolution during that time, Envoy will wait for DNS timeout and not be active until then.

kyessenov:
I'd appreciate code reviews here. The bug slipped through the previous ones. Coordinating multiple processes and go-routines is complicated despite the few lines of code involved.

mandarjog:
@kyessenov `ps aux` showed envoy was dead. Is that not definitely proof ?

kyessenov:
I'll send a follow-up fix to cert watcher spam. There should not have been two restarts triggered within milliseconds.

proxy/agent.go
```diff
@@ -214,18 +214,26 @@ func (a *agent) Run(ctx context.Context) {
...
+			// the current config will change on abort, hence retrying prior to abort will not progress.
+			// that means that aborted envoy might need to re-schedule a retry if it was not already scheduled.
+			if status.err != nil {
+				// skip retrying twice by checking retry restart delay
+				if a.retry.restart == nil {
+					if a.retry.budget > 0 {
+						delayDuration := a.retry.InitialInterval * (1 << uint(a.retry.MaxRetries-a.retry.budget))
+						restart := time.Now().Add(delayDuration)
+						a.retry.restart = &restart
+						a.retry.budget = a.retry.budget - 1
```
rkpagadala: is there no -- operator in go?
a.retry.budget--

rkpagadala: Just a suggestion how about the following less nested code?

	if status.err == nil {
		continue
	}
	// skip retrying twice by checking retry restart delay
	if a.retry.restart == nil {
		glog.V(2).Infof("Epoch %d: restart already scheduled", status.epoch)
        	continue
	}

	if a.retry.budget > 0 {
		delayDuration := a.retry.InitialInterval * (1 << uint(a.retry.MaxRetries-a.retry.budget))
		restart := time.Now().Add(delayDuration)
		a.retry.restart = &restart
		a.retry.budget--
		glog.V(2).Infof("Epoch %d: set retry delay to %v, budget to %d", status.epoch, delayDuration, a.retry.budget)
	} else {
		glog.Error("Permanent error: budget exhausted trying to fulfill the desired configuration")
		a.proxy.Panic(a.desiredConfig)
		return
	}

kyessenov: `--` is fine, just doesn't fit with the rest of the code style. i tried not to reorganize code too much, but your suggestion looks better i think.

PR #1335, opened 2017-09-22
Add dampening to file watcher. It will only deliver 1 event per maxDelay

mandarjog:
File watcher waits for maxDelay duration to deliver notification to updatefunc().

Current file watcher passes all events from fsnotify up to the agent. 
This can (and has) cause excessive event generation.

 It can also cause related events to be processed separately. For example cert and key file change events may be delivered separately even though they change together.

This is related to pr #1331 

```release-note
NONE
```

### DIFF
--- a/proxy/envoy/watcher.go
+++ b/proxy/envoy/watcher.go
@@ -68,8 +68,8 @@ func NewWatcher(config proxyconfig.ProxyConfig, agent proxy.Agent, role proxy.No
 	}
 }
 
-// defaultMaxDelay is the maximum amount of time to wait before delivering events.
-const defaultMaxDelay = 5 * time.Second
+// defaultMinDelay is the minimum amount of time between delivery of two successive events via updateFunc.
+const defaultMinDelay = 10 * time.Second
 
 func (w *watcher) Run(ctx context.Context) {
 	// agent consumes notifications from the controller
@@ -84,7 +84,7 @@ func (w *watcher) Run(ctx context.Context) {
 		certDirs = append(certDirs, cert.Directory)
 	}
 
-	go watchCerts(ctx, certDirs, watchFileEvents, defaultMaxDelay, w.Reload)
+	go watchCerts(ctx, certDirs, watchFileEvents, defaultMinDelay, w.Reload)
 
 	<-ctx.Done()
 }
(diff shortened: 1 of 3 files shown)


proxy/envoy/watcher.go
```diff
@@ -97,13 +103,52 @@ func (w *watcher) Reload() {
...
 
-// watchCerts watches a certificate directory and calls the provided
+type watchFileEventsFn func(ctx context.Context, wch <-chan *fsnotify.FileEvent,
+	maxDelay time.Duration, notifyFn func())
+
+// watchFileEvents watches for changes on a channel and notifies via notifyFn().
+// The function batches changes so that related changes are processed together.
+// The function ensures that notifyFn() is called no more than one time per maxDelay.
+// The function does not return until the the context is cancelled.
+func watchFileEvents(ctx context.Context, wch <-chan *fsnotify.FileEvent, maxDelay time.Duration, notifyFn func()) {
```
rkpagadala: How about using the following code? Ticker is part of the go library, which buffers only one timer event and drops the rest

	var ev *fsnotify.FileEvent
	// ticker and channel for managing maxDelay.
	var timeChan <-chan time.Time
	ticker := time.NewTicker(maxDelay)
	defer ticker.Stop()

	for {
		select {
		case ev = <-wch:
			glog.Infof("watchFileEvents: %s", ev.String())
			timeChan = ticker.C
		case <-timeChan:
			// wait for next event
			timeChan = nil

			glog.Info("watchFileEvents: notifying")
			notifyFn()
		case <-ctx.Done():
			glog.V(2).Info("watchFileEvents has terminated")
			return
		}
	}

mandarjog: great suggestion. I have updated the code.

proxy/envoy/watcher.go
```diff
@@ -68,6 +68,9 @@ func NewWatcher(config proxyconfig.ProxyConfig, agent proxy.Agent, role proxy.No
 	}
 }
 
+// defaultMaxDelay is the maximum amount of time to wait before delivering events.
```
kyessenov: minimum amount of time?

proxy/envoy/watcher.go
```diff
@@ -97,13 +103,45 @@ func (w *watcher) Reload() {
 	w.agent.ScheduleConfigUpdate(config)
 }
 
-// watchCerts watches a certificate directory and calls the provided
+type watchFileEventsFn func(ctx context.Context, wch <-chan *fsnotify.FileEvent,
+	maxDelay time.Duration, notifyFn func())
+
+// watchFileEvents watches for changes on a channel and notifies via notifyFn().
+// The function batches changes so that related changes are processed together.
+// The function ensures that notifyFn() is called no more than one time per maxDelay.
```
kyessenov: minDelay. That's the minimum time between invocations.

proxy/envoy/watcher.go
```diff
@@ -97,13 +103,45 @@ func (w *watcher) Reload() {
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case ev := <-wch:
+			glog.Infof("watchFileEvents: %s", ev.String())
+			timeChan = ticker.C
+		case <-timeChan:
+			// reset timer
+			timeChan = nil
```
kyessenov: i think having a boolean that is set to true in wch block and then reset in timeChan block is cleaner. setting to `nil` makes me suspicious what happens with reading from nil channel.

proxy/envoy/watcher.go
kyessenov: The approach seems fine. It doesn't eliminate the race fully. You can receive an FS event at 5s-\epsilon and capture inconsistent file system state at the next ticker. But the chances are pretty slim.

proxy/envoy/watcher.go
mandarjog: A nil channel is never ready. So it is never selected.

rkpagadala: reading from nil channel is idomatic go way of blocking

kyessenov:
It's not the linter `gotype` that's the problem; it's incorrect parameters from metalinter.

sebastienvas:
Can we merge this ? Can we fix the linter in another commit @kyessenov

mandarjog:
@kyessenov or I will fix the linter in another commit.

PR #1348, opened 2017-09-25
Use released version of gometalinter

sebastienvas:
Use released version of gometalinter instead of head of master.

```release-note
none
```


sebastienvas:
Updated to gometalinter.v1, disabled megacheck as it is not supported with 1.2.1, but re-enabled staticcheck and unused which produced even more issues. @ZackButcher @kyessenov I am closing this. Feel free to fix the linter issue.

PR #1349, opened 2017-09-25
remove nodeport from eureka service

ijsnellf:
Removes `NodePort` designation from Eureka service. This might cause issues with the automated tests if it is not removed. Fortunately, no Eureka tests have been enabled yet, so it shouldn't be causing issues right now.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/integration/testdata/eureka.yaml.tmpl
+++ b/test/integration/testdata/eureka.yaml.tmpl
@@ -9,7 +9,6 @@ spec:
     - protocol: TCP
       port: 8080
       name: http
-  type: NodePort
 ---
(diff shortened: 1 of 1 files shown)


ijsnellf:
tests fail because the linters got more strict. not related to this PR. will re-test once that is fixed

gyliu513:
@ijsnellf can you please show more detail for why does the `NodePort` will cause issue here, not quite catch the reason of this PR yet, thanks!

ldemailly:
@kyessenov ptal (doesn't seem like it'd affect k8s or tests so looks safe to me)

ldemailly:
tickle

PR #1352, opened 2017-09-25
Switch to using a stable snapshot of gometalinter

ZackButcher:
**What this PR does / why we need it**: Pilot presubmits have been failing due to linter issues introduced by changing linter versions. This PR pins pilot to using gometalinter.v1.

I also "fix" any new linter issues that popped up. Unfortunately gas doesn't recognize `_, _ = foo()` as "handling" an error, even if we explicitly ignore the error, so I had to annotate quite a few places as nolint. In config.go I actually attempted to handle the errors the linter was complaining about, but that resulted in test failures so I left them as-is.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/bin/install-prereqs.sh
+++ b/bin/install-prereqs.sh
@@ -3,8 +3,8 @@
 set -ex
 
 # Install linters
-go get -u github.com/alecthomas/gometalinter
-gometalinter --install --update --vendored-linters
+go get -u gopkg.in/alecthomas/gometalinter.v1
+gometalinter.v1 --install --update --vendored-linters
 
 # Install buildifier BUILD file validator
 go get -u github.com/bazelbuild/buildifier/buildifier
--- a/cmd/istioctl/main.go
+++ b/cmd/istioctl/main.go
@@ -565,6 +565,7 @@ func readInputs() ([]model.Config, []crd.IstioKind, error) {
 }
 
 // Print a simple list of names
+// nolint: errcheck, gas
 func printShortOutput(_ *crd.Client, configList []model.Config) {
 	var w tabwriter.Writer
 	w.Init(os.Stdout, 0, 8, 0, '\t', 0)
@@ -577,7 +578,7 @@ func printShortOutput(_ *crd.Client, configList []model.Config) {
 		)
 		fmt.Fprintf(&w, "%s\t%s\t%s\n", c.Name, kind, c.Namespace)
 	}
-	w.Flush() // nolint: errcheck
+	w.Flush()
 }
 
 // Print as YAML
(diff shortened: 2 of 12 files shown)


ZackButcher:
Heh, so I guess we're blocked waiting for the upsteam fix for gotype. Maybe just ignore this PR then.

cmd/istioctl/main.go
```diff
@@ -565,6 +565,7 @@ func readInputs() ([]model.Config, []crd.IstioKind, error) {
 }
 
 // Print a simple list of names
+// nolint: errcheck, gas
```
ldemailly: we probably shouldn't do that in non test files

kyessenov:
Is the stable snapshot also broken for `x_test` packages?

cmd/istioctl/main.go
kyessenov: megacheck replaces several older linters. we lose performance I think (may time out).

ZackButcher:
Forgot to include in my PR description: megacheck doesn't exist at gometalinter.v1's cut, so it doesn't work. gometalinter.v1 fails at execution time with an error about an unrecognized linter, even when megalinter is downloaded/on the path locally.

linsun:
Is this really needed for v0.2 since we are in the last week?

mandarjog:
Linter is not pinned in Pilot, so two days ago build started failing because the linter got auto updated. It has to be addressed.

Since this only affects the presubmit process it is low risk. It cannot introduce a problem in the actual compiled code.

ijsnellf:
If we can freeze to a specific of gometalinter, we can alternately use a commit after this fix is merged: https://github.com/alecthomas/gometalinter/pull/356

Or a commit before the bug was introduced.

bin/install-prereqs.sh
```diff
@@ -3,8 +3,8 @@
 set -ex
 
 # Install linters
-go get -u github.com/alecthomas/gometalinter
-gometalinter --install --update --vendored-linters
+go get -u gopkg.in/alecthomas/gometalinter.v1
+gometalinter.v1 --install --update --vendored-linters
```
dnephin: You can always pin to a specific version without having to pin to such an old version by using a git sha:

https://github.com/docker/cli/blob/master/dockerfiles/Dockerfile.lint#L6-L9

kyessenov: Will that install vendored listeners at the right SHA, too?
Seems like a docker image is probably cleanest way to accomplish this.

kyessenov:
Looking into pinning linters to a SHA with docker...

ZackButcher:
Yeah I think that's definitely preferable. If y'all get it working we'll update Mixer to do the same too.

PR #1363, opened 2017-09-27
Freeze linters

kyessenov:
Freezes gometalinter using a docker image.
Removes separate install-prerequisites step for buildifier and gometalinter.

Vendorization script is adjusted to remove double-symlinks since they don't work well with docker mounts. Generated files are placed into `vendor/` whenever possible for simpler cleanup.

**Release note**:

```release-note
```


bin/check.sh
```diff
@@ -1,11 +1,20 @@
...
+$buildifier -showlog -mode=check \
+  $(find . -type f \( -name 'BUILD' -or -name 'WORKSPACE' -or -wholename '.*bazel$' -or -wholename '.*bzl$' \) -print )
 
 NUM_CPU=$(getconf _NPROCESSORS_ONLN)
 
-gometalinter --concurrency=${NUM_CPU} --enable-gc --deadline=300s --disable-all\
+gometalinter="docker run \
+  -v $(bazel info output_base):$(bazel info output_base) \
+  -v $(pwd):/go/src/istio.io/pilot \
+  gcr.io/istio-testing/linter:bfcc1d6942136fd86eb6f1a6fb328de8398fbd80"
```
kyessenov: The SHA corresponds to gometalinter SHA. We can automate Dockerfile.lint once we want to get monthly upstream updates.

kyessenov:
cc @ZackButcher @sebastienvas if this looks fine to you, I can do a similar PR in mixer.

bin/check.sh
```diff
@@ -26,10 +36,9 @@ gometalinter --concurrency=${NUM_CPU} --enable-gc --deadline=300s --disable-all\
   --enable=varcheck\
   --enable=vet\
   --enable=vetshadow\
-  ./...
+  /go/src/istio.io/pilot/...
```
ldemailly: that won't work on people's machine will it ? (my directory is ~/go no /go)

kyessenov:
This refers to a directory inside the docker image.

bin/check.sh
ldemailly: oh right, sorry, I didn't see the docker run on line 11
is that image built each time ? (cached maybe but still) ? 
can we put the linters in some prow image we already have ?

kyessenov:
It's built once. I didn't do automation since linters shouldn't be changing
that often realistically.

bin/check.sh
sebastienvas: Will there be a permission issue here ? Will the link be created by root and not be readable by the user ?

kyessenov: This image is supposed to be created once a month or even more rarely. Linters don't change that fast. The image should be public readable.

rshriram: So, this would require docker on the machine in order to even run the linter? That seems like a bummer :(

rshriram: But if it stabilizes the build infrastructure, so be it.

kyessenov: i think we can change the script later to use system gometalinter. there's just that line at the end that needs to be changed to `./...` and use `-w` in docker.

PR #1364, opened 2017-09-27
fix: remove hyphens from istioctl types help text 

sublimino:
Updated example test on `istioctl`.

Fixes this `istioctl` command output:

```
$ istioctl 

Istio configuration command line utility.

Create, list, modify, and delete configuration resources in the Istio
system.

Available routing and traffic management configuration types:

        [route-rule ingress-rule egress-rule destination-policy]

```

The routing and traffic management types should now be hyphen-less as per rest of CLI. 

**Which issue this PR fixes** 
None 

**Release note**:

```
NONE
```

### DIFF
--- a/cmd/istioctl/main.go
+++ b/cmd/istioctl/main.go
@@ -68,7 +68,7 @@ var (
 		Short:             "Istio control interface",
 		SilenceUsage:      true,
 		DisableAutoGenTag: true,
-		Long: fmt.Sprintf(`
+		Long: `
 Istio configuration command line utility.
 
 Create, list, modify, and delete configuration resources in the Istio
@@ -76,12 +76,12 @@ system.
 
 Available routing and traffic management configuration types:
 
-	%v
+	[routerule ingressrule egressrule destinationpolicy]
 
 See http://istio.io/docs/reference for an overview of routing rules
 and destination policies.
 
-`, model.IstioConfigTypes.Types()),
+`,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			defaultNamespace = getDefaultNamespace(kubeconfig)
 		},


kyessenov:
It might be easier to clean up dashes in the `istioctl` main directly as some tests seem to depend on old config type names.

ayj:
[supportedTypes()](https://github.com/istio/pilot/blob/master/cmd/istioctl/main.go#L615) can be used to print the proper resource names (see line 83). This is already used for `istioctl get ...`

ldemailly:
@sublimino see https://github.com/istio/pilot/pull/1364#issuecomment-332600780
/retest

sublimino:
@ayj thanks for the pointer. When [`rootCmd` is declared in `var (...)`](https://github.com/istio/pilot/blob/master/cmd/istioctl/main.go#L65-L83) it looks like the CRD client needs [kubeconfig](https://github.com/istio/pilot/blob/master/cmd/istioctl/main.go#L488) to enumerate supported CRDs.

I'm not keen on reworking `main.go` for this unless recommended to (not sure how to offhand), otherwise I'll have a dig for the outdated tests.

adapter/config/crd/conversion.go
```diff
@@ -82,17 +83,25 @@ func KabobCaseToCamelCase(s string) string {
 	for _, word := range words {
 		out = out + strings.Title(word)
 	}
+	r, _ := regexp.Compile(`^([^-]*?)rule(.*)$`)
```
sublimino: This is a dirty regex hack (eugh, sorry). Test cases updated to use real model types, but I doubt the logic here is sufficiently future-proof.

The original logic expects input in the format `route-rule-x` to output `RouteRuleX`. With the first hyphen removed in this PR (so `routerule-x` to `RouteRuleX`) there is no easy way to use the previous logic for general strings (as this now treats `somerule` and `somepolicy` specially).

Possible solutions: 
- enumerating the allowed types (probably best?) 
- using this regex (to be tidied up if so)
- another way?

Awaiting feedback then I'll refactor this into something more maintainable (is that's ever possible with a regex).

ayj:
@kyessenov, how much work would it be to remove the dashed config names? That should be completely internal to Pilot without affecting any user-facing configuration.

adapter/config/crd/conversion.go
kyessenov: @rshriram you sure you want to keep this regex hack?

ayj:
NACK on this in its current form. As already stated, the regex is ugly and isn't future proof anyway. Something like the following would be safer and limit hacks/tmp changes to a single location.

```diff
$ git diff cmd/istioctl/main.go 
diff --git a/cmd/istioctl/main.go b/cmd/istioctl/main.go
index f754ce1..4c490a5 100644
--- a/cmd/istioctl/main.go
+++ b/cmd/istioctl/main.go
@@ -67,7 +67,7 @@ var (
                Short:             "Istio control interface",
                SilenceUsage:      true,
                DisableAutoGenTag: true,
-               Long: fmt.Sprintf(`
+               Long: `
 Istio configuration command line utility.
 
 Create, list, modify, and delete configuration resources in the Istio
@@ -75,12 +75,12 @@ system.
 
 Available routing and traffic management configuration types:
 
-       %v
+       [routerule ingressrule egressrule destinationpolicy]
 
 See http://istio.io/docs/reference for an overview of routing rules
 and destination policies.
 
-`, model.IstioConfigTypes.Types()),
+`,
        }
 
        postCmd = &cobra.Command{
```

sublimino:
This PR requested review, it is not ready to merge.

Agree hard coding is simplest was to satisfy the title @ajy.

The title of this PR should actually be "remove hyphens from model/config.go" as the tests are where the complexity lies. 

I suspect the fix to the regex hack will be answered by https://github.com/istio/pilot/pull/1364#issuecomment-334020756

Shall I reduce this PR to https://github.com/istio/pilot/pull/1364#issuecomment-334883028 and PR the `model/config.go` refactor separately?

ayj:
> Shall I reduce this PR to #1364 (comment) and PR the model/config.go refactor separately?

Yes please. It's much more likely we can get the reduced PR into 0.2.x release whereas the larger refactoring probably ought to wait until 0.3.x.

PR #1373, opened 2017-09-29
istioctl: Use namespace from kube config

emaildanwilson:
Changes istioctl to use the namespace for the current context in local kube config if it exists.

fixes istio/istio#979

Let me know if you think this is the right approach.

### DIFF
--- a/cmd/istioctl/main.go
+++ b/cmd/istioctl/main.go
@@ -50,11 +50,12 @@ const (
 var (
 	platform string
 
-	kubeconfig     string
-	namespace      string
-	istioNamespace string
-	istioContext   string
-	istioAPIServer string
+	kubeconfig       string
+	namespace        string
+	istioNamespace   string
+	istioContext     string
+	istioAPIServer   string
+	defaultNamespace string
 
 	// input file name
 	file string
@@ -81,6 +82,9 @@ See http://istio.io/docs/reference for an overview of routing rules
 and destination policies.
 
 `, model.IstioConfigTypes.Types()),
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			defaultNamespace = getDefaultNamespace(kubeconfig)
+		},
 	}
 
 	postCmd = &cobra.Command{
@@ -102,8 +106,8 @@ and destination policies.
 				return errors.New("nothing to create")
 			}
 			for _, config := range varr {
-				if config.Namespace == "" {
-					config.Namespace = namespace
+				if config.Namespace, err = handleNamespaces(config.Namespace); err != nil {
+					return err
 				}
 
 				var configClient *crd.Client
@@ -174,8 +178,8 @@ and destination policies.
 				return errors.New("nothing to replace")
 			}
 			for _, config := range varr {
-				if config.Namespace == "" {
-					config.Namespace = namespace
+				if config.Namespace, err = handleNamespaces(config.Namespace); err != nil {
+					return err
 				}
 
 				var configClient *crd.Client
@@ -280,14 +284,15 @@ and destination policies.
 				return err
 			}
 
+			ns, _ := handleNamespaces(v1.NamespaceAll)
 			var configs []model.Config
 			if len(args) > 1 {
-				config, exists := configClient.Get(typ.Type, args[1], namespace)
+				config, exists := configClient.Get(typ.Type, args[1], ns)
 				if exists {
 					configs = append(configs, *config)
 				}
 			} else {
-				configs, err = configClient.List(typ.Type, namespace)
+				configs, err = configClient.List(typ.Type, ns)
 				if err != nil {
 					return err
 				}
@@ -362,8 +367,8 @@ and destination policies.
 				return errors.New("nothing to delete")
 			}
 			for _, config := range varr {
-				if config.Namespace == "" {
-					config.Namespace = namespace
+				if config.Namespace, err = handleNamespaces(config.Namespace); err != nil {
+					return err
 				}
 
 				// compute key if necessary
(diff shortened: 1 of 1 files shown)


rshriram:
Thanks a lot for this fix.. And thanks @kyessenov for pointing him in the right direction.

kyessenov:
LGTM from release standpoint

linsun:
LGTM from release standpoint

cmd/istioctl/main.go
```diff
@@ -491,7 +491,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&istioNamespace, "istioNamespace", "i", kube.IstioNamespace,
 		"Istio system namespace")
 
-	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", v1.NamespaceDefault,
+	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", getDefaultNamespace(),
```
ayj: `getDefaultNamespace()` uses `kubeconfig` but `kubeconfig` is also a flag. `getDefaultNamespace` should be invoked after flags are parsed.

cmd/istioctl/main.go
```diff
@@ -679,3 +679,17 @@ func prepareClientForOthers(configs []crd.IstioKind) (*rest.RESTClient, map[stri
 	}
 	return client, resources, nil
 }
+
+func getDefaultNamespace() string {
+	configAccess := clientcmd.NewDefaultPathOptions()
+	// use specified kubeconfig file for the location of the config to create or modify
+	configAccess.GlobalFile = kubeconfig
```
ayj: Can you pass `kubeconfig` as an argument to `getDefaultNamespace`? The less global var the better and it makes flag dependencies more obvious (see other comment on line 494).

emaildanwilson: yes, I'll update this. good catch!

emaildanwilson:
@ayj updated this per your requested change. 
Let me know if this is what you were expecting.

cmd/istioctl/main.go
```diff
@@ -491,7 +491,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&istioNamespace, "istioNamespace", "i", kube.IstioNamespace,
 		"Istio system namespace")
 
-	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", v1.NamespaceDefault,
+	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", getDefaultNamespace(kubeconfig),
```
ayj: Closer :)

`namespace` should be set _after_ flags are parsed instead of when they are defined. Specifically, add the following code to  `rootCmd.PersistentPreRun` 

```golang
if namespace == "" {
    namespace = getDefaultNamespace(kubeconfig)
}
```

emaildanwilson: ok, thank you for the additional information. I've never used cobra before but I think I get it now.

emaildanwilson:
I think there is some more work to do here since there are 3 namespaces to take into account.

- Command - if one is provided using -n
- Spec - in the yaml specs being sent for commands that accept them
- Default - of the current context if the kube config exists

For this to work the same as kubectl my understanding is that Command namespace should always override Spec and Default and the Default namespace should be used only if it doesn't exist in the Spec. Which makes Command > Spec > Default. Therefore we can't override the `namespace` field used to store the Command provided namespace with the Default namespace from config like I initially thought because they need to be applied differently based on whether or not there is a namespace defined in the spec.

I'll rework this early next week to store a separate value for the Default namespace from the config and apply this logic that matches kubectl.

Please let me know if I'm missing something and thank you for the reviews thus far.

cc @ayj @rshriram

rshriram:
I think we already had a provision for handling the namespace in the spec. IIRC. But @kyessenov would have more idea.

ayj:
Ah, yes you're right. It also looks like Command and Object namespace must match if both are specified, otherwise kubectl prints an error, i.e.

> error: the namespace from the provided object "A" does not match the namespace "B

emaildanwilson:
@ayj good point on the error. 
I've updated this to handle each namespace type correctly now.
I can make any changes you need that are out of spec if you can point me in the right direction.

Thank you!

ayj:
https://github.com/istio/pilot/pull/1373#issuecomment-333286037 looks good to me with the caveat about namespace mismatch between object and command.

emaildanwilson:
@ayj 
ok great. Yes, the logic I implemented includes the error when both the object and command line namespace are provided and do not match.

emaildanwilson:
I think this is ready for review again.
Let me know if anything else is needed on my side.
Thank you!
cc @ayj @rshriram

cmd/istioctl/main.go
```diff
@@ -280,14 +284,16 @@ and destination policies.
 				return err
 			}
 
+			objectNamespace := ""
+			ns, _ := handleNamespaces(objectNamespace)
```
ayj: Can you use `""` directly here?

emaildanwilson: will do, it's only for clarity.

ayj: Understood. Maybe use `model.NamespaceAll` or `v1.NamespaceAll`?

emaildanwilson: ok. I can do that for line 499 too.

emaildanwilson: updated now.

emaildanwilson:
/test pilot-presubmit

Looks like a flake.
`Post https://35.186.174.136/api/v1/namespaces: dial tcp 35.186.174.136:443: i/o timeout`

emaildanwilson:
yeh, something is up with the testing environment. Similar error on a different part of the tests now. `Post https://35.186.174.136/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions: dial tcp 35.186.174.136:443: i/o timeout` I'll try to rerun the tests later.

emaildanwilson:
@ayj I just noticed your comment about preprocMixerConfig and added that in.

emaildanwilson:
@kyessenov can this be approved now?

PR #1379, opened 2017-10-02
add locking to pod cache

### LINKED ISSUE #1377: Pilot crashed with " fatal error: concurrent map read and map write" error
Stack trace can be found here:
https://gist.github.com/mclarke47/50ff5bcfaa89a0b067e3cc10d6a59fbd

Pilot seemed to crash causing the ingresses to not know where to route traffic resulting in lots of `404 NR` errors in the envoy access logs.

Version 0.2.6

Deploying the same as in the [sample](https://github.com/istio/istio/blob/master/install/kubernetes/istio.yaml).

As a side note, is pilot able to operate in a HA mode? So that further crashes can be mitigated.

rkpagadala:
PodCache has no mutex protecting its internal state. Add RWMutex to fix this.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1377 

**Release note**:

```release-note
Fix race in Pilot PodCache
```

### DIFF
--- a/platform/kube/cache.go
+++ b/platform/kube/cache.go
@@ -75,9 +75,6 @@ func (pc *PodCache) getPodByIP(addr string) (*v1.Pod, bool) {
 
 // labelsByIP returns pod labels or nil if pod not found or an error occurred
 func (pc *PodCache) labelsByIP(addr string) (model.Labels, bool) {
-	pc.rwMu.RLock()
-	defer pc.rwMu.RUnlock()
-
 	pod, exists := pc.getPodByIP(addr)
 	if !exists {
 		return nil, false
--- a/platform/kube/cache.go
+++ b/platform/kube/cache.go
@@ -15,6 +15,8 @@
 package kube
 
 import (
+	"sync"
+
 	"k8s.io/api/core/v1"
 
 	"istio.io/pilot/model"
@@ -22,6 +24,7 @@ import (
 
 // PodCache is an eventually consistent pod cache
 type PodCache struct {
+	rwMu sync.RWMutex
 	cacheHandler
 
 	// keys maintains stable pod IP to name key mapping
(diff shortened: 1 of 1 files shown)


platform/kube/cache.go
```diff
@@ -66,6 +75,9 @@ func (pc *PodCache) getPodByIP(addr string) (*v1.Pod, bool) {
 
 // labelsByIP returns pod labels or nil if pod not found or an error occurred
 func (pc *PodCache) labelsByIP(addr string) (model.Labels, bool) {
+	pc.rwMu.RLock()
+	defer pc.rwMu.RUnlock()
+
 	pod, exists := pc.getPodByIP(addr)
```
kyessenov: double read lock?

rkpagadala: fixed PTAL

andraxylia: +1, please revert lines 78-80, not needed.

andraxylia:
hopefully this will not hurt performance significantly.

platform/kube/cache.go
```diff
@@ -53,6 +59,9 @@ func newPodCache(ch cacheHandler) *PodCache {
 
 // getPodByIp returns the pod or nil if pod not found or an error occurred
 func (pc *PodCache) getPodByIP(addr string) (*v1.Pod, bool) {
+	pc.rwMu.RLock()
+	defer pc.rwMu.RUnlock()
```
mandarjog: This looks fine.
Can you ensure that same problem is not present in getByKey ?

Also tests?

rkpagadala: Looking at GetByKey, tests coming in a separate PR

PR #1383, opened 2017-10-03
Port of iptables image to CentOS

sdake:
CentOS base OS requires iptables from CentOS containers and
its library dependencies, not from Ubuntu.  One possible solution
to this problem is to statically build iptables which may fix the
problem and allow one consistent set of images to be used for all
of the pilot repo.  I'll explore that shortly via COPR and see if
I can build a static xbtables.

This PR is for those people wanting to develop and deploy on CentOS.  I'm not certain it is suitable for merging, however, it works at present.

Compiling pilot binaries on CentOS results in nonfunctional containers in master.  This patch demonstrates one approach to working on CentOS.  It may be possible to approach this problem using a distroless approach with all static binaries.  

**Release note**:

```release-note NONE
```


bin/push-docker
```diff
@@ -63,7 +63,7 @@ IFS=',' read -ra tags <<< "${tags}"
 IFS=',' read -ra hubs <<< "${hubs}"
 
 pushd docker
-  for image in app proxy proxy_init proxy_debug pilot sidecar_initializer testmixer eurekamirror; do
+  for image in app proxy proxy_init_ubuntu proxy_init_centos proxy_debug pilot sidecar_initializer testmixer eurkamirror; do
```
ijsnellf: typo: eurkamirror -> eurekamirror

sdake: good catch - rebase gone wild :)

docker/Dockerfile.proxy_init_ubuntu
```diff
@@ -0,0 +1,3 @@
+FROM docker.io/centosistio/iptables_ubuntu:20170924-897e396
```
ldemailly: why is the ubuntu base image changed ?

bin/push-iptables-image
```diff
@@ -26,15 +27,19 @@ set -o pipefail
 
 PROJECT="istio-testing"
 DATE_PART=$(date +"%Y%m%d")
-SHA_PART=$(git show -q HEAD --pretty=format:%h)
+SHA_PART=$(git show HEAD --pretty=format:%h | head -1 || /bin/true)
 DOCKER_TAG="${DATE_PART}-${SHA_PART}"
-IMAGE_NAME="gcr.io/${PROJECT}/iptables:${DOCKER_TAG}"
+IMAGE_NAME_UBUNTU="docker.io/${PROJECT}/iptables_ubuntu:${DOCKER_TAG}"
```
ldemailly: we should make the hub a variable but not just change from gcr.io/istio-testing to docker.io/istio-testing

cc @sebastienvas

sdake: agreed.

docker/Dockerfile.proxy_init_ubuntu
sdake: a base image of iptables from Ubuntu with centos binaries results in 503 errors.  I tracked this down partially to the fact that xtables-multi receives a permission denied (via strace) so the iptables rules are never set.  The root cause is still unclear.

To be more clear, I think the correct answer to this problem is to build all dependency binaries statically (i.e. iptables) and use FROM scratch.  This could result in a functional mesh of ubuntu-built binaries loaded into a scratch container that works on both operating systems.

I'm not sure I have time for the next 3 weeks to investigate a distroless solution in further detail.  Hence I submitted a PR so others interested in CentOS can possibly take this up (or I'll sort it out when I return from travel).

ldemailly: thanks for the general explanation to my toplevel question, this inline question though was specifically about why the _ubuntu base image is changing the gcr.io current base to docker.io/centosistio/iptables_ubuntu (ie nothing related to centos)

sdake: The change of the hub and namespace was simply for my personal testing and should not merge as is.  As I can't push to gcr.io/istio-testing, this would be a regression for a commit to the pilot repository.  I'm not all that sure we really want to take this model forward (unless distroless doesn't work), however, I wanted to provide some context for others that may take up this work in the event that someone wants to solve it now, vs in a few weeks when I return from travel.  The change of the image name to iptables_ubuntu is needed by the patch to work correctly given the explanation to your top level question.

rshriram:
closing for the moment.

PR #1396, opened 2017-10-05
Redis filter

rshriram:
envoy supports mongo/redis for telemetry. This PR enables the redis filter.
Just putting it out there..

**Release note**:

```release-note
Support for telemetry gathering from redis connections
```

### DIFF
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/golang/glog"
 	multierror "github.com/hashicorp/go-multierror"
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -372,14 +372,8 @@ func buildTCPListener(tcpConfig *TCPRouteConfig, ip string, port int, protocol m
 		},
 	}
 
-	baseListener := &Listener{
-		Name:    fmt.Sprintf("tcp_%s_%d", ip, port),
-		Address: fmt.Sprintf("tcp://%s:%d", ip, port),
-		Filters: []*NetworkFilter{baseTCPProxy},
-	}
-
 	switch protocol {
-	case model.ProtocolMONGO:
+	case model.ProtocolMongo:
 		// TODO: add a watcher for /var/lib/istio/mongo/certs
 		// if certs are found use, TLS or mTLS clusters for talking to MongoDB.
 		// User is responsible for mounting those certs in the pod.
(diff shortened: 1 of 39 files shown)


proxy/envoy/config.go
```diff
@@ -390,13 +396,35 @@ func buildTCPListener(tcpConfig *TCPRouteConfig, ip string, port int, protocol m
...
+						StatPrefix:  "redis",
+						ConnPool: &REDISConnPool{
+							OperationTimeoutMS: int64(REDISDefaultOpTimeout),
+						},
+					},
+				}},
+			}
 		}
 	}
+	return baseListener
```
ijsnellf: readability nit: no reason for `baseListener` to be defined above the switch block, since it isn't used anywhere but this return statement. Either move the `baseListener` definition down, or just return the structure.

proxy/envoy/resources.go
```diff
@@ -84,6 +84,13 @@ const (
 	// MONGOProxyFilter is the name of the MONGO Proxy network filter.
 	MONGOProxyFilter = "mongo_proxy"
 
+	// REDISProxyFilter is the name of the REDIS Proxy network filter.
+	REDISProxyFilter = "redis_proxy"
+
+	// REDISDefaultOpTimeout is the op timeout used for REDIS Proxy filter
+	// TODO - Allow this to be configured.
+	REDISDefaultOpTimeout = 30 * time.Second / time.Millisecond
```
ijsnellf: either use int64 instead or do the conversion when you use it ([line 420](https://github.com/istio/pilot/pull/1396/files#diff-18db4698a9b14dcd564faeefd7fbd74aR420)). the 2nd option is probably better.

rshriram: I am using the second option in line 420. Do you want to change it?

ijsnellf: `REDISDefaultOpTimeout = 30 * time.Second / time.Millisecond` => 30µs, which is incorrect. I assume you want to set OperationTimeoutMS to 30 seconds, so what I'm recommending is
`REDISDefaultOpTimeout = 30 * time.Second`
and
`OperationTimeoutMS: int64(REDISDefaultOpTimeout / time.Millisecond)` => 30000
Basically, do all the conversion in one place.

proxy/envoy/resources.go
```diff
@@ -84,6 +84,13 @@ const (
 	// MONGOProxyFilter is the name of the MONGO Proxy network filter.
 	MONGOProxyFilter = "mongo_proxy"
 
+	// REDISProxyFilter is the name of the REDIS Proxy network filter.
```
ayj: style nit: This should probably be `RedixProxyfilter` unless REDIS is an [initialism](https://github.com/golang/go/wiki/CodeReviewComments#initialisms). Similar comment applies elsewhere w.r.t. REDIS and MONGO.

rshriram:
@ayj any more comments?

rshriram:
@kyessenov ?

proxy/envoy/resources.go
```diff
@@ -81,8 +81,16 @@ const (
...
+	// MongoProxyFilter is the name of the Mongo Proxy network filter.
+	MongoProxyFilter = "mongo_proxy"
+
+	// RedisProxyFilter is the name of the Redis Proxy network filter.
+	RedisProxyFilter = "redis_proxy"
+
+	// RedisDefaultOpTimeout is the op timeout used for Redis Proxy filter
+	// Currently it is set to 30ms (conversion happens in the filter)
+	// TODO - Allow this to be configured.
+	RedisDefaultOpTimeout = 30 * time.Second
```
kyessenov: 30 milliseconds, not seconds

rshriram: intentionally set to 30s.. i.e. no timeout

PR #1401, opened 2017-10-05
Ingress bug fixes

rshriram:
Some drive by bug fixes (attach port to the vhost so that requests with port can also be matched), use ingress instances when filtering routes, so that route rules can be scoped to ingress only.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes https://github.com/istio/issues/issues/87
https://github.com/istio/issues/issues/60

**Release note**:

```release-note
Ingress will allow port numbers in the authority header
```

### DIFF
--- a/adapter/config/ingress/conversion.go
+++ b/adapter/config/ingress/conversion.go
@@ -107,6 +107,10 @@ func createIngressRule(name, host, path, domainSuffix string,
 				MatchType: &proxyconfig.StringMatch_Exact{Exact: path},
 			}
 		}
+	} else {
+		rule.Match.Request.Headers[model.HeaderURI] = &proxyconfig.StringMatch{
+			MatchType: &proxyconfig.StringMatch_Prefix{Prefix: "/"},
+		}
 	}
 
 	return model.Config{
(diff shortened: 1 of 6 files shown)


rshriram:
@markeijsermans @rfevang @maruina when this test passes, can you test with the Pilot image from this test to see if the port number issue is mitigated? (i.e. if your clients send foo.com:80, the ingress will accept it. [In the ingress resource, you can continue using the old `host: foo.com` format]

rshriram:
This is the image: gcr.io/istio-testing/pilot:dc0d0e462a8db90d565c33eca9faa975e824ad40 (change istio.yaml to use this pilot image)

rfevang:
Confirmed, hostname routing works using the given image.

markeijsermans:
This fixes host routing issues. Paths are still problematic for grpc (see https://github.com/istio/issues/issues/87#issuecomment-334887649)

rshriram:
do the *new* steps defined here help: https://istio.io/docs/tasks/traffic-management/ingress.html ? @markeijsermans @rfevang @maruina

PR #1407, opened 2017-10-06
[DO NOT MERGE] added validate_clusters true

ijsnellf:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/proxy/envoy/resources.go
+++ b/proxy/envoy/resources.go
@@ -90,7 +90,7 @@ const (
 	// RedisDefaultOpTimeout is the op timeout used for Redis Proxy filter
 	// Currently it is set to 30ms
 	// TODO - Allow this to be configured.
-	RedisDefaultOpTimeout = 30 * time.Millisecond
+	RedisDefaultOpTimeout = 30 * time.Second
 
 	// WildcardAddress binds to all IP addresses
 	WildcardAddress = "0.0.0.0"
--- a/proxy/envoy/resources.go
+++ b/proxy/envoy/resources.go
@@ -330,7 +330,7 @@ func (host *VirtualHost) clusters() Clusters {
 
 // HTTPRouteConfig definition
 type HTTPRouteConfig struct {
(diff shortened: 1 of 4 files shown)


proxy/envoy/resources.go
```diff
@@ -88,9 +88,9 @@ const (
 	RedisProxyFilter = "redis_proxy"
 
 	// RedisDefaultOpTimeout is the op timeout used for Redis Proxy filter
-	// Currently it is set to 30ms (conversion happens in the filter)
+	// Currently it is set to 30ms
 	// TODO - Allow this to be configured.
-	RedisDefaultOpTimeout = 30 * time.Second
+	RedisDefaultOpTimeout = 30 * time.Millisecond
```
ldemailly: seems unrelated ? isn't this short ?

ijsnellf: completely unrelated, just accidentally got swept into the commit

ijsnellf: should not effect anything.

proxy/envoy/resources.go
```diff
@@ -88,9 +88,9 @@ const (
 	RedisProxyFilter = "redis_proxy"
 
 	// RedisDefaultOpTimeout is the op timeout used for Redis Proxy filter
-	// Currently it is set to 30ms (conversion happens in the filter)
+	// Currently it is set to 30ms
```
andraxylia: 30s is a more realistic timeout, isn't it?
let's update just the comment to say 30s and not change the timeout value.

ijsnellf: yes, you are right. please ignore, though. it doesn't have anything to do with the rest of the PR. i'll clean it up

ldemailly:
Can someone make that PR pass so I can try the Sha with the 404 test?

ldemailly:
I think we should still explore this option, at least confirm that the PR was generating the right config and it was used by envoy before dropping it

andraxylia:
Right, validate_cluster is still something that we want to add in master.

PR #1410, opened 2017-10-07
Add liveness and readiness to Ingress and Egress

andraxylia:
Reduce likelihood of 404 for Ingress and Egress by ensuring the container does not receive traffic when envoy does not run istio/istio#1038

Matches istio/istio#1055

The perfect fix is ADS, which will come later.


rshriram:
closing for the moment until migration.

PR #1411, opened 2017-10-09
Enable ingress mode for Eureka/Consul

rshriram:
Ingress is a generic concept applicable across platforms. K8S has ingress as first class resource (except its broken in several ways). In non k8s platforms, we cannot use this ingress resource spec reliably, as it requires enabling ingress controller without the status syncer (adds more complexities), especially if people are just adding it for the dummy ingress.

This PR defines a new proxy agent mode (LB) that allows the sidecar to obtain just outbound configs, and run as standalone envoy [which becomes ingress in Consul/Eureka]. With this, we would no longer have to expose services directly (like the way we do today). An added benefit is that this would allow people to run envoy as standalone using load balancer service in kubernetes, with support for TCP and HTTP services.

**Release note**:

```release-note
new Load Balancer mode for sidecar to enable ingress functionality in non-k8s environments
```

### DIFF
--- a/proxy/envoy/config.go
+++ b/proxy/envoy/config.go
@@ -117,7 +117,7 @@ func buildConfig(listeners Listeners, clusters Clusters, lds bool, config proxyc
 // buildListeners produces a list of listeners and referenced clusters for all proxies
 func buildListeners(env proxy.Environment, node proxy.Node) Listeners {
 	switch node.Type {
-	case proxy.Sidecar:
+	case proxy.Sidecar, proxy.LB:
 		instances := env.HostInstances(map[string]bool{node.IPAddress: true})
 		listeners, _ := buildSidecarListenersClusters(env.Mesh, instances,
 			env.Services(), env.ManagementPorts(node.IPAddress), node, env.IstioConfigStore)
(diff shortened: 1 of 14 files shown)


proxy/envoy/config.go
```diff
@@ -439,9 +443,17 @@ func buildOutboundListeners(mesh *proxyconfig.MeshConfig, sidecar proxy.Node, in
 	httpOutbound = buildEgressFromSidecarHTTPRoutes(mesh, instances, config, httpOutbound)
 
 	for port, routeConfig := range httpOutbound {
-		listeners = append(listeners,
-			buildHTTPListener(mesh, sidecar, instances, routeConfig, WildcardAddress, port,
-				fmt.Sprintf("%d", port), false, EgressTraceOperation))
+		var l *Listener
```
kyessenov: nit: use variable to pass different trace operation, use remote address flag values (e.g. `useRemoteAddress := true; operation := EgressTraceOperation`)

rshriram:
@kyessenov I fixed the TCP listener part.. It will only open listeners with wildcard listeners.. Also, removed support for egress rules from teh router (aka load balancer) as semantics are not really clear. I also found some stray implementation about egress proxy support in the egress rule, and got rid of that as well. 

And renamed the localhost:8888 to something more readable (istio-pilot/mixer.istio-system) so that the goldens are bit more grok-able..

Issue #1433, opened 2017-10-11
Add  ignoredNamespaces to sidecar initializer configMap


gyliu513:
More discussion here https://github.com/istio/istio.github.io/issues/648

PR #1434, opened 2017-10-11
Added new parameter to ignore inject namespaces.

gyliu513:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes https://github.com/istio/pilot/issues/1433

**Release note**:

```release-note
Added new parameter of ignoredNamespaces to sidecar initializer configMap
```

/cc @linsun @ayj

### DIFF
--- a/platform/kube/inject/inject.go
+++ b/platform/kube/inject/inject.go
@@ -159,7 +159,10 @@ type Config struct {
 	Policy InjectionPolicy `json:"policy"`
 
 	// deprecate if InitializerConfiguration becomes namespace aware
-	Namespaces []string `json:"namespaces"`
+	IncludeNamespaces []string `json:"includeNamespaces"`
+
+	// deprecate if InitializerConfiguration becomes namespace aware
+	ExcludeNamespaces []string `json:"excludeNamespaces"`
 
 	// Params specifies the parameters of the injected sidcar template
 	Params Params `json:"params"`
@@ -188,6 +191,10 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
 	var c Config
 	if err := yaml.Unmarshal([]byte(data), &c); err != nil {
 		return nil, err
+	}
+
+	if len(c.IncludeNamespaces) >= 1 && len(c.ExcludeNamespaces) >= 1 {
+		return nil, fmt.Errorf("Cannot configure both IncludeNamespaces and ExcludeNamespaces")
 	}
 
 	// apply safe defaults if not specified
@@ -456,6 +463,17 @@ func intoObject(c *Config, in interface{}) (interface{}, error) {
 	// skip special kubernetes system namespaces
 	for _, namespace := range ignoredNamespaces {
 		if obj.GetNamespace() == namespace {
+			return out, nil
+		}
+	}
+
+	// skip customized exclude namespaces
+	for _, excludeNamespace := range c.ExcludeNamespaces {
+		if excludeNamespace == "" {
+			return out, nil
+		}
+
+		if obj.GetNamespace() == excludeNamespace {
 			return out, nil
 		}
 	}
--- a/platform/kube/inject/inject_test.go
+++ b/platform/kube/inject/inject_test.go
@@ -152,6 +152,10 @@ func TestIntoResourceFile(t *testing.T) {
 			in:   "testdata/replicationcontroller.yaml",
 			want: "testdata/replicationcontroller.yaml.injected",
 		},
+		{
+			in:   "testdata/hello-ibm.yaml",
+			want: "testdata/hello-ibm.yaml.inject",
+		},
 	}
 
 	for _, c := range cases {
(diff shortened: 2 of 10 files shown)


platform/kube/inject/inject.go
```diff
@@ -460,6 +463,29 @@ func intoObject(c *Config, in interface{}) (interface{}, error) {
...
+			if c.Namespaces[0] == "" {
+				glog.V(2).Infof("The namespace %s will be ignored by sidecar initializer", ignoredNamespace)
+			} else {
+				for _, namespace := range c.Namespaces {
+					if ignoredNamespace == namespace {
+						glog.V(2).Infof("The namespace %s will be ignored by sidecar initializer", ignoredNamespace)
+						break
+					}
+				}
+			}
```
linsun: Is it necessary to do the if/else block?  It seems you'll ignore the namespace as long as it is on the ignoredNamespace list, regardless if it is specified in c.Namespaces.

gyliu513: Here the logic is that I want to make the `ignoredNamespace` has high priority than `Namespaces`.

The if/else logic is mainly used to print logs telling end user that the `ignoredNamespace` is overwriting the `Namespaces`.

We can remove the if/else logic, but then we will not able to check if the `ignoredNamespace` is overwriting the `Namespaces` or not.

linsun: I see your point now.  If it is mainly for the purpose of logging, you may want to clarify in that log that you are overwriting the `Namespaces`.

platform/kube/inject/inject.go
```diff
@@ -161,6 +161,9 @@ type Config struct {
 	// deprecate if InitializerConfiguration becomes namespace aware
 	Namespaces []string `json:"namespaces"`
```
rshriram: rename variable to IncludeNamespaces and the one below to ExcludeNamespaces.
Ideally, we should rename the `json:"namespaces"` to `json:"includeNamespaces"` as well.

platform/kube/inject/inject.go
```diff
@@ -460,6 +463,29 @@ func intoObject(c *Config, in interface{}) (interface{}, error) {
...
+		if obj.GetNamespace() == ignoredNamespace {
+			// If the namesepace was defined in both IgnoredNamespaces and Namespaces,
+			// then IgnoredNamespaces will overwrite the Namespaces.
+			if c.Namespaces[0] == "" {
+				glog.V(2).Infof("The namespace %s will be ignored by sidecar initializer", ignoredNamespace)
+			} else {
+				for _, namespace := range c.Namespaces {
+					if ignoredNamespace == namespace {
+						glog.V(2).Infof("The namespace %s will be ignored by sidecar initializer", ignoredNamespace)
+						break
```
rshriram: This is unnecessary and will cause confusion. Its better to have a mutually exclusive model.
User can specify either IncludeNamespaces or ExcludeNamespaces, but not both, where the semantics are
if ExcludeNamespaces is provided, sidecars will be injected on ALL except the namespaces provided..
if IncludeNamespaces is provided, sidecars will be injected on ONLY the namespaces provided.

platform/kube/inject/inject_test.go
```diff
@@ -152,6 +152,14 @@ func TestIntoResourceFile(t *testing.T) {
...
 			want: "testdata/replicationcontroller.yaml.injected",
 		},
+		{
+			in:   "testdata/hello-ibm.yaml",
+			want: "testdata/hello-ibm.yaml.inject",
+		},
+		{
+			in:   "testdata/hello-google.yaml",
+			want: "testdata/hello-google.yaml.inject",
+		},
```
rshriram: why do you need two ? They both seem to be excluding ibm-system namespace. Might as well just have one.

platform/kube/inject/inject.go
gyliu513: Sure, let's wait some comments from @ayj  ;-)

platform/kube/inject/inject.go
linsun: I'm concerned on the renaming...  our docs and 0.2.7 is out, I'd prefer not rename the existing `namespaces` support in a point release. I also agree with @gyliu513 that we need to support namespaces: [""] and a few excluded namespace.  This design was discussed in details in https://github.com/istio/istio.github.io/issues/648

rshriram: You don't have to necessarily rename the json field. Just rename the variables internally at the very least. And see my comment below.. the excludedNamespaces option supports the use case you mentioned.

platform/kube/inject/inject.go
gyliu513: Done

platform/kube/inject/inject_test.go
gyliu513: My original thinking is that i want to test the cases where the excludeNamespaces have multiple values, but yes, have one value is good enough.

platform/kube/inject/inject.go
gyliu513: OK, I created another patch here https://github.com/istio/pilot/pull/1437 , and will close this one.

gyliu513:
Please check the new PR here https://github.com/istio/pilot/pull/1437 

/cc @linsun @rshriram @ayj

PR #1437, opened 2017-10-11
Added new parameter to exclude inject namespaces.

### LINKED ISSUE #1433: Add  ignoredNamespaces to sidecar initializer configMap

gyliu513:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1433

**Release note**:

```release-note
Added new parameter to exclude inject namespaces.
```

/cc @linsun @rshriram @ayj


platform/kube/inject/inject.go
```diff
@@ -460,6 +467,17 @@ func intoObject(c *Config, in interface{}) (interface{}, error) {
 		}
 	}
 
+	// skip customized exclude namespaces
+	for _, excludeNamespace := range c.ExcludeNamespaces {
+		if excludeNamespace == "" {
+			return out, nil
+		}
```
rshriram: why do you need this check? is it for the case where someone specifies
`excludeNamespaces: [""]` ? Because that is as good as not running the initializer. Might as well disallow that option right?

platform/kube/inject/inject.go
```diff
@@ -460,6 +467,17 @@ func intoObject(c *Config, in interface{}) (interface{}, error) {
 		}
 	}
 
+	// skip customized exclude namespaces
+	for _, excludeNamespace := range c.ExcludeNamespaces {
```
rshriram: don't you need to check for a non-nil ? like c.ExcludeNamespaces != nil ?

platform/kube/inject/inject.go
```diff
@@ -190,6 +193,10 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
 		return nil, err
 	}
 
+	if len(c.IncludeNamespaces) >= 1 && len(c.ExcludeNamespaces) >= 1 {
+		return nil, fmt.Errorf("Cannot configure both namespaces and excludeNamespaces")
```
linsun: This needs to be updated if we decided to use IncludedNamespaces

platform/kube/inject/inject_test.go
```diff
@@ -161,8 +165,9 @@ func TestIntoResourceFile(t *testing.T) {
 		}
 
 		config := &Config{
-			Policy:     InjectionPolicyEnabled,
-			Namespaces: []string{v1.NamespaceAll},
+			Policy:            InjectionPolicyEnabled,
+			IncludeNamespaces: []string{v1.NamespaceAll},
```
ayj: This combination is disallowed by code you added [here](https://github.com/istio/pilot/pull/1437/files#diff-3f71ca8a1c3836bc4a3360d83741f2fbR196).

platform/kube/inject/inject.go
ayj: I don't think this is what you want. Further down in `inject_test.go` you specify `IncludeNamespaces: NamespaceAll` and `ExcludeNamespaces: "ibm-system"` which this check disallows. You either need to add an excemption for `NamespaceAll` or define precedence for `include` vs. `exclude` to coexist. Either way will need more tests.

rshriram: my point is do we need include and exclude to coexist? this is like what we did with the prepare_proxy.sh script.. We can pick either include or exclude. Both allow you to achieve the same purpose..With exclude, you will be able to say do not inject initializers for ibm-system, istio-system, kube-system, but inject for everything else. (we simplify this further by automatically excludign istio-system and kube-system). 

With include, you say, do not inject for anything other than these x,y,z namespaces.. Having both include and exclude leads to confusion and more mechanisms like precedence, etc. Its not like we support regex etc. in the namespace. Even then, I don't see a use case where you want to have both include and exclude in same configuration.

platform/kube/inject/inject.go
gyliu513: Done

platform/kube/inject/inject.go
linsun: I agree it is confusing if we name it to be IncludedNamespace & ExcludedNamespace.   I would much prefer to have the current namespaces and adding IgnoredNamespace (or ExcludedNamespace) which overwrite what is in the namespaces... as a lot of times, I would not know the namespaces ahead of time as an admin of the cluster and I would only exclude certain system ones that admin have in control.

platform/kube/inject/inject_test.go
gyliu513: Done

platform/kube/inject/inject.go
ayj: >  You either need to add an exemption for NamespaceAll or ...

My point is the code as-is won't work. A namespace (or set of namespaces) needs to be set, otherwise nothing will be injected. [Initializer.initialize](https://github.com/istio/pilot/blob/8b06fc3a77efd6a78f7bdb285a4f3c00cef810b1/platform/kube/inject/initializer.go#L172) will end up skipping the namespae and the [Wath/List](https://github.com/istio/pilot/blob/master/platform/kube/inject/initializer.go#L114) code won't watch any namespaces.

Simplifying the logic here is fine, but additional code changes are needed to watch/handle `NamespaceAll` when excludeNamespace is used.

platform/kube/inject/inject.go
```diff
@@ -190,6 +193,20 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
 		return nil, err
 	}
 
+	if c.IncludeNamespaces != nil && c.ExcludeNamespaces != nil {
+		return nil, fmt.Errorf("Cannot configure both namespaces and excludeNamespaces")
```
ayj: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

s/Cannot/cannot

platform/kube/inject/inject.go
```diff
@@ -190,6 +193,20 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
...
+	}
+
+	if c.IncludeNamespaces == nil {
+		c.IncludeNamespaces = []string{""}
+	}
+
+	for _, excludeNamespace := range c.ExcludeNamespaces {
+		if excludeNamespace == "" {
+			return nil, fmt.Errorf("Cannot configure ExcludeNamespaces as NamespaceAll")
```
ayj: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

s/Cannot/cannot

platform/kube/inject/inject.go
```diff
@@ -190,6 +193,20 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
 		return nil, err
 	}
 
+	if c.IncludeNamespaces != nil && c.ExcludeNamespaces != nil {
+		return nil, fmt.Errorf("Cannot configure both namespaces and excludeNamespaces")
+	}
+
+	if c.IncludeNamespaces == nil {
+		c.IncludeNamespaces = []string{""}
```
ayj: Please use `v1.NamespaceAll` instead of `""` here and throughout the k8s specific initializer code.

platform/kube/inject/inject.go
```diff
@@ -190,6 +193,20 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
 		return nil, err
 	}
 
+	if c.IncludeNamespaces != nil && c.ExcludeNamespaces != nil {
```
ayj: Please add additional test cases to TestGetInitializerConfig() for these new error cases

platform/kube/inject/inject.go
gyliu513: Done

platform/kube/inject/inject.go
gyliu513: Done

platform/kube/inject/inject.go
gyliu513: Done

platform/kube/inject/inject.go
gyliu513: Done

platform/kube/inject/inject.go
```diff
@@ -190,6 +193,20 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
...
+		return nil, fmt.Errorf("cannot configure both namespaces and excludeNamespaces")
+	}
+
+	if c.IncludeNamespaces == nil {
+		c.IncludeNamespaces = []string{v1.NamespaceAll}
+	}
+
+	for _, excludeNamespace := range c.ExcludeNamespaces {
+		if excludeNamespace == "" {
+			return nil, fmt.Errorf("cannot configure ExcludeNamespaces as NamespaceAll")
```
linsun: Why can't?  for example, I don't want any namespace to be injected.

platform/kube/inject/inject.go
```diff
@@ -190,6 +193,20 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
 		return nil, err
 	}
 
+	if c.IncludeNamespaces != nil && c.ExcludeNamespaces != nil {
+		return nil, fmt.Errorf("cannot configure both namespaces and excludeNamespaces")
```
linsun: error log needs to be updated to includeNamespaces

platform/kube/inject/inject.go
```diff
@@ -190,6 +193,20 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
...
+	}
+
+	if c.IncludeNamespaces == nil {
+		c.IncludeNamespaces = []string{v1.NamespaceAll}
```
linsun: would this change the output of describe the configmap?  ` kubectl describe configmap istio-inject  -n istio-system`

linsun: I found it is odd we don't allow user to set includeNamespaces but we are setting it ourselves... just wanted to make sure it is not surfaced to the user when user views the configmap.

platform/kube/inject/inject.go
gyliu513: @ayj @rshriram any comments for this? According to @rshriram 's comments above, setting `excludeNamespace` as "" is same as not start initializer, so we do not need to support this value but just let user do not bother to start the initializer is good enough. 

I think @rshriram 's comments is good enough by not supporting setting `excludeNamespace` as "" , comments? @linsun

platform/kube/inject/inject.go
gyliu513: `kubectl describe configmap istio-inject -n istio-system` only show resources in the yaml template.

I think that we need to update the document by clarify that if both `namespaces` and `excludeNamespaces` were not set, initializer will inject data to all namespaces.

Comments @linsun @ayj @rshriram

platform/kube/inject/inject.go
gyliu513: I want the log show key word in the configuration file, and we are still using `namespace` in configuration file to keep backward compatibility.

gyliu513:
If both `namespaces` and `excludeNamespaces` were not set, initializer will inject data to all namespaces. 

The test case was already handled in this commit: https://github.com/istio/pilot/pull/1437/commits/0d29969d436cabf71ec1aeb4f578cb8c0e28e2b9 , I was adding `IncludeNamespaces: []string{v1.NamespaceAll},` if both `namespaces` and `excludeNamespaces` were not set. 

@linsun

platform/kube/inject/inject.go
ayj: You can use `namespace=v1.NamespaceAll` with `policy=off` or `policy=disabled` to achieve the same effect.

platform/kube/inject/inject.go
ayj: `if c.IncludeNamespaces == nil {}` is only needed to make injection work when `excludeNamespaces` is non-empty. You could add `if c.IncludeNamespaces == nil && c.ExcludeNamespaces == nil {}` to error out when user doesn't specify a namespace.

gyliu513: done

andraxylia:
The included and excluded config is nice to have, but does not help the developer who wants control of her namespace. We still need a namespace annotation #1463

gyliu513:
Thanks @andraxylia I will work for this in 0.3 cycle.

platform/kube/inject/inject.go
```diff
@@ -190,6 +193,24 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
 		return nil, err
 	}
 
+	if c.IncludeNamespaces == nil && c.ExcludeNamespaces == nil {
+		return nil, fmt.Errorf("cannot leave both namespaces and excludeNamespaces")
```
linsun: maybe reword it a bit to indicate what actions user needs to do, such as please either configure namespaces or excludeNamespaces.

platform/kube/inject/inject.go
linsun: ok sounds good.  Would be good to have this documented.  You may want to change the error message to indicate the action user needs to do, e.g. namespace=v1.NamespaceAll with policy=off or uninstall initializer completely.

platform/kube/inject/inject.go
ayj: ... or we default to `c.IncludeNamespaces = []string{v1.NamespaceAll}` if nothing is set.

platform/kube/inject/inject.go
```diff
@@ -190,6 +193,24 @@ func GetInitializerConfig(kube kubernetes.Interface, namespace, injectConfigName
...
+	if c.IncludeNamespaces != nil && c.ExcludeNamespaces != nil {
+		return nil, fmt.Errorf("cannot configure both namespaces and excludeNamespaces")
+	}
+
+	if c.IncludeNamespaces == nil {
+		c.IncludeNamespaces = []string{v1.NamespaceAll}
+	}
+
+	for _, excludeNamespace := range c.ExcludeNamespaces {
+		if excludeNamespace == "" {
```
ayj: s/`""`/`v1.NamespaceAll`

platform/kube/inject/inject.go
```diff
@@ -460,6 +481,17 @@ func intoObject(c *Config, in interface{}) (interface{}, error) {
 		}
 	}
 
+	// skip customized exclude namespaces
+	for _, excludeNamespace := range c.ExcludeNamespaces {
+		if excludeNamespace == "" {
```
ayj: s/`""`/`v1.NamespaceAll`

platform/kube/inject/inject.go
ayj: This shouldn't be needed anymore with the check on line 208?

platform/kube/inject/inject.go
gyliu513: Yes, this seems more simple

platform/kube/inject/inject.go
gyliu513: Done

platform/kube/inject/inject.go
gyliu513: I was removing the check here as we already guard that the excludeNamsepace cannot be allNamespaces when get initialize.

PR #1440, opened 2017-10-11
fix for istio-initializer needs to ignore hostNetwork pod specs

linsun:

PR for https://github.com/istio/istio.github.io/issues/655, see #655 for details

**Release note**:

```release-note
When a pod uses hostNetwork: true, the pod will be disabled from side car injection on purpose because we don't want the envoy side car to change the network configuration at the host level.
```


linsun:
Thank you Jason and Shriram!

ldemailly:
`istioctl kube-inject` is immune to this somehow ? or is it a problem for istioctl in 0.2.7 ?

ayj:
The problem is specific to automatic injection. The user can be selective with `kube-inject` and avoid injecting the sidecar into pods with host networking.

PR #1443, opened 2017-10-11
Per-service auth enablement via service annotation

diemtvu:
This PR implements the first step to support fine-grained authentication enablement, i.e provide opt-in / opt-out options via service annotation. For more detail, see the [design doc](https://docs.google.com/document/d/1D7wZCQjVB72Wlwr5ZxP5WUmn3FUDr-XzfX8OodPXe8Y/edit?ts=59dbdea8#)

This PR partially addresses [this issue](https://github.com/istio/auth/issues/267). More works are needed to (auto) disable mTLS for connection with control services, handle health check, and smooth traffic migration.

**Release note**:

```NONE
```


model/service.go
```diff
@@ -77,6 +77,11 @@ type Port struct {
 
 	// Protocol to be used for the port.
 	Protocol Protocol `json:"protocol,omitempty"`
+
+	// In combine with the mesh's AuthPolicy, controls authentication for
+	// Envoy-to-Envoy communication.
+	// This value is extracted from service annotation.
+	SecurityOption SecurityOption `json:"security_option,omitempty"`
```
rshriram: Do we need an option here? A mesh has either opt-in or opt-out.. no point in having this for per port. I suggest using a simple boolean  like `Secure bool`.

model/service.go
```diff
@@ -107,6 +112,19 @@ const (
...
+	// Enable authentication. This overrules mesh's AuthPolicy.
+	SecurityEnable SecurityOption = "enable"
+
+	// Disable authentication. This overrules mesh's AuthPolicy.
+	SecurityDisable SecurityOption = "disable"
+
+	// Set authentication as specified by mesh's AuthPolicy.
+	SecurityDefault SecurityOption = ""
+)
+
```
rshriram: This should move into platform/kube .. These annotations are platform specific. Its responsibility of platform adapter to enable/disable the boolean under the port. After that, you shouldn't have to worry.

proxy/envoy/config.go
```diff
@@ -354,10 +354,13 @@ func buildHTTPListener(mesh *proxyconfig.MeshConfig, node proxy.Node, instances
 	}
 }
 
-func applyInboundAuth(listener *Listener, mesh *proxyconfig.MeshConfig) {
-	switch mesh.AuthPolicy {
-	case proxyconfig.MeshConfig_NONE:
-	case proxyconfig.MeshConfig_MUTUAL_TLS:
+func shouldApplyAuth(mesh *proxyconfig.MeshConfig, security_option model.SecurityOption) bool {
+	return security_option == model.SecurityEnable ||
+		(mesh.AuthPolicy == proxyconfig.MeshConfig_MUTUAL_TLS && security_option != model.SecurityDisable)
```
rshriram: You don't need the first or condition.. and the second part of the and is just a boolean check

proxy/envoy/config.go
```diff
@@ -711,15 +717,14 @@ func buildInboundListeners(mesh *proxyconfig.MeshConfig, sidecar proxy.Node,
...
 
-			listeners = append(listeners, listener)
-
 		default:
 			glog.V(4).Infof("Unsupported inbound protocol %v for port %#v", protocol, servicePort)
 		}
-	}
 
-	for _, listener := range listeners {
-		applyInboundAuth(listener, mesh)
```
rshriram: just pass the endpoint to the applyInboundAuth function and do the checks there

proxy/envoy/config.go
```diff
@@ -354,10 +354,13 @@ func buildHTTPListener(mesh *proxyconfig.MeshConfig, node proxy.Node, instances
...
-	switch mesh.AuthPolicy {
-	case proxyconfig.MeshConfig_NONE:
-	case proxyconfig.MeshConfig_MUTUAL_TLS:
+func shouldApplyAuth(mesh *proxyconfig.MeshConfig, security_option model.SecurityOption) bool {
+	return security_option == model.SecurityEnable ||
+		(mesh.AuthPolicy == proxyconfig.MeshConfig_MUTUAL_TLS && security_option != model.SecurityDisable)
+}
+
+func mayApplyInboundAuth(listener *Listener, mesh *proxyconfig.MeshConfig, security_option model.SecurityOption) {
+	if shouldApplyAuth(mesh, security_option) {
```
rshriram: remove

proxy/envoy/config.go
```diff
@@ -354,10 +354,13 @@ func buildHTTPListener(mesh *proxyconfig.MeshConfig, node proxy.Node, instances
 	}
 }
 
-func applyInboundAuth(listener *Listener, mesh *proxyconfig.MeshConfig) {
-	switch mesh.AuthPolicy {
-	case proxyconfig.MeshConfig_NONE:
-	case proxyconfig.MeshConfig_MUTUAL_TLS:
+func shouldApplyAuth(mesh *proxyconfig.MeshConfig, security_option model.SecurityOption) bool {
```
rshriram: remove new function

proxy/envoy/config.go
```diff
@@ -354,10 +354,13 @@ func buildHTTPListener(mesh *proxyconfig.MeshConfig, node proxy.Node, instances
...
+	return security_option == model.SecurityEnable ||
+		(mesh.AuthPolicy == proxyconfig.MeshConfig_MUTUAL_TLS && security_option != model.SecurityDisable)
+}
+
+func mayApplyInboundAuth(listener *Listener, mesh *proxyconfig.MeshConfig, security_option model.SecurityOption) {
+	if shouldApplyAuth(mesh, security_option) {
 		listener.SSLContext = buildListenerSSLContext(proxy.AuthCertsPath)
```
rshriram: pass the endpoint to the buildListenerSSLContext. and in that function, if endpoint.Port has secure==false, just return nil. If you are entering buildListenerSSLContext, then you know that global auth was enabled.

platform/kube/conversion.go
```diff
@@ -53,11 +55,29 @@ func convertLabels(obj meta_v1.ObjectMeta) model.Labels {
...
+		return model.SecurityDefault
+	}
+	switch obj.Annotations[portSecurityAnnotationKey(int(port.Port))] {
+	case "disable":
+		return model.SecurityDisable
+	case "enable":
+		return model.SecurityEnable
+	default:
+		return model.SecurityDefault
+	}
```
rshriram: Can't you just make the decision to enable/disable auth here? Read the annotation, and then decide to enable/disable based on the global auth policy.

proxy/envoy/testdata/lds-v0-egress-rule-auth-optin.json.golden
```diff
@@ -0,0 +1,658 @@
+{
+  "listeners": [
+   {
```
rshriram: kill. We don't need to update all golden files. You just need two files, to showcase optin and optout.

model/service.go
diemtvu: I think we do. The point of this is to overwrite the (global) mesh's policy. A service can have more than one port, and we want to have a fine grained control for each of them.

diemtvu: Also, boolean may not be enough. We want a 'state' that indicates to use mesh's policy to decide. Furthermore, it's possible to add other options in future (e.g regular TLS instead of mTLS, etc)

model/service.go
diemtvu: As mentioned above, boolean is too limit. This is more or less an enum of all options.

platform/kube/conversion.go
diemtvu: Possible, but I'm not sure it's the right place to do. It's easier this way to extend to more options later.

model/service.go
rshriram: Make sense. Then just change it to AuthenticationPolicy

proxy/envoy/config.go
rshriram: Please do this though. You don't need to restructure or create new functions. Just pass the endpoint object to the applyInboundAuth and then take the decision/skip auth there.

diemtvu: I'm not sure if I understand your suggestion correctly. The reason I split the applyInboundAuth function is to reuse shouldApplyAuth in policy.go (for clusters). Moving the call to this applyInboundAuth from the separate "for loop" to the main loop is to make it easier to pass endpoint.

proxy/envoy/testdata/lds-v0-egress-rule-auth-optin.json.golden
diemtvu: Done. I add separated test for optin and optout so just need 2 golden files now.

platform/kube/conversion.go
```diff
@@ -53,11 +55,29 @@ func convertLabels(obj meta_v1.ObjectMeta) model.Labels {
 	return out
 }
 
-func convertPort(port v1.ServicePort) *model.Port {
+// Extracts security option for given port from annotation. If there is no such
+// annotation, or the annotation value is not recognized, returns
+// model.SecurityDefault.
+func extractAuthenticationPolicy(port v1.ServicePort, obj meta_v1.ObjectMeta) model.AuthenticationPolicy {
```
rshriram: Can you also extract from namespace level? It might become cumbersome to annotate every service's port. And for system managed namespaces / administrative namespaces (e.g., kube-system, calico-system, ibm-system), we need an exception mechanism for the entire namespace.

diemtvu: From user point of view, I think we need this fine grained control (e.g they may need to turn of TLS for a some service which already have TLS, eg. DB). Ask them to move the job to different namespace for that purpose seems unnatural.

We may added a config at namespace level later, if there are some demands. Or at least add "wild-card" to this to mach to all ports.

proxy/envoy/config.go
```diff
@@ -742,15 +747,14 @@ func buildInboundListeners(mesh *proxyconfig.MeshConfig, sidecar proxy.Node,
...
 			glog.V(4).Infof("Unsupported inbound protocol %v for port %#v", protocol, servicePort)
 		}
-	}
 
-	for _, listener := range listeners {
-		applyInboundAuth(listener, mesh)
+		if listener != nil {
+			mayApplyInboundAuth(listener, mesh, endpoint.ServicePort.AuthenticationPolicy)
+			listeners = append(listeners, listener)
+		}
```
rshriram: Can you validate that this works for headless services? In our tests, we are disabling auth for headless services because it doesn't work. If you could fix that as well, while you are here, it would be great.

proxy/envoy/discovery_test.go
```diff
@@ -157,7 +157,18 @@ func TestClusterDiscoveryCircuitBreaker(t *testing.T) {
 	compareResponse(response, "testdata/cds-circuit-breaker.json", t)
 }
 
-func TestClusterDiscoveryWithSSLContext(t *testing.T) {
+func TestClusterDiscoveryWithSecurityOptin(t *testing.T) {
```
rshriram: AuthOptIn

proxy/envoy/discovery_test.go
```diff
@@ -168,6 +179,24 @@ func TestClusterDiscoveryWithSSLContext(t *testing.T) {
 	compareResponse(response, "testdata/cds-ssl-context.json", t)
 }
 
+func TestClusterDiscoveryWithSecurityOptout(t *testing.T) {
```
rshriram: AuthOptOut

proxy/envoy/discovery_test.go
```diff
@@ -421,6 +451,33 @@ func TestListenerDiscoverySidecar(t *testing.T) {
 	}
 }
 
+func TestListenerDiscoverySidecarAuthOptin(t *testing.T) {
```
rshriram: AuthOptIn

proxy/envoy/discovery_test.go
```diff
@@ -421,6 +451,33 @@ func TestListenerDiscoverySidecar(t *testing.T) {
...
+	// Auth opt-in on port 80
+	mock.HelloService.Ports[0].AuthenticationPolicy = model.AuthenticationEnable
+	ds := makeDiscoveryService(t, registry, &mesh)
+	url := fmt.Sprintf("/v1/listeners/%s/%s", "istio-proxy", mock.HelloProxyV0.ServiceNode())
+	response := makeDiscoveryRequest(ds, "GET", url, t)
+	compareResponse(response, "testdata/lds-v0-none-auth-optin.json", t)
+	mock.HelloService.Ports[0].AuthenticationPolicy = model.AuthenticationDefault
+}
+
+func TestListenerDiscoverySidecarAuthOptout(t *testing.T) {
```
rshriram: AuthOptOut (capital O)

proxy/envoy/testdata/cds-ssl-context-optin.json.golden
```diff
@@ -0,0 +1,206 @@
...
+    "name": "out.e5c9564b7c4dbb0355a4f740e9d29277ccca97cd",
+    "service_name": "hello.default.svc.cluster.local|http",
+    "connect_timeout_ms": 1000,
+    "type": "sds",
+    "lb_type": "round_robin",
+    "ssl_context": {
+     "cert_chain_file": "/etc/certs/cert-chain.pem",
+     "private_key_file": "/etc/certs/key.pem",
+     "ca_cert_file": "/etc/certs/root-cert.pem",
+     "verify_subject_alt_name": []
```
rshriram: why is this empty?

proxy/envoy/testdata/cds-ssl-context-optout.json.golden
```diff
@@ -0,0 +1,286 @@
...
+    "hosts": [
+     {
+      "url": "tcp://istio-egress.istio-system:8888"
+     }
+    ],
+    "ssl_context": {
+     "cert_chain_file": "/etc/certs/cert-chain.pem",
+     "private_key_file": "/etc/certs/key.pem",
+     "ca_cert_file": "/etc/certs/root-cert.pem",
+     "verify_subject_alt_name": []
```
rshriram: why is this empty?

proxy/envoy/testdata/lds-v0-none-auth-optout.json.golden
```diff
@@ -0,0 +1,624 @@
...
+       ],
+       "access_log": [
+        {
+         "path": "/dev/stdout"
+        }
+       ]
+      }
+     }
+    ],
+    "bind_to_port": false
```
rshriram: where is the SSL context for this listener?

proxy/envoy/testdata/lds-v0-none-auth-optout.json.golden
```diff
@@ -0,0 +1,624 @@
...
```
rshriram: Same. Missing SSL context, since this is a OptOut file.

proxy/envoy/testdata/lds-v0-none-auth-optout.json.golden
diemtvu: This listener is for outgoing traffic right? ssl_context is added for inbound listener only.  This golden file is branched from lds-v0-none-auth. FYI, here is the diff:

diff proxy/envoy/testdata/lds-v0-none-auth-optout.json.golden proxy/envoy/testdata/lds-v0-none-auth.json.golden 

```
523a524,529
>     "ssl_context": {
>      "cert_chain_file": "/etc/certs/cert-chain.pem",
>      "private_key_file": "/etc/certs/key.pem",
>      "ca_cert_file": "/etc/certs/root-cert.pem",
>      "require_client_certificate": true
>     },
624c630
<  }
---
>  }
\ No newline at end of file
```

proxy/envoy/testdata/lds-v0-none-auth-optout.json.golden
diemtvu: Same as above.

proxy/envoy/discovery_test.go
diemtvu: Done

proxy/envoy/discovery_test.go
diemtvu: Done

proxy/envoy/discovery_test.go
diemtvu: Done

proxy/envoy/discovery_test.go
diemtvu: Done

proxy/envoy/testdata/cds-ssl-context-optin.json.golden
diemtvu: I guess the mock setup doesn't have service account for this service. 
https://github.com/istio/pilot/blob/master/test/mock/service.go#L257

This ssl_context is the same as the one in the existing proxy/envoy/testdata/cds-ssl-context.json.golden

proxy/envoy/testdata/cds-ssl-context-optout.json.golden
diemtvu: Same as above

proxy/envoy/config.go
diemtvu: Not quite sure what does it involve :). May be different PR?

proxy/envoy/config.go
```diff
@@ -368,10 +368,14 @@ func buildHTTPListener(mesh *proxyconfig.MeshConfig, node proxy.Node, instances
 	}
 }
 
-func applyInboundAuth(listener *Listener, mesh *proxyconfig.MeshConfig) {
-	switch mesh.AuthPolicy {
-	case proxyconfig.MeshConfig_NONE:
-	case proxyconfig.MeshConfig_MUTUAL_TLS:
+func shouldApplyAuth(mesh *proxyconfig.MeshConfig, authenticationPolicy model.AuthenticationPolicy) bool {
+	return authenticationPolicy == model.AuthenticationEnable ||
```
wattli: Might worth a little comment here.

diemtvu:
Is this PR good to be submitted? I know we still have discussion of how to do traffic migration properly, but I think having this option is still valuable at this point for early testing. What do you think?

model/service.go
```diff
@@ -107,6 +112,23 @@ const (
 	ProtocolRedis Protocol = "Redis"
 )
 
+// AuthenticationPolicy defines authentication policy for port.
+type AuthenticationPolicy string
```
kyessenov: Suggestion: extend AuthPolicy proto definition https://github.com/istio/api/blob/master/proxy/v1/config/proxy_mesh.proto#L21 with a default value. Make the default be the default. And then explain the overriding principle in the API proto definition: per-service settings takes precedence over the per-mesh.

platform/kube/conversion.go
andraxylia: We will definitely need to annotate hierarchically across istio, for auth, the automatic injection, and anything else.

But it is ok to do it incrementally and start with services.

platform/kube/conversion.go
```diff
@@ -53,11 +57,29 @@ func convertLabels(obj meta_v1.ObjectMeta) model.Labels {
...
+	if obj.Annotations == nil {
+		return model.AuthenticationDefault
+	}
+	switch obj.Annotations[portAuthenticationAnnotationKey(int(port.Port))] {
+	case "disable":
```
andraxylia: can you use some const instead of harcoded strings?

andraxylia:
Do you plan to add integration or e2e tests?

platform/kube/conversion.go
diemtvu: This is mapping a text to a constant. Do you think it's easier to follow this way?

model/service.go
diemtvu: Do you suggest extend API enum and use it here? Do we want to do it now or it can wait?

proxy/envoy/config.go
costinm: +1 that needs to be fixed - but smaller PRs are easier to review and test.

costinm:
+1 on an integration test - I think simplest option that will also work with the 'stable environments' (weekly long running tests, etc) is to just add one more service to the 'auth disabled' suite, with opt-in, and make sure it works and it has auth enabled.

model/service.go
costinm: I think a separate PR would be far better, no need to make this too complex.

model/service.go
```diff
@@ -107,6 +112,23 @@ const (
...
+
+const (
+	// AuthenticationEnable indicates authentication should be enable (on the
+	// associated port). This overrules mesh's AuthPolicy.
+	AuthenticationEnable AuthenticationPolicy = "enable"
```
costinm: It may be worth having an 'API review' focused on consistency. 
The global AuthPolicy uses 'none' and mutualtls' - could we use the same here ? In future we may add other options 
(like 'auto' ), but we should be consistent in terms across istio.

platform/kube/conversion.go
```diff
@@ -43,6 +43,10 @@ const (
 
 	// IstioURIPrefix is the URI prefix in the Istio service account scheme
 	IstioURIPrefix = "spiffe"
+
+	// PortAuthenticationAnnotationKeyPrefix is the annotation key prefix that used to define
+	// authentication policy.
+	PortAuthenticationAnnotationKeyPrefix = "tls.istio.io"
```
costinm: How about authpolicy.istio.io  - TLS is just one protocol for authentication. Maybe we'll use JWT, or QUIC (which is UDP and not real TLS), or ipsec transport - the setting is to specify the policy or config for encryption, not to customize tls.

platform/kube/conversion.go
costinm: We need per-service for sure and first - so individual services can be moved incrementally. 
We do need namespace level as well - but I think it is better to keep PRs small and have incremental changes.

kyessenov:
Please provide integration tests. It'd help to also spell out testing plan. This is a new feature, and we are trying to stabilize existing feature set, so the baseline for adding new functionality is aimed at production readiness.

diemtvu:
e2e test added. Please take another look.

rshriram:
Can you explain what this e2e is doing? I only see the negative tests.

diemtvu:
I've added service "d" for the e2e test. This service has one http port annotated enabled and one disabled auth, and one tcp, one grcp port enabled. The tests show those ports are able to serve request from envoy-client in all cases, but cannot (or can, for the disable case) receive request from "t" (legacy client), regardless of the mesh's policy.

test/integration/testdata/app.yaml.tmpl
```diff
@@ -5,6 +5,13 @@ metadata:
   name: {{.service}}
   labels:
     app: {{.service}}
+{{if eq .perServiceAuth "true"}}
+  annotations:
+    tls.istio.io/80: enable
```
xiaolanz: I strongly recommend to declare per service TLS through a Istio CRD instead of pushing it through annotations. Istio is building it's own config, grinding istio features down into k8s semantics makes it hard to layer them in future.

Issue #1007, opened 2017-08-08
Secure connection between discovery and proxy/agent


kyessenov:
Envoy uses plaintext HTTP to speak with xDS. Agent also fetches ingress secrets on the same HTTP endpoint. Both should be secured.

cc @costinm

PR #1444, opened 2017-10-11
[WIP] Securing istio components pilot and mixer

rkpagadala:
This PR is for securing communication with istio components.

#1007 
 Putting this out for review and comments.

**Release note**:

```release-note
```


cmd/pilot-agent/main.go
```diff
@@ -220,6 +247,9 @@ func init() {
 		"IP Address and Port of a statsd UDP listener (e.g. 10.75.241.127:9125)")
 	proxyCmd.PersistentFlags().IntVar(&proxyAdminPort, "proxyAdminPort", int(values.ProxyAdminPort),
 		"Port on which Envoy should listen for administrative commands")
+	proxyCmd.PersistentFlags().IntVar(&infraAuthPolicy, "infraAuthPolicy", int(values.InfraAuthPolicy),
```
costinm: How about a boolean flag 'allowInsecure' ( to make it clear what it does - it's not really a policy ).

cmd/pilot-agent/main.go
```diff
@@ -118,6 +123,28 @@ var (
...
+						if len(parts) > 0 && len(parts[0]) > 0 {
+							ns = parts[0]
+						} else {
+							ns = os.Getenv("POD_NAMESPACE")
+						}
+					} else {
+						ns = os.Getenv("POD_NAMESPACE")
+					}
+				}
+				proxyConfig.PilotSan = envoy.GetPilotSAN(pilotDomain, ns)
```
costinm: Few comments ? I assume you want the service account of pilot ? Can we have a default value (matching what we have in the default config ), and a flag to override ?

cmd/pilot-agent/main.go
costinm: Per discussion on 'upgradeability' and backward compat, I think it is important to start with both old port (8080) and new secure port on the server, and 'allowInsecure' set to true. 

This will allow us to test, compare performance, etc - and will follow the model of gradual adoption.

proxy/envoy/testdata/envoy-sidecar-auth.json.golden
```diff
@@ -0,0 +1,120 @@
...
+          {
+            "url": "tcp://localhost:15003"
+          }
+        ],
+        "ssl_context": {
+          "cert_chain_file": "/etc/certs/cert-chain.pem",
+          "private_key_file": "/etc/certs/key.pem",
+          "ca_cert_file": "/etc/certs/root-cert.pem",
+          "verify_subject_alt_name": [
+            "spiffe://cluster.local/ns/istio-system-auth/sa/istio-pilot-service-account"
```
costinm: What is this namespace ? We should test with the ns where the default install is using (istio-system)

rshriram:
@rkpagadala please cleanup the PR in istio/apis. I proposed some changes, unfortunately, I couldn't comment before the PR got merged. Those are smallish ones, that improve some usability aspects and involves some API changes. That will have an impact on this PR.

rkpagadala:
>  The config is 100% static.
I still need to pick up Mixer address and Pilot Discovery address from Mesh config and write the appropriate envoy startup config, a 100% static does not work. There is no ip table magic here, so envoy listen port has  to match the configured Mixer/Pilot address

Specifically in file envoy-mixer.json.golden 15004 comes from Mixer Address
 +      "address": "tcp://0.0.0.0:15004",
 +      "name": "tcp_0.0.0.0_15004",

In file envoy-pilot.json.golden 15003 comes from Discovery Address
 +      "address": "tcp://0.0.0.0:15003",
 +      "name": "tcp_0.0.0.0_15003",

> - it is totally unclear how you will capture and load balance external services that mixer adapters rely on.

I am not trapping outgoing traffic at all, it just goes straight through, a it does right now

rshriram:
On Thu, Oct 12, 2017 at 1:29 AM Krishna Pagadala <notifications@github.com>
wrote:

> The config is 100% static.
> I still need to pick up Mixer address and Pilot Discovery address from
> Mesh config and write the appropriate envoy startup config, a 100% static
> does not work. There is no ip table magic here, so envoy listen port has to
> match the configured Mixer/Pilot address
>
>
Why ? Just listen on 0.0.0.0:9080 and 9443, with http. You certainly don’t
need pilot address.

This will work definitely in kubernetes. For other platforms, you have much
bigger problems with multiple network interfaces, lack of pods, etc.

costinm:
That was my first reaction - and I still agree (for pilot). And I still
have the 'use hand-made custom envoy config' option on my TODO list.

For mixer - it will need to make outbound calls to other components, for
plugins. Those in turn will need sidecars and mTLS.
However it is not clear why mixer can't use the normal sidecar config - all
it needs is an 'override' for finding the mixer.
Having a way to override the mixer address on a per-pod or per-service
basis would be a critical feature for hybrid/expansions, so
this feature will be super useful in future.

CA is already using custom TLS - so will need the new feature that auth
team is developing for turning off mTLS selectively. Need
to double check with the auth team.

rshriram:
The mixer external access story has not been sorted out. If the current use case is to just enforce auth and auth only, (may be some free metrics from envoy), then it is dead simple integration with a http connection manager config, pointer to a statsd server. The only piece of code required is a switch on the proxy agent to decide whether a sidecar should use SSL config for its pilot cluster.

On top of it all, it would be good to have some idea of the performance hit and potential issues as we move to gRPC on pilot as well. I am more concerned about the mixer envoy, given that we already suffer some serious performance issues when mixer is enabled (network latency costs). 

Secondly, istio does not imply sidecars. So, if auth is the main goal at this moment, then it is a much more simple process to directly add TLS support in Pilot, that will read from the certs and use them. After all, we are also thinking of library integration for Istio, aren't we? the mixer is designed to be called by services directly as well. If anything, I would just focus all energy there. Peppering Pilot and Mixer with plenty of stats (on cache hits/misses, k8s sync issues, etc.), and streaming all these stats to statsd or prometheus. External access can also be sorted out by explicitly using Envoy (this is how Lyft operates 100s of different services. I am sure we can operate Pilot and Mixer like this, if we decide to go down the sidecar route). 

But in my mind, Istio is about uniform telemetry, auth, load balancing/policy. The rest is implementation details. The latter two don't apply for pilot and mixer. The former two do not need a sidecar. The metrics you get from Envoy for Pilot calls will not be of much use, because a lot of K8S specific stuff, caching metrics, etc. are in Pilot. Same applies to Mixer. These cannot be gathered by Mixer filter either.

Once again, all this was pointed out in the doc in comments, with no response (https://docs.google.com/document/d/1YzYPddihbLgJhme27-md9COn0NnKUKX_xCJ_GPXU4Fw/edit?ts=598b4be1).

cmd/pilot-agent/main.go
wattli: It would be really great if we could add more comments here. The code review UI is not very user-friendly.

cmd/pilot-agent/main.go
wattli: Maybe allowInsecurePolicy

proxy/envoy/infra_auth.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+)
+
+const (
+	mixerSvcAccName string = "istio-mixer-service-account"
+	pilotSvcAccName string = "istio-pilot-service-account"
+)
+
+func GetMixerSAN(domain, ns string) []string {
```
wattli: Please add some comment and state that these are hard-coded and the reason behind it. THanks.

proxy/envoy/mixer.go
```diff
@@ -88,6 +89,16 @@ func buildMixerCluster(mesh *proxyconfig.MeshConfig) *Cluster {
 		},
 	}
 	mixerCluster.Features = ClusterFeatureHTTP2
+
+        // apply auth policies 
+        switch mesh.DefaultConfig.InfraAuthPolicy {
+        case proxyconfig.AuthPolicy_NONE:
+                // do nothing
+        case proxyconfig.AuthPolicy_MUTUAL_TLS:
+                // apply SSL context to enable mutual TLS between Envoy proxies for mixer
```
wattli: "and mixer"

proxy/envoy/mixer.go
```diff
@@ -79,7 +79,8 @@ type FilterMixerConfig struct {
 
 func (*FilterMixerConfig) isNetworkFilterConfig() {}
 
-func buildMixerCluster(mesh *proxyconfig.MeshConfig) *Cluster {
+// buildMixerCluster builds an outbound mixer cluster
```
wattli: I think mixer should take inbound traffic, i.e., proxy calls mixer, not the other way around, right?

proxy/envoy/testdata/envoy-mixer-auth.json.golden
```diff
@@ -0,0 +1,51 @@
...
+                {
+                  "cluster": "in.9091"
+                }
+              ]
+            }
+          }
+        }
+      ],
+      "ssl_context": {
+        "cert_chain_file": "/etc/certs/cert-chain.pem",
```
wattli: This is what i am talking about in previous comment. It is the listener where we put the ssl_context, not cluster.

platform/kube/inject/testdata/auth.cert-dir.yaml.injected
```diff
@@ -41,7 +41,7 @@ spec:
         - --parentShutdownDuration
         - 3s
         - --discoveryAddress
-        - istio-pilot:8080
+        - istio-pilot:15003
```
wattli: any reason for this change?

kyessenov:
I think we should start with a static configuration template for mixer/pilot sidecars. The template can reside in pilot for now, but can also move to istio once docker packaging goes there.

We only need certification rotation, the rest is almost entirely fixed. The port mapping is an implementation detail that can be resolved in the installation templates. The rest of Istio only needs to know 1) address of pilot service (fronted by envoy or not); 2) the service account for pilot. 
The first item is already available in the mesh config. We can add the second as a flag to the agents or to the mesh config if it needs to be dynamic at some point.

I also think we should apply http filter, not just TCP passthrough filter.

rkpagadala:
Thanks everybody for your review. Closing this will send out a new PR

PR #1453, opened 2017-10-12
fix for istio-initializer needs to ignore hostNetwork pod specs (#1440)

linsun:
Automatic merge from submit-queue

fix for istio-initializer needs to ignore hostNetwork pod specs

PR for https://github.com/istio/istio.github.io/issues/655, see #655 for details

**Release note**:

```release-note
When a pod uses hostNetwork: true, the pod will be disabled from side car injection on purpose because we don't want the envoy side car to change the network configuration at the host level.
```

This is the same PR but for 0.2 release branch.


rshriram:
Will let @ayj review.

PR #1476, opened 2017-10-13
bugfix: remove initializer when specific namespaces are set

### LINKED ISSUE #1470: CRITICAL: Istio initializer causing timeouts in kube API server
@dcberg commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660)

I have a situation with the 0.2.7 istio-initializer where I see kube API server timeouts when deploying pods. It appears that the problem happens only in a namespace that is NOT having the sidecar injected. This is very important point. 

Here are the steps to recreate the problem.

1) Install istio-initializer as normal (I installed using the instructions from the 0.2.7 release).
2) Create `istio-demo` namespace.
`kubectl create ns istio-demo`
3) Edit the istio-inject configmap to set a single namespace to enable istio injection.
`kubectl edit istio-inject -n istio-system`
Add `istio-demo` as the injected namespace.
```
apiVersion: v1
data:
  config: |-
    policy: "enabled"
    namespaces: ["istio-demo"] # everything, aka v1.NamepsaceAll, aka cluster-wide
    initializerName: "sidecar.initializer.istio.io"
    params:
      initImage: docker.io/istio/proxy_init:0.2.7
      proxyImage: docker.io/istio/proxy:0.2.7
      verbosity: 2
      version: 0.2.7
      meshConfigMapName: istio
      imagePullPolicy: IfNotPresent
kind: ConfigMap
```
4) Save istio-inject configmap.
5) Delete istio-initializer pod to pick up configmap changes.
`kubectl delete pod -n istio-system $(kubectl get pod -n istio-system -l istio=initializer -o jsonpath='{.items[0]..metadata.name}')`
6) Deploy a simple nginx pod in the `istio-demo` namespace. All works fine.
```
kubectl run nginx --image=nginx --replicas=2 -n istio-demo
deployment "nginx" created
```
7) Now try to deploy the same simple nginx pod in the default namespace. You will see a timeout this time.
```
kubectl run nginx --image=nginx --replicas=2
Error from server (Timeout): Timeout: request did not complete within allowed duration
```

The timeout issue is only present when operating within a namespace that is NOT included in the istio injection.

Other information....
```
kubectl version
Client Version: version.Info{Major:"1", Minor:"7", GitVersion:"v1.7.6", GitCommit:"4bc5e7f9a6c25dc4c03d4d656f2cefd21540e28c", GitTreeState:"clean", BuildDate:"2017-09-15T08:51:09Z", GoVersion:"go1.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"7+", GitVersion:"v1.7.4-1+1540c973d4ff9d", GitCommit:"1540c973d4ff9da2d6a204a7709084488a459ed4", GitTreeState:"clean", BuildDate:"2017-10-06T15:08:43Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}
```

```
kubectl apiversions
Command "apiversions" is deprecated, use "api-versions" instead
admissionregistration.k8s.io/v1alpha1
apiextensions.k8s.io/v1beta1
apiregistration.k8s.io/v1beta1
apps/v1beta1
authentication.k8s.io/v1
authentication.k8s.io/v1beta1
authorization.k8s.io/v1
authorization.k8s.io/v1beta1
autoscaling/v1
batch/v1
batch/v2alpha1
certificates.k8s.io/v1beta1
config.istio.io/v1alpha2
extensions/v1beta1
networking.k8s.io/v1
policy/v1beta1
rbac.authorization.k8s.io/v1alpha1
rbac.authorization.k8s.io/v1beta1
settings.k8s.io/v1alpha1
storage.k8s.io/v1
storage.k8s.io/v1beta1
v1
```

---

@linsun commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336241528)

@ayj this is the issue I chatted with you about yesterday.  Dan sorted out the scenario where it times out on any other namespaces not specified.  Are you aware of anything in initializer or istio-initializer that may cause the namespace not functional?  I'll try recreate this on my system too.

---

@dcberg commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336244517)

Note....if I delete the istio-initializer deployment then the timeout in the default namespace no longer fails.

---

@ayj commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336245553)

This was by design to allow multiple initializers managing different sets of namespaces to co-exist in the same cluster. This was necessary given our shared testing infrastructure. Any namespace that is not managed by the initializer (i.e. included in the `namespace` list) is _completely_ ignored by the initializer.

https://github.com/istio/pilot/pull/1437 and `excludeNamespace` could possibly be enhanced to support your use case.

> Note....if I delete the istio-initializer deployment then the timeout in the default namespace no longer fails

Did you delete just the k8s deployment or the entire initializer configuration (i.e. kubectl delete -f istio-initializer.yaml)? 

---

@dcberg commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336246511)

I used `kubectl delete -f istio-initializer.yaml`

If I only include a single namespace for istio to operate against then I would expect no impact when operating against resources in other namespaces. However, this is not the case. I literally cannot use any other namespace in when the initializer is configured against a subset of namespaces for the cluster. Clearly there is a problem with the istio-initializer when operating against excluded namespaces.

---

@linsun commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336248902)

I agree, if we can enhance excludeNamespace for this scenario, then we can handle this in istio/pilot#1437.  cc @gyliu513 

As a user, @dcberg can specify namespaces to be included for initializer then everything else will be excluded by the initializer by default.

---

@dcberg commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336271643)

@linsun I agree that excludeNamespace will be useful but I don't see how that will correct this problem. As a user, if I only want istio applied to a single namespace then I would simply set the one namespace to be used and I would not use `excludeNamespace`. In this scenario all other namespaces in my cluster will be broken due to the istio-initializer.

---

@gyliu513 commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336329128)

@dcberg Please check here https://github.com/gyliu513/jay-work/blob/master/k8s/example/kube-initializer-tutorial/medium-article.md#troubleshooting for how to troubleshooting such issue.

It was caused by that the initializers was added to your deployment, will check more for this how to fix.

```
initializers:
      pending:
      - name: sidecar.initializer.istio.io
```

---

@linsun commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336330066)

@gyliu513 thanks, that link is super helpful and seems exactly the issue that Dan hit.

@dcberg the reason we think the excludeNamespace will help is that we'll treat all other namespaces that is not specified in the namespaces array as excluded from initialization, similar as how we treat the istio-system, kube-system and kube-public namespaces in 0.2.7.

---

@gyliu513 commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336331184)

@ayj this seems to be a serious issue, the current `initializerConfiguration` is cluster level, even if we exclude some namespaces, there will still be `pending initializers` be injected to the deployment and will cause all deployment cannot work. 

I think it is a kubernetes initializer design issue, we may need to file bugs to kubernetes and work there for a solution, comments?

---

@mandarjog commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336331382)

@ayj the default behaviour seems to be tailored to our test infra use case. Perhaps we should add a bool option shared_cluster or similar that defaults to "false", so initializer will assume it is the only one running unless specified otherwise.

---

@linsun commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336334003)

@gyliu513 I'm thinking the namespaces aren't properly excluded because I don't hit such behavior in istio-system namespace, as I deployed grafana, prometheus and zipkin after I installed the initializer.

@mandarjog I agree it is pretty dangerous to design the system around the shared test infra.  I just referenced this issue to the #internal_notice for general awareness.

---

@gyliu513 commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336334129)

@dcberg @linsun @ayj as a workaround for such issue, you may want to set initializers as [] for your deployment as following:

```
metadata:
  initializers:
    pending: []
```

This is an example:
```
[root@ib17b03 ~]# cat gyliu/nginx.yaml
kind: Deployment
metadata:
  labels:
    run: my-nginx
  name: my-nginx
  namespace: default
  initializers:
      pending: []
spec:
  # replicas: 6
  selector:
    matchLabels:
      run: my-nginx
  template:
    metadata:
      labels:
        run: my-nginx
    spec:
      containers:
      - image: nginx:1.8.1
        imagePullPolicy: IfNotPresent
        name: my-nginx
        ports:
        - containerPort: 80
          protocol: TCP
```

I also clarified the behavior here https://github.com/gyliu513/jay-work/blob/master/k8s/example/kube-initializer-tutorial/medium-article.md#deploy-the-sidecar-initializer

---

@dcberg commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336336372)

@linsun note that if you treat excludeNamespaces the same way that kube-system and istio-system are excluded then you are going to have the same exact problem. I confirmed this by trying to deploy the same nginx image in the `istio-system` namespace which is ignored. I get the same failure as I expected. Ignored namespaces have the same problem. The initializer is still being added since it is global but the namespace is ignored leaving the initializer in a pending state.

```
kind: Deployment
metadata:
  creationTimestamp: 2017-10-13T02:49:32Z
  generation: 1
  initializers:
    pending:
    - name: sidecar.initializer.istio.io
  labels:
    run: nginx
  name: nginx
  namespace: istio-system
  resourceVersion: "62356"
  selfLink: /apis/extensions/v1beta1/namespaces/istio-system/deployments/nginx
  uid: 2405a77e-afc1-11e7-948d-96812b1a68e5
spec:
  replicas: 2
  selector:
    matchLabels:
      run: nginx
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 1
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        run: nginx
    spec:
      containers:
      - image: nginx
        imagePullPolicy: Always
        name: nginx
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
status: {}
```

---

@gyliu513 commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336336787)

@linsun I think that the reason you can deploy prometheus success is because we already set `namespaces: [""]` here https://github.com/istio/istio/blob/master/install/kubernetes/istio-initializer.yaml#L14 and also ignoring `inject data` for prometheus deployment  https://github.com/istio/istio/blob/master/install/kubernetes/addons/prometheus.yaml#L59-L60

The `namespaces: [""]` can make sure all namespaces will be covered the initializer, so there will be chance for initializer to clear the pending initializers.

But if we set namespaces as a specified value such as `namespaces: ["ibm-system"]` etc, the pending initializers will still be added to all of the kubernetes resources (deployment etc), but the initializer will have no chance to remove the pending initializers.

To be a quick solution without kubernetes fix, we may need to revisit the `sidecar initializer` code to add logic of remove the pending initializers for resources which was excluded from `excludeNamespaces` or not included in `Namespaces`.

/cc @ayj 

---

@dcberg commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336337073)

@gyliu513 I'm sure that the workaround will work but it isn't sufficient. I can work around the problem by only using namespaces that are included in the istio-initializers namespaces. The problem is that global won't work because some namespaces have to be ignored. As soon as even one namespace is ignored which is by default, you can hit this problem (see previous comment).

---

@dcberg commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336337476)

@gyliu513 your suggestion seems reasonable. Basically the initializer logic always runs to remove the pending initializer which is globally added. If the namespace is to be ignored then the pending initializer is removed with no action. Otherwise the pending initializer is removed and the sidecar is injected.

i.e., the initializer cannot be left in a pending state.

---

@gyliu513 commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336338547)

@dcberg right, so we may need to update the istio initializer to handle such logic. There are also some discussion with same direction as here in Kubernetes: https://github.com/kubernetes/kubernetes/issues/53859

I will try to see if I can create a PR this weekend, but it may not able to catch up with the 0.2.8 release.

/cc @ayj @linsun @rshriram can we merge https://github.com/istio/pilot/pull/1437 first so that I can continue work based on current logic.

---

@andraxylia commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336345473)

@dcberg Thanks for finding this bug. The opt-in/opt-out was a stretch goal for 0.2, so this issue is not critical, the workaround is to use the initializer with the global policy enable. This bug has to be fixed for 0.3 and for production readiness.

---

@andraxylia commented on [Thu Oct 12 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336347041)

We should probably start by removing this from the istio.io website, since it does not work. This issue is opened against istio.io website, which is good.

And right, the fix is for the initializer to remove itself from any workload, even if it performed no action. The includeNamespace / excludeNamespace are handled exclusively in the Istio code, there is no change required in Kubernetes.

I would opt for completely removing includeNamespace / excludeNamespace and using namespace annotation to override the cluster-wide policy, as described here istio/pilot#1463  (add code for that instead of fixing the logic for includeNamespaces). The namespaces are somewhat dynamic and under the user's control, and one cannot expect a regular user to perform this workflow: create a namespace, change the configMap owned by admin, and delete a pod owned by the admin using a fancy command, just to to get their services injected.

We do not need to support initializers on the shared cluster, we have a cluster wide test.

---

@linsun commented on [Fri Oct 13 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336454390)

> To be a quick solution without kubernetes fix, we may need to revisit the sidecar initializer code to add logic of remove the pending initializers for resources which was excluded from excludeNamespaces or not included in Namespaces.

@gyliu513 I like this suggestion... I think it is a reasonable small fix... which will basically add automatically for the user's deployment metadata similarly what is added manually for prometheus or zipkin deployment yaml... With the fix, we can even remove the inject data for prometheus or other istio-system deployment https://github.com/istio/istio/blob/master/install/kubernetes/addons/prometheus.yaml#L59-L60 as they should not be needed.  I'm off vacation today, I'd be happy to work on this on Monday if you need me to.

@andraxylia This is one of the most exciting features of 0.2... I think we should leave in istio.io as we already said it is an alpha feature and try fix it in 0.2.  It is much better to have such feedback in 0.2 than 0.3.  This feature is really cool and shows our longer term direction.

---

@ayj commented on [Fri Oct 13 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336505531)

> ... add logic of remove the pending initializers for resources which was excluded from excludeNamespaces 

Ignored known system namespaces already have this behavior - the pending initializer was removed without injection. https://github.com/istio/pilot/pull/1437 extends that behavior to user-specified namespaces via `excludeNamespace`. I don't think any other changes are needed as an immediate fix. 

---

@dcberg commented on [Fri Oct 13 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336510462)

@ayj is the removal of the pending initializer in an ignored system namespace part of the 0.2.7 release? I ask because I verified that the pending initializers were NOT being removed in the `istio-system` namespace as part of 0.2.7.

---

@dcberg commented on [Fri Oct 13 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336511334)

@andraxylia I am in full agreement that the initializer should move to use namespace annotations to over the global behavior.

---

@ayj commented on [Fri Oct 13 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336517450)

Yes, `istio-system` should be ignored in 0.2.7. I tried the following with the 0.2.7 release and see that a deployment created in the `istio-system` namespace is created successfully and do not have a sidecar injected. I created the same deployment in the default namespace and verified the sidecar was injected.

```bash
$ kubectl apply -f install/kubernetes/istio.yaml
$ kubectl apply -f install/kubernetes/istio-initializer.yaml
$ kubectl apply -f samples/sleep/sleep.yaml -n istio-system

$ kubectl -n istio-system get po/sleep-3432132412-ns1b9
NAME                     READY     STATUS    RESTARTS   AGE
sleep-3432132412-ns1b9   1/1       Running   0          14m

$ kubectl apply -f samples/sleep/sleep.yaml 
$ kubectl get pod
NAME                    READY     STATUS    RESTARTS   AGE
sleep-140275861-vzprz   2/2       Running   0          12m
```

I also removed the sidecar annotation from `install/kubernetes/addons/zipkin.yaml` and `install/kubernetes/addons/prometheus.yaml` and was able to create them successfull in the `istio-system` namespace.

---

@geeknoid commented on [Fri Oct 13 2017](https://github.com/istio/istio.github.io/issues/660#issuecomment-336521408)

Should this issue be moved to a different repo? It's not related to istio.github.io as far as I can tell.

ayj:

The initializer was not removing itself from the list of pending initializers when specific namespaces were specified in the configuration. This PR fixes that problem so that deployments created in ignored and excluded namespaces can be created properly.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1470 

**Release note**:

```release-note
NONE
```


platform/kube/inject/initializer.go
```diff
@@ -162,29 +162,38 @@ func NewInitializer(restConfig *rest.Config, config *Config, cl kubernetes.Inter
 	return i, nil
 }
 
-func (i *Initializer) initialize(in interface{}, patcher patcherFunc) error {
-	obj, err := meta.Accessor(in)
-	if err != nil {
-		return err
+func modifyRequired(objNamespace string, ignore, exclude, include []string) bool {
```
andraxylia: The purpose of this function is a bit confusing, because there is also the injectRequired, which already has control logic if to inject or not based on the namespaces include exclude etc.

I thought the initializer would need to remove itself from the list of initializers no matter what. So the logic would be like:
if injectRequired(c.Policy, obj) {
   inject()
} else {
    glog.V(2).Infof("Skipping %s/%s due to policy check", obj.GetNamespace(), obj.GetName())
}
removeFromPending()

ayj: Do we want to retain the option of running multiple initializers in the same cluster? If not, we can remove the "off" policy and make the initializer only operate cluster-wide. That would simplify some of this code.

If concurrent initializers is supported, then initializer needs to do one of three things: 

1) ignore a deployment completely (and let another initializef handle it)
2) remove itself from the list of pending initialiers (and do nothing; ignored/excluded namespaces)
3) (2) plus injection the sidecar 

`modifyRequired` checks if (2) is required.
`injectRequired` checks if (3) is required.

andraxylia: Got it, thanks for clarifying.

platform/kube/inject/initializer_test.go
```diff
@@ -237,3 +237,61 @@ func TestInitialize(t *testing.T) {
...
+			ignore:    []string{"istio-system"},
+			include:   []string{"istio-system"},
+			want:      true,
+		},
+		{
+			namespace: "default",
+			ignore:    []string{"istio-system"},
+			exclude:   []string{"default"},
+			want:      true,
+		},
```
linsun: Is this really true when default namespace is on he excluded list?

platform/kube/inject/initializer.go
```diff
@@ -162,29 +162,38 @@ func NewInitializer(restConfig *rest.Config, config *Config, cl kubernetes.Inter
...
+		if namespace == objNamespace {
+			return true
+		}
+	}
+	for _, namespace := range include {
 		if namespace == v1.NamespaceAll {
-			inject = true
-			break namespaceSearch
-		} else if namespace == obj.GetNamespace() {
+			return true
```
linsun: Should it also return true if it is on the included namespace list?  I asked because the PR discription also said excluded and ignored namespaces.

platform/kube/inject/initializer_test.go
linsun: I think this one makes sense... but I had qs is 2 tests below where namespace is foo-bar and foo-bar is on the included list.  Do we really want true there?

gyliu513:
Thanks @ayj for the update, but I do want to see more test cases as this:

In https://github.com/istio/pilot/blob/master/platform/kube/inject/inject_test.go#L58 , we have the function of `TestIntoResourceFileWithIncludeNamespaces`, but this test is using `IncludeNamespaces: []string{v1.NamespaceAll},`.

What I want to test is we can probably add a new test case named as `TestIntoResourceFileWithIncludeNotAllNamespaces`, then for this test cases, we can set `IncludeNamespaces: []string{"ibm-system"},` or the namespace as you want and do the test to see if deployments in other namespaces can work well. This will also introduce some new YAML templates to test, hope it helps.

platform/kube/inject/initializer.go
ayj: Yes, line 179 checks to see if the object's namespace matches the current value in the list of included namespaces.

platform/kube/inject/initializer_test.go
ayj: Yes, the resource should be modified if it's namespace is on the include list. Another way to look at it is the resource should _always_ be modified unless it's namespace is not on the include, exclude, or ignore lists.

ayj:
>  set IncludeNamespaces: []string{"ibm-system"}, or the namespace as you want and do the test to see if deployments in other namespaces can work well. 

What do you mean by "work well"? Other namespaces not in the included list will be completely ignored by the initializer and won't have the pending initializer cleared.

ayj:
FWIW, `includeNamespace` doesn't have any affect on `kube-inject` since the assumption is the user running the tool implies they want it injected. This fix is more important for the initializer case which runs unconditionally on all deployments.

gyliu513:
> What do you mean by "work well"? Other namespaces not in the included list will be completely ignored by the initializer and won't have the pending initializer cleared.

@ayj do you mean that the resources in other namespaces will still have the pending initializer created? If so, then those resources will still be pending, can you elaborate? Also the `work well` I mean is the resources created by kubernetes should not be pending but should be created successfully, which means that the pending initializers should be cleared.

ayj:
@gyliu513, I believe what you're suggesting is to

ayj:
@gyliu513, yes resources that aren't in `include`, `ignore`, or `exclude` namespace lists will still have their pending initializer. We could change this behavior and make the initializer truely cluster wide, but that is a change in design (see https://github.com/istio/pilot/pull/1476#discussion_r144669080). I'm fine making that change if others agree, but don't want to rush into it.

linsun:
> yes resources that aren't in include, ignore, or exclude namespace lists will still have their pending initializer. 

@ayj Here is a common scenario:  I configure the istio-initializer to exclude namespace X, and I'm deploying bookinfo to namespace Y.   Would auto side car injection work for this bookinfo deployment in namespace Y?  I would expect it to work.

linsun:
@ayj The other scenario is what @gyliu513 mentioned:   I want to configure istio-initializer to include namespace demo.  I can deploy bookinfo to namespace demo with side car auto injected.   If I deploy bookinfo to namespace Y, I should be able to deploy, without side car injected.   I think these two scenarios should work, and I don't see e-2-e injection yaml tests for them.  

My understanding from your comment is that the 2 scenarios won't work because the current design is working around the shared cluster issue.  I think we need to make istio initializer cluster wide because that is the scope people is expecting it to work, similar as istio itself.

gyliu513:
@ayj I think that this is a critical issue for istio initializer that we must fix, otherwise, it will cause some applications cannot deployed. We can merge this first.

ayj:
The codecov checks are producing false negatives. Manual verification gives the following results.

| file | before | after | delta |
| ---- | ---------| -------| ------ |
| platform/kube/inject/configmap.go | 100% | 100% | 0 |
| platform/kube/inject/http.go | 79.3% | 79.3% | 0 |
| platform/kube/inject/initializer.go | 73.9% | 75.3% | + 1.4% |
| platform/kube/inject/inject.go | 89.7% | 90.9% | + 0.4% |
| total | 84.2% | 85.2% | + 1% |

ayj:
> My understanding from your comment is that the 2 scenarios won't work because the current design is working around the shared cluster issue. I think we need to make istio initializer cluster wide because that is the scope people is expecting it to work, similar as istio itself.

That is correct. The current implementation does not allow for that. We need to make initializer cluster-wide to address the root problem.

yutongz:
@ayj yeah I think it's a bug from codecov, but it doesn't block merge.

ayj:
This PR has LGTM and `codecov/patch` and `codecov/project` are both X'd out for me. It looks like its blocking merge. Maybe I'm missing someone else?

yutongz:
You need a "approve" label, you can comment "/approve" to your own PR to get this label since your are an owner here.

PR #1480, opened 2017-10-14
make fmt update.

gyliu513:

**Release note**:

```release-note
NONE
```
/cc @rshriram

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -15,6 +15,7 @@ git_repository(
 load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_repository", "go_register_toolchains")
 
 go_rules_dependencies()
+
 go_register_toolchains()
 
 ##


WORKSPACE
```diff
@@ -15,6 +15,7 @@ git_repository(
 load("@io_bazel_rules_go//go:def.bzl", "go_rules_dependencies", "go_repository", "go_register_toolchains")
 
 go_rules_dependencies()
+
```
ldemailly: ?

gyliu513: I was just running `make fmt` and it helped adjust the format.

I think we should always try to `make fmt` first before submit a PR.

@ldemailly

ldemailly: I thought we had a  check in presubmit forcing the format already, maybe I am confusing with mixer repo.

Any idea about the smoke failure?

gyliu513: @ldemailly  I found that the mixer also do not have the `fmt` check there, can you correct me if I'm wrong?

Also I have just fixed an `make fmt` issue here https://github.com/istio/pilot/pull/1430 , so +1 to adding `fmt` check in presubmit.

Also can you please show more detail for how does the prow/pilot-presubmit.sh was triggered?

cmluciano: Found [fmt script](https://github.com/istio/test-infra/blob/master/scripts/fmt.sh) which seems to be doing this mainly on .go files. I'm not sure if there is anything touching WORKSPACE or Bazel build files.

I will look over it again.

gyliu513: @cmluciano thanks, also I want to do as this: Add some pre-check before submit PR, so that's why I want to make the presubmit script to have some a `fmt` check, but not sure how does the `prow/pilot-presubmit.sh` was triggered when submit patch, can you help?

cmluciano: Not sure I understand, do you want to run `go fmt` against the Bazel files? Is the buildifier lint not sufficient?

PR #1493, opened 2017-10-16
cleanups to readme, makefiles, etc.

rshriram:
**Release note**:
```release-note
NONE
```


Makefile
```diff
@@ -25,22 +42,30 @@ fmt:
...
 
 .PHONY: clean
 clean:
 	@bazel clean
 
 .PHONY: test
-test: build
-	@bazel test //...
+test:
+	@bazel test ${BUILD_FLAGS} //...
```
kyessenov: I wonder: does this work fine on native mac OS? Should we restrict k8 to e2e test only?

README.md
```diff
@@ -4,19 +4,64 @@
...
 
-If you are interested in contributing to the project, please take a look at the [build instructions](doc/build.md) and the [testing infrastructure](doc/testing.md).
+
+To learn how you can contribute to Istio Pilot, please
+see the Istio
+[contribution guidelines](https://github.com/istio/istio/blob/master/CONTRIBUTING.md).
+
+# Quick start
+
+1. *Install Bazel:* [Bazel 0.6.0](https://github.com/bazelbuild/bazel/releases/tag/0.6.0) or
```
kyessenov: 0.6.1 is the latest (some bug fixes). You can leave 0.6.0 here, I'm hoping 0.6.x series is somewhat backwards compatible.

bin/init.sh
```diff
@@ -5,6 +5,14 @@ set -o nounset
 set -o pipefail
 set -x
 
+detected_OS=`uname -s 2>/dev/null || echo not`
```
kyessenov: I think we don't really need the entire build here. `fetch` is sufficient. I can try to see whether that works later.

doc/testing.md
```diff
@@ -69,16 +69,7 @@ To run the complete set of tests, use the following commands:
...
-the kube config file on the host to platform/kube instead of symlinking it,
-and change the paths to minikube certs.
-
-    cp ~/.kube/config platform/kube/
-    sed -i 's!/Users/<username>!/home/ubuntu!' platform/kube/config
-
-Also, copy the same file to `/home/ubuntu/.kube/config` in the VM, and make
-sure that the file is readable to user `ubuntu`.
-
+_Note1_: If you are using OS X, append `--cpu=k8` to the bazel test command.
```
kyessenov: _ to build linux binaries _

Makefile
rshriram: Yeah, its a bit of pain when this test starts e2e as well. if we restrict k8s to e2e alone, then it makes instructions and stuff easier. I could probably even code up the refresh goldens here.

bin/init.sh
rshriram: comment on the wrong file? :)..

Makefile
rshriram: Actually, it works perfectly on Mac. Except the proxy/envoy test. And you don't even need the build flags. Updated PR

doc/testing.md
rshriram: No longer needed. bazel test //... will run unit tests on Mac. bin/e2e.sh will automatically compile for linux and push to docker hub.

rshriram: Don't you love shell scripts? ;)

PR #1538, opened 2017-10-18
Use the net packages to validate IP and CIDR correctness

cmluciano:
Use the Golang net pkg functions to validate the correctness of IP addresses and CIDRs

**Release note**:
```
NONE
```

### DIFF
--- a/model/validation.go
+++ b/model/validation.go
@@ -17,6 +17,7 @@ package model
 import (
 	"errors"
 	"fmt"
+	"net"
 	"regexp"
 	"strconv"
 	"strings"
@@ -467,22 +468,16 @@ func ValidateSubnet(subnet string) error {
 // validateCIDR checks that a string is in "CIDR notation"
 func validateCIDR(cidr string) error {
 	// We expect a string in "CIDR notation", i.e. a.b.c.d/xx form
-	parts := strings.Split(cidr, "/")
-	if len(parts) != 2 {
-		return fmt.Errorf("%q is not valid CIDR notation", cidr)
+	ip, _, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return fmt.Errorf("%v is not a valid CIDR block", cidr)
+	}
+	// The current implementation only supports IP v4 addresses
+	if ip.To4() == nil {
+		return fmt.Errorf("%v is not a valid IPv4 address", cidr)
 	}
 
-	var errs error
-
-	if err := ValidateCIDRBlock(parts[1]); err != nil {
-		errs = multierror.Append(errs, err)
-	}
-
-	if err := ValidateIPv4Address(parts[0]); err != nil {
-		errs = multierror.Append(errs, err)
-	}
-
-	return errs
+	return nil
 }
 
 // ValidateIPv4Subnet checks that a string is in "CIDR notation" or "Dot-decimal notation"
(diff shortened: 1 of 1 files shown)


model/validation.go
```diff
@@ -506,17 +507,10 @@ func ValidateCIDRBlock(cidr string) error {
 
 // ValidateIPv4Address validates that a string in "CIDR notation" or "Dot-decimal notation"
 func ValidateIPv4Address(addr string) error {
-	octets := strings.Split(addr, ".")
-	if len(octets) != 4 {
+	err := net.ParseIP(addr)
```
ayj: s/`err`/`ip`/. The return value is of type `net.IP` and not error.

Also, we should check if `ip` is IPv4 if existing semantics are to be preserved. Otherwise we'll allow IPv6.

model/validation.go
```diff
@@ -474,7 +475,7 @@ func validateCIDR(cidr string) error {
 
 	var errs error
 
-	if err := ValidateCIDRBlock(parts[1]); err != nil {
+	if err := ValidateCIDRBlock(cidr); err != nil {
```
ayj: `net.ParseCIDR` calls `net.ParseIP`. Depending on whether we want to support IPv6 or not you could invoke `net.PraseCIDR` directly here and remove `validateCIDRBlock`

cmluciano: Ack, I'm not sure if we specifically call out IPv6 support in Istio. It isn't 100% ready in the k8s sense, but it may be released in the alpha state for 1.9.

model/validation.go
cmluciano: Is it necessary to also validate the subnet mask? I noted that we have a separate function for it and the test case here https://github.com/cmluciano/pilot/blob/c9ce6417fbac824ad505e6f07f48cab88b6e9690/model/validation_test.go#L481 does not make sense. 

We specify a parameter for SourceSubnet: []string{"1.2.3.4"} but we do not mark the subnet mask explicitly with proper CIDR notation. Is this test case a subtly bug or do we not care about passing the explicit mask and only care if the IP address is IPv4?

ayj: The subnet is optional (see [here](https://github.com/istio/api/blob/93848bb1c495ff6ebc1a938e2e0a403d52d345c1/proxy/v1/config/route_rule.proto#L269)). `ValidateIPv4Subnet` validates both cases.

cmluciano: Ok, I can migrate the ValidateIPv4Subnet to use Go net libraries for doing more of this validation in this PR as well. WDYT?

ayj: sgtm

model/validation.go
```diff
@@ -467,22 +468,16 @@ func ValidateSubnet(subnet string) error {
...
-		return fmt.Errorf("%q is not valid CIDR notation", cidr)
-	}
-
-	var errs error
-
-	if err := ValidateCIDRBlock(parts[1]); err != nil {
-		errs = multierror.Append(errs, err)
+	ip, _, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return fmt.Errorf("/%v is not a valid CIDR block", cidr)
```
ayj: Why the `/` as the first character in the error messages here and on line 477?

cmluciano: mistype

PR #1567, opened 2017-10-20
CircleCI integration

rshriram:
Compile, and run e2e tests on CircleCI for Kubernetes. Will automate this to add support for Consul and Eureka as well. This will help others use a familiar tool, and not rely on opaque and provider-specific test infrastructure.

These tests will run on every PR, but will not block any PR. Its essentially the equivalent of pilot presubmit.

**Release note**:

```release-note
NONE
```


.circleci/config.yml
```diff
@@ -0,0 +1,36 @@
...
+          command: |
+            JSONPATH='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}'; until sudo kubectl get nodes -o jsonpath="$JSONPATH" 2>&1 | grep -q "Ready=True"; do sleep 1; done
+      - run:
+          command: |
+            sudo -E kubectl cluster-info
+      - run:
+          command: |
+            ln -s ~/.kube/config platform/kube/
+      - run: bazel build //...
+      #- run: bazel test //...
```
ijsnellf: this meant to be left out?

.circleci/config.yml
```diff
@@ -0,0 +1,36 @@
...
+      - run: make docker HUB=docker.io/rshriram TAG=test
```
ijsnellf: i assume we'll migrate away from your username at some point?

.circleci/config.yml
rshriram: Yes. Because one of the admission controller tests fail with this.. Also circle CI is not replacing anything (yet)

.circleci/config.yml
rshriram: Never!

rshriram: Also, we don't push to dockerhub. This is 100% local. So, any username will do.

rshriram:
To clarify, the circle CI setup will not be pushing anything to dockerhub. It mimicks a test setup on a VM, deploys localkube on a VM, runs the full set of pilot integration tests.

louiscryan:
@sebastienvas FYI

.circleci/config.yml
sebastienvas: Could you just call one script instead of writing it on the yaml file ? That way people can also call it locally.

PR #1569, opened 2017-10-20
Enable unit tests for circle CI

rshriram:
@ayj FYI. Thanks for the pointer to our testing docs for minikube options.
```release-note
NONE
```


ayj:
I didn't look too closely at the original CircleCI PR. Was there an issue/doc with motivation, long term plans, etc. for this approach?

rshriram:
Its just something for non google folks. Having a commonly understood test infrastructure, and something that others can debug/view easily. Its fast, and low overhead as well, as you can see from the PR (just one circle ci config), and easily reproducible locally. Basically, a very boring test infrastructure :).

rshriram:
And its a backup. The nice thing about this is its completely isolated. One VM per test. We don't have to worry about resources leaking, or underlying infrastructure running out of IPs, etc.

Issue #1501, opened 2017-10-16
handling service registry client errors


GregHanson:
The current interface which the various service discovery clients implement is the following:

```
type ServiceDiscovery interface {
	Services() []*Service
	GetService(hostname string) (*Service, bool)
	Instances(hostname string, ports []string, labels LabelsCollection) []*ServiceInstance
	HostInstances(addrs map[string]bool) []*ServiceInstance
	ManagementPorts(addr string) PortList
}
```

pilot/proxy/envoy/discovery.go calls this interface and updates its cache accordingly.  With the current interface, if the client experiences an error calling the consul/eureka/etc server, it will currently return an empty array, which discovery.go will updates its cache with and return the new empty result to the xDS API's

current suggestions would be to either:

1. add a return error for the interface methods so discovery.go ignores the empty array and returns what it has in the cache

2. caching or better/consistent error handling on the client side model.ServiceDiscovery implementations

Thoughts?

rshriram:
Lets start with 1. And possibly, along the way, split the handlers for consul, eureka, k8s, such that only one of the caches is cleared when service handlers are called, instead of clearing all caches.

There are several low hanging fruits in the cache and we need to have more finer grained cache invalidation.

kyessenov:
The client of the registry already does caching (the discovery service that ultimately constructs the config). So adding another caching to the model adaptor seems redundant. Perhaps, returning an error indicator for "retrying later" is a good option then. I agree that returning an empty array on error is not a good behavior.

GregHanson:
resolved by #1610

PR #1610, opened 2017-10-24
Discovery client errors

GregHanson:
modifed model.ServiceDiscovery interface to return errors to prevent cache clears if client experiences an error talking to the Discovery server.  the xDS API's return a 503 in this case to tell envoy to keep what it currently has cached and and try again later

issue #1501 

**Release note**:

```release-note
NONE
```


proxy/envoy/discovery.go
```diff
@@ -416,7 +438,13 @@ func (ds *DiscoveryService) ListClusters(request *restful.Request, response *res
 			return
 		}
 
-		clusters := buildClusters(ds.Environment, role)
+		clusters, err := buildClusters(ds.Environment, role)
+		if err != nil {
+			// If client experiences an error, 503 error will tell envoy to keep its current
+			// cache and try again later
+			errorResponse(response, http.StatusServiceUnavailable, "CDS "+err.Error())
```
rshriram: have you verified this behavior?

GregHanson: asked the question on the envoy-users google group 
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/envoy-users/s9ka4-au9LY/QzS4ZSyHIQAJ

adapter/serviceregistry/aggregate/controller.go
```diff
@@ -47,22 +48,36 @@ func (c *Controller) AddRegistry(registry Registry) {
...
-			return service, true
+		service, err := r.GetService(hostname)
+		if err != nil {
+			errs = multierror.Append(errs, err)
+		}
+		if service != nil {
+			if errs != nil {
+				glog.Warningf("GetService() found match but encountered an error: %v", errs)
+			}
+			return service, nil
```
andraxylia: Why return nil instead of the error?

adapter/serviceregistry/aggregate/controller.go
```diff
@@ -79,23 +94,45 @@ func (c *Controller) ManagementPorts(addr string) model.PortList {
...
+		var err error
+		instances, err = r.Instances(hostname, ports, labels)
+		if err != nil {
+			errs = multierror.Append(errs, err)
+		}
+		if len(instances) > 0 {
+			if errs != nil {
+				glog.Warningf("Instances() found match but encountered an error: %v", errs)
+			}
+			return instances, nil
```
andraxylia: Same here.

adapter/serviceregistry/aggregate/controller.go
GregHanson: It is a design decision that is up for debate.  Of the N registries, at least one of them experienced an error - but one of the registries found the service that was requested.  Does one broken registry effectively break the aggregator?

andraxylia: What you describe is not the logic I see in the code. What I see is the situation when service was found in a registry, but an error was reported by the same registry. For this, we should return it, it is up to the caller of this API to decide if the service is valid (I would say it's never valid).

For the N registries: I am for returning the service if found and nil error, only if the error from the registry where the service was found is nil, otherwise you have to return the error as well.  But the current code does not distinguish between the case where the error was from previous iterations, or from the current iteration. You need to re-arrange the ifs logic.

Please add the comment to explain the above decision, and log the errors that would otherwise be swallowed.

rshriram:
@GregHanson can you investigate why this is failing on circle ci? Its minikube, and runs directly on linux (no vms). Tests are failing for auth

adapter/serviceregistry/aggregate/controller.go
```diff
@@ -79,23 +94,45 @@ func (c *Controller) ManagementPorts(addr string) model.PortList {
...
 		}
 	}
-	return instances
+	return instances, errs
 }
 
 // HostInstances lists service instances for a given set of IPv4 addresses.
-func (c *Controller) HostInstances(addrs map[string]bool) []*model.ServiceInstance {
+func (c *Controller) HostInstances(addrs map[string]bool) ([]*model.ServiceInstance, error) {
 	instances := make([]*model.ServiceInstance, 0)
```
ijsnellf: maybe switch this to `out` so the variable names are clearer than `insts` and `instances`? same comment applies to the rest of the aggregate controller

platform/consul/controller.go
```diff
@@ -44,53 +44,59 @@ func NewController(addr, datacenter string, interval time.Duration) (*Controller
...
+	return convertService(endpoints), nil
 }
 
-func (c *Controller) getServices() map[string][]string {
+func (c *Controller) getServices() (map[string][]string, error) {
 	data, _, err := c.client.Catalog().Services(nil)
 	if err != nil {
 		glog.Warningf("Could not retrieve services from consul: %v", err)
-		return make(map[string][]string)
+		return make(map[string][]string), err
```
ijsnellf: `return nil, err`

platform/consul/controller.go
```diff
@@ -44,53 +44,59 @@ func NewController(addr, datacenter string, interval time.Duration) (*Controller
...
+	return data, nil
 }
 
-func (c *Controller) getCatalogService(name string, q *api.QueryOptions) []*api.CatalogService {
+func (c *Controller) getCatalogService(name string, q *api.QueryOptions) ([]*api.CatalogService, error) {
 	endpoints, _, err := c.client.Catalog().Service(name, "", q)
 	if err != nil {
 		glog.Warningf("Could not retrieve service catalogue from consul: %v", err)
-		return []*api.CatalogService{}
+		return []*api.CatalogService{}, err
```
ijsnellf: `return nil, err`

platform/kube/controller.go
```diff
@@ -263,23 +263,24 @@ func (c *Controller) ManagementPorts(addr string) model.PortList {
...
 	if !exists {
-		return nil
+		return nil, nil
 	}
 
 	// Locate all ports in the actual service
 	svc := convertService(*item, c.domainSuffix)
 	if svc == nil {
-		return nil
+		// TODO is this an error case?
```
ijsnellf: remove comment. it just filtering

proxy/envoy/config.go
```diff
@@ -128,34 +129,56 @@ func buildConfig(config proxyconfig.ProxyConfig, pilotSAN []string) *Config {
...
 
 // buildListeners produces a list of listeners and referenced clusters for all proxies
-func buildListeners(env proxy.Environment, node proxy.Node) Listeners {
+func buildListeners(env proxy.Environment, node proxy.Node) (Listeners, error) {
 	switch node.Type {
 	case proxy.Sidecar, proxy.Router:
-		instances := env.HostInstances(map[string]bool{node.IPAddress: true})
+		instances, err := env.HostInstances(map[string]bool{node.IPAddress: true})
+		if err != nil {
+			return Listeners{}, err
```
ijsnellf: `return nil, err`

proxy/envoy/config.go
```diff
@@ -128,34 +129,56 @@ func buildConfig(config proxyconfig.ProxyConfig, pilotSAN []string) *Config {
...
+		}
+		services, err := env.Services()
+		if err != nil {
+			return Listeners{}, err
```
ijsnellf: same etc

PR #1638, opened 2017-10-27
let getDefaultNamespace() returns "default" if nothing is found.

### LINKED ISSUE #1637: fails to create mixer configs when namespace field is empty
derived from https://github.com/istio/issues/issues/110, the same discussion can be applied to the mixer config, and that is not solved. Indeed, some mixer configs in the current bookinfo samples are in this situation, thus:

```
% istio-0.2.9/bin/istioctl create -f istio/samples/bookinfo/kube/mixer-rule-deny-label.yaml
Error: 3 errors occurred:

* an empty namespace may not be set during creation
* an empty namespace may not be set during creation
* an empty namespace may not be set during creation
```

jmuk:
https://github.com/istio/issues/issues/110 reminds me that istioctl
can set empty namespace for mixer configs, which will be delivered
directly and therefore it can cause same error of "empty namespace".

When the obtained namespace happens to be empty, I think it should
return the default value.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1637

**Release note**:

```release-note
none
```

### DIFF
--- a/cmd/istioctl/main.go
+++ b/cmd/istioctl/main.go
@@ -697,7 +697,11 @@ func getDefaultNamespace(kubeconfig string) string {
 		return v1.NamespaceDefault
 	}
 
-	return config.Contexts[config.CurrentContext].Namespace
+	namespace := config.Contexts[config.CurrentContext].Namespace
+	if namespace == "" {
+		return v1.NamespaceDefault
+	}
+	return namespace
 }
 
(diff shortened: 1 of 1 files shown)


andraxylia:
@jmuk could you please cherry-pick this into 0.2? Our request routing instructions no longer work.

PR #1644, opened 2017-10-27
Distribute new protobuf representation of mixerclient configuration

ayj:

1) Generate mixerclient protobuf code. This will likely be simplified
(or go away) once we start storing generated artifacts as part of repo
consolidation.

2) Distribute the new protobuf mixerclient configuration as part of
the existing mixer filter configuration. The protobuf is encoded as a
JSON map value under "v2" key in the filter config struct map
alongside the legacy ad-hoc configuration. This is a backwards
compatible change to allow proxies to incrementally adopt the new
configuration. The legacy ad-hoc config is deprecated and will
eventually be removed.

3) annotate inbound route's opaque_config with the destination.service
of the server instance. Mixerclient can use this to look-up
per-service quota/api in the filter configuration.

NOTE #1): this depends on https://github.com/istio/api/pull/220 for corresponding protobuf changes to make gogoslick_proto_library happy.

NOTE #2): posted for early review. I'll rebase+merge after repo consolidation is finished.

**Release note**:

```release-note
```


proxy/envoy/mixer.go
```diff
@@ -125,20 +147,52 @@ func mixerHTTPRouteConfig(role proxy.Node, service string) *FilterMixerConfig {
...
 		MixerAttributes: map[string]string{
 			AttrDestinationIP:  role.IPAddress,
 			AttrDestinationUID: "kubernetes://" + role.ID,
 		},
-		DisableTCPCheckCalls: !check,
 	}
+	v2 := &mccpb.MixerFilterConfig{
+		MixerAttributes: &mpb.Attributes{
+			Attributes: map[string]*mpb.Attributes_AttributeValue{
+				AttrDestinationIP:  {Value: &mpb.Attributes_AttributeValue_StringValue{role.IPAddress}},
```
qiwzhang: IP should use "bytes" type

ayj: Thanks - fixed

proxy/envoy/mixer.go
```diff
@@ -125,20 +150,70 @@ func mixerHTTPRouteConfig(role proxy.Node, service string) *FilterMixerConfig {
 		},
 		QuotaName: MixerRequestCount,
 	}
-	if len(service) > 0 {
-		r.MixerAttributes[AttrDestinationService] = service
+	v2 := &mccpb.MixerFilterConfig{
+		MixerAttributes: &mpb.Attributes{
+			Attributes: map[string]*mpb.Attributes_AttributeValue{
+				AttrDestinationIP:  {Value: &mpb.Attributes_AttributeValue_BytesValue{net.ParseIP(role.IPAddress)}},
+				AttrDestinationUID: {Value: &mpb.Attributes_AttributeValue_StringValue{"kubernetes://" + role.ID}},
```
rshriram: Can we use this opportunity to externalize or parameterize this bit? May be have some platform identity as a core attribute of the role struct (defaults to k8s). You might have to have a separate PR to change the role struct and all proxy agent invocation (eg k8s~...~..). Then this becomes usable in non k8s environments as well.

proxy/envoy/testdata/lds-httpproxy.json.golden
```diff
@@ -33,7 +33,41 @@
...
+               }
+              }
+             }
+            }
+           },
+           "defaultDestinationService": "hello.default.svc.cluster.local",
+           "forwardAttributes": {
+            "attributes": {
+             "source.ip": {
+              "bytesValue": "AAAAAAAAAAAAAP//CgEBAA=="
```
qiwzhang: Is "bytes" type in JSON base64 encoded?

proxy/envoy/mixer.go
```diff
@@ -59,22 +67,35 @@ const (
...
 	// an empty name means no quota will be charged.
 	QuotaName string `json:"quota_name,omitempty"`
 
-	// If set to true, disables mixer check calls for TCP connections
+	// DEPRECATED: If set to true, disables mixer check calls for TCP connections
 	DisableTCPCheckCalls bool `json:"disable_tcp_check_calls,omitempty"`
+
+	// istio.io/api/mixer/v1/config/client configuration protobuf
+	// encoded as a generic map using canonical JSON encoding.
+	V2 map[string]interface{} `json:"v2,omitempty"`
```
qiwzhang: Please add comment
// if v2 field is not empty,  deprecated legacy fields are discarded.

proxy/envoy/mixer.go
ayj: +1 - I'll follow-up in another PR.

proxy/envoy/testdata/lds-httpproxy.json.golden
ayj: Yes - see https://developers.google.com/protocol-buffers/docs/proto3#json.

ayj: Clarification - the jsonpb encoding of bytes type is base64 encoded. The `v2` object in the mixerclient JSON object is jsonpb encoded but the parent JSON is normal json encoded.

kyessenov:
Merging prior to repo unification. @mandarjog please update ISTIO_API to head, the matching PR got merged.

kyessenov:
A follow-up question to @ayj : would it make more sense to set control service key in the forwarded attributes instead of mixer attributes? The client can pass the normalized host name to the destination. I assume there's some logic in mixerfilter to normalize the host head at the moment?

ayj:
Forwarded attributes aren't present on ingress. It also marginally increases the request size for something that could be derived on the local server-side proxy. Is there a functional advantage to including `destination.service` in the forwarded attributes?

Host header normalization introduces platform specific naming complexity to the mixerclient, e.g.  mapping of `name` to `name.namespace.svc.cluster.local` requires mixerclient know the destination is a k8s service and which `namespace` its running in. This logic isn't present in mixerclient today. +@qiwzhang for additional comments.

kyessenov:
We may want to preserve the address that the client used to address the server. Think of two services hosted on the same server. At TCP level, and potentially at HTTP level, all you have is IP address (hostnames are optional). I see that as a strong advantage of carrying the address from the client-side balancer to the server. Of course, we still need to do something when the client it outside of the mesh (e.g. ingress or non-istio pods).

ayj:
In your example, do HTTP and TCP share the same address:port? 

Forwarding attributes is a function of the mixerclient which, for now, means Envoy is making the client request. Don't we already assume Envoy includes the hostname? Otherwise our shared port HTTP services wouldn't work.

kyessenov:
Right, client envoy requires using Envoy. We still need to normalize the hostname, and that needs to be done at the client. For example, if a request sent from one namespace to another, and there are two identical services in the two namespaces, we cannot resolve hostname correctly in the destination namespace.

ayj:
How does envoy know which service instance to route to in that case? My understanding is that the app would need to either use (a) namespace qualified service name or (b) service or endpoint IP address. Both cases should be recognized by server's virtual_host domain list.

kyessenov:
It depends on the DNS context of the proxy. If the client and server are in different namespaces, then the hostname extension is different. The client can still use a short name. This is mostly a problem with multiple services sharing a pod.

ayj:
Two functionally identical services in different namespaces would be different services for the purposes of service naming and distributing mixerclient/quota/apispec configuration. In this case, a client could use the short short name (i.e. no namespace qualification) and it be routed to the service in the same namespace which the server-side proxy would map to the correct FQ destination service name.

kyessenov:
That's not necessarily the case. If my pod is reachable through names "a.ns1" and "a.ns2", and the client in "ns1" calls it through "a", then how does the server on the pod know which name was used? The server sees two options for services "a.ns1" and "a.ns2", and the host header "a". What is even worse is that if the request does not come from Istio, then host name is completely unreliable. It could be a random external IP if it's load balanced through some middle proxies.

PR #1666, opened 2017-10-31
let getDefaultNamespace() returns "default" if nothing is found.

jmuk:
https://github.com/istio/issues/issues/110 reminds me that istioctl
can set empty namespace for mixer configs, which will be delivered
directly and therefore it can cause same error of "empty namespace".

When the obtained namespace happens to be empty, I think it should
return the default value.

(cherry picked from commit c67bb1450974fa8916f74862dee5c3f8f37f0edf)

**Release note**:

```release-note
none
```

### DIFF
--- a/cmd/istioctl/main.go
+++ b/cmd/istioctl/main.go
@@ -697,7 +697,11 @@ func getDefaultNamespace(kubeconfig string) string {
 		return v1.NamespaceDefault
 	}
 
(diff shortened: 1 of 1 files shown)


yutongz:
I triggered it with [mkpj](https://github.com/istio/test-infra/tree/master/prow#manually-trigger-a-prow-job), but not sure why it was not being triggered at first place.

